PR #49, opened 2018-06-22
Maker Keeper Framework -related artifacts removed

reverendus:
There is no single Maker Keeper Framework anymore. Also the `keeper` repository doesn't exist anymore, it has been broken down into several smaller ones, dedicated to individual keepers. Even if this derivation is still able to clone the old version of the `keeper` repo, it is a very outdated codebase and probably won't even work as it requires specific older versions of some Python dependencies which isn't reflected in this derivation.


mbrock:
Aha, thanks. The specific versions of the Python dependencies might be the ones listed in `nix/python.nix`, but we should probably remove those too since we don't need them right now. I'll go ahead and remove them.

reverendus:
Oh right, I haven't spotted that they are in `nix/python.nix`. But some of them were incorrect versions anyway (`web3` for example, no idea why). I see you removed everything in https://github.com/dapphub/dapptools/commit/f23b2a3d334abfa5234b34939f93ba7d81c21e6c, cool.

Btw I messaged you in the Maker RocketChat regarding recreating that `nix` config for keepers in a different place.

PR #60, opened 2018-09-06
Bugfix - the 'Trace' panel wasn't using all the vertical space available

reverendus:


asymmetric:
Can we delete merged branches or is there a reason why we keep them around?

PR #62, opened 2018-09-08
Look for config in XDG_CONFIG_HOME

asymmetric:
To conform with the [XDG Base Directory
Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).


asymmetric:
I was actually debating whether we should deprecate `~/.sethrc` and remove it in a couple of versions, but I don't know if it's worth it.

PR #66, opened 2018-09-19
dapp: testnet: add command to reset chainstate

xwvvvvwx:


xwvvvvwx:
Just realised that `dapp testnet` does this already here: https://github.com/dapphub/dapptools/blob/master/src/dapp/libexec/dapp/dapp---testnet-launch#L114

So the error I was getting about a conflicting genesis block must have happened when `dapp` didn't exit cleanly?

Is it still worth adding this command or should we investigate why the testchain doesn't clean up after itself sometimes?

nanexcool:
Let's investigate a bit... but, if it finds a previous blockchain, instead of just failing it should say:

`Found previous data directory at blah, (possibly because of an error), run dapp testnet-clean or whatever to fix)`

xwvvvvwx:
or just `rm -rf "$chaindir"` every time it starts?

nanexcool:
I guess that works as well. Trying to think of an instance where this could
be dangerous

asymmetric:
I think it's better not to remove testchain data on start, as it's not very intuitive/expected, and maybe there's some test data that people want to look up that we risk deleting.

Probably the best thing to do would be investigating why `dapp` didn't exit cleanly.

If this is not happening to you anymore though, I'd consider closing and reopening whenever the issue comes up again.

asymmetric:
Do we still need this?

xwvvvvwx:
sorry, forgot about this :flushed: 

I just tried on my machine and I had this error where `dapp` had not cleaned up the chainstate properly, but I am unable to reproduce it again.

> Probably the best thing to do would be investigating why dapp didn't exit cleanly.

Agree this is the best way forward. I assume that there are some edge cases where traps are not called correctly.

> I think it's better not to remove testchain data on start, as it's not very intuitive/expected, and maybe there's some test data that people want to look up that we risk deleting.

What's the difference between removing the chaindata when the chain starts, and deleting it once the chain is stopped (as is the current indented behaviour)?

PR #71, opened 2018-10-10
(re)enable cheat address

rainbreak:
Experimental feature. This allows for setting the block timestamp from
within ds-test.

This feature should be used carefully as it is hevm specific and
prevents your ds-tests from running on arbitrary EVM implementations.

Usage:

```
hevm = Hevm(0x7109709ECfa91a80626fF3989D68f67F5b1DD12D);
hevm.warp(42);
```

where

```
contract Hevm {
    function warp(uint256) public;
}
```

### DIFF
--- a/src/hevm/CHANGELOG.md
+++ b/src/hevm/CHANGELOG.md
@@ -1,5 +1,10 @@
 # hevm changelog
 
+## 0.19 - 2018-10-09
+- Enable experimental 'cheat' address, allowing for modification of the
+  EVM environment from within the tests. Currently just the block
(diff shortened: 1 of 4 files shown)


rainbreak:
@livnev @mbrock

asymmetric:
Would it make sense to add a comment in the code explaining the tradeoffs?

jparklev:
Would this let `warp` set the block number and the `block.timestamp` or just the latter? @rainbreak

rainbreak:
In general anything is possible, but here it only sets the timestamp.

PR #76, opened 2018-11-15
Fully implement REVERT

### LINKED ISSUE #8: Implement 0xfd: REVERT fully
<a href="https://github.com/rainbreak"></img></a> **Issue by [rainbreak](https://github.com/rainbreak)**
_Monday May 28, 2018 at 12:45 GMT_
_Originally opened as https://github.com/dapphub/hevm/issues/38_

----

Needs to return data, similar to RETURN, but revert state changes.

rainbreak:
Populate the returndata buffer on revert. Fixes #8

### DIFF
--- a/src/hevm/src/EVM.hs
+++ b/src/hevm/src/EVM.hs
@@ -357,34 +357,11 @@ exec1 = do
 
     fees@(FeeSchedule {..}) = the block schedule
 
-    doStop =
-      case vm ^. frames of
-        [] ->
-          assign result (Just (VMSuccess ""))
-        (nextFrame : remainingFrames) -> do
-          popTrace
-          assign frames remainingFrames
-          assign state (view frameState nextFrame)
-          case view frameContext nextFrame of
-            CreationContext _ -> do
-              -- Move back the gas to the parent context
-              assign (state . gas) (the state gas)
-
-            CallContext _ _ _ _ _ _ -> do
-              -- Take back the remaining gas allowance
-              modifying (state . gas) (+ the state gas)
-              modifying burned (subtract (the state gas))
-          push 1
+    doStop = finishFrame (FrameReturned "")
 
   if the state pc >= num (BS.length (the state code))
     then
-      case view frames vm of
-        (nextFrame : remainingFrames) -> do
-          assign frames remainingFrames
-          assign state (view frameState nextFrame)
-          push 1
-        [] ->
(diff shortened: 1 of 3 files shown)


src/hevm/src/EVM.hs
```diff
@@ -1009,7 +1009,36 @@ exec1 = do
 
         -- op: REVERT
         0xfd ->
-          vmError Revert
+          case stk of
+            (xOffset:xSize:_) ->
+              accessMemoryRange fees xOffset xSize $ do
+                let output = readMemory (num xOffset) (num xSize) vm
+                case vm ^. frames of
+                  [] -> assign result (Just (VMFailure (Revert output)))
```
rainbreak: One bit I wasn't totally sure about, whether a revert counts as a VMFailure or a VMSuccess. @mbrock?

MrChico: I think many implementations have a custom VMRevert status for revert. This is at least what KEVM and https://github.com/ethereum/evmjit/blob/gh-pages/libevmjit/JIT.h#L92 does

mbrock: In our case `VMFailure` is right because it means the transaction didn't succeed. The `VMResult` (which is either success or failure) is only used to denote the result when the whole "transaction" exits (when the VM terminates and the stack frame is empty).

mbrock:
@rainbreak Your code looks correct to me, but there were already before some duplications of the frame stack popping logic, so I took the opportunity to try and refactor it into a big function. I'm going to run it through `dapp selftest` and the EVM test suite. It should be easier to overview the logic this way. Will push the patch here soon.

mbrock:
The `dapp selftest` passes and the EVM test suite runs as before. @rainbreak wanna check out the new function?

asymmetric:
Don’t understand the code but love the added comments 🎉

rainbreak:
@mbrock I like it :)

rainbreak:
Added formatting for the revert error trace. Also thinking about converting to text in the case that it is a solidity `Error(string)` message (like from revert and require), but the existing showAbi didn't seem to cover this for dynamic strings.

mbrock:
@rainbreak Cool, I'll merge. By the way today I almost finished a refactoring that gets rid of the `Blob` and `Byte` newtypes and the useless `forceConcreteBlob` function; I'll commit it tomorrow. Then I'd like to do a pass over EVM.hs to tidy up and add more comments...

Issue #74, opened 2018-10-29
hevm 0.21 fails to compile using Stack because secp256k1 is not properly linked


ggrieco-tob:
To reproduce, let's create a new stack project:

```
$ stack init test-hevm
$ cd test-hevm
```

Now, add `hevm` into `test-hevm.cabal` file as a dependency and add the following packages as `extra-deps` in `stack.yaml`

```
- hevm-0.21
- ghci-pretty-0.0.2@sha256:5f6c2a9f1d391aa0ad3c3d0bfe1f39decfe94be50fbf5a5279e2dbc18f8b4c51
- multiset-0.3.4@sha256:a82739a88411afb4a8b0d4d24a89038e7e6f7291e5721816bfbfa8138498fd54
- readline-1.0.3.0@sha256:db011062b90fbfbbc8b02666d835742c1d4de1f990a89fc056ef8082f3e3cba0
- restless-git-0.7@sha256:346a5775a586f07ecb291036a8d3016c3484ccdc188b574bcdec0a82c12db293
- rosezipper-0.2@sha256:aeff4e712360c5665747905381982a0765ba2c3831e352e8caeb951fc8c107b7
- s-cargot-0.1.4.0@sha256:61ea1833fbb4c80d93577144870e449d2007d311c34d74252850bb48aa8c31fb
- text-format-0.3.2@sha256:ea4e0cfd66e5a0224cbff2753eb160e71c9e383e86f293db77adeba5706bab94
- tree-view-0.5@sha256:85fbc67b53c1ef47f020a69051e6a29b27481698fe802cd2ed8ab0108aa69a38
- HSH-2.1.3@sha256:71ded11b224f5066373ce985ec63b10c87129850b33916736dd64fa2bea9ea0a
- ipprint-0.6@sha256:48ffe7c5acab06fbd70a68aa12bc980cc33dcf8f41bd76d692e3dee8805584c7
- sr-extra-1.46.3.2@sha256:d4eadfbd2238479559d9683dae6d0857534de87a67e05897f30fe582f7602713
- Unixutils-1.54.1@sha256:c6afbaf6d68ba229d0d262e066376c300f25b798d5752f162476155cadf1114d
```

Now download secp256k1 in a separated directory:

```
$ git clone https://github.com/bitcoin-core/secp256k1 --depth 1
$ cd secp256k1
$ ./autogen.sh
$ ./configure
$ make
$ cd ..
```

Finally:

```
[...]

PR #77, opened 2018-11-19
Added Mac OS hevm installation subtleties

Jasagredo:
Mac OS users have to deal with `readline` and `secp256k1` installations before being able to install hevm with Stack. This PR clarifies it on the README.

Also it updates the directories for the stack and cabal installations (they were referencing the old repository) and speaks about the possibility of installing hevm via Hackage.

### DIFF
--- a/src/hevm/README.md
+++ b/src/hevm/README.md
@@ -63,11 +63,9 @@ have GHC (the Haskell compiler) installed already, is to use
 [Stack](https://docs.haskellstack.org/en/stable/README/), which can
 take care of installing GHC for you.  These commands should work:
 
-```
-$ curl -sSL https://get.haskellstack.org/ | sh
-$ git clone https://github.com/dapphub/dapptools.git
-$ cd dapptools/src/hevm && stack setup && stack install
-```
+    $ curl -sSL https://get.haskellstack.org/ | sh
+    $ git clone https://github.com/dapphub/dapptools.git
+    $ cd dapptools/src/hevm && stack setup && stack install
 
 Also, hevm is in Hackage so you can execute `stack install hevm` to
 get it up and running.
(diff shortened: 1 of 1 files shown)


src/hevm/README.md
```diff
@@ -58,20 +58,37 @@ set the `TERMINFO` environment variable; on Ubuntu, you should do
...
-    $ curl -sSL https://get.haskellstack.org/ | sh
-    $ git clone https://github.com/dapphub/hevm
-    $ cd hevm && stack setup && stack install
-
-If you prefer to use your own installation of GHC and the basic
-Haskell package manager, Cabal, simply run:
-
-    $ git clone https://github.com/dapphub/hevm
-    $ cd hevm && cabal configure && cabal install
+If you can't or won't use Nix, the easiest way especially if you don't have GHC (the Haskell compiler) installed already, is to use [Stack](https://docs.haskellstack.org/en/stable/README/), which can take care of installing GHC for you.  These commands should work:
```
asymmetric: Why remove the hard linebreaks? I quite like them in markdown files, as they make navigation with vim much easier.

src/hevm/README.md
```diff
@@ -58,20 +58,37 @@ set the `TERMINFO` environment variable; on Ubuntu, you should do
...
+Also, hevm is in Hackage so you can execute `stack install hevm` to get it up and running.
+
+If you prefer to use your own installation of GHC and the basic Haskell package manager, Cabal, simply run:
+
+```
+$ git clone https://github.com/dapphub/dapptools.git
+$ cd dapptools/src/hevm && cabal configure && cabal install
+```
+
+**Note:** If you are on Mac OS environment when building with Stack you will have to previously deal with `readline` and `secp256k1`. These commands should be enough (rearding you have brew already installed):
```
asymmetric: If you are on macOS

asymmetric: > rearding you have brew already installed

What's `rearding` here?

src/hevm/README.md
```diff
@@ -58,20 +58,37 @@ set the `TERMINFO` environment variable; on Ubuntu, you should do
...
+$ git clone https://github.com/bitcoin-core/secp256k1.git
+$ cd secp256k1
+$ ./autogen.sh
+$ ./configure --enable-module-recovery # for generating secp256k1_recovery.h
+$ make
+$ sudo make install
+$ cd .. && rm -rf secp256k1 # optional (cleanup)
+```
+
+Then you can proceed with the stack installation as all the dependecies would be successfully fulfilled but when invoking `stack install` it's necessary to pass the flags `--extra-include-dirs=/usr/local/opt/readline/include --extra-lib-dirs=/usr/local/opt/readline/lib` for it to find readline.
```
asymmetric: It's not clear what the user should do here. Run `stack --extra-include-dirs=/usr/local/opt/readline/include --extra-lib-dirs=/usr/local/opt/readline/lib`? If so, would be better to just add it as another code block.

src/hevm/README.md
Jasagredo: Oh, there isn't a reason. I can break them and request a new pull request.

src/hevm/README.md
Jasagredo: Typo, wanted to write `regarding` trying to say "I suppose you have already installed brew"

src/hevm/README.md
asymmetric: Could you just undo these changes and push to this branch? So we can stay on the same PR.

src/hevm/README.md
```diff
@@ -63,15 +63,47 @@ have GHC (the Haskell compiler) installed already, is to use
 [Stack](https://docs.haskellstack.org/en/stable/README/), which can
 take care of installing GHC for you.  These commands should work:
 
-    $ curl -sSL https://get.haskellstack.org/ | sh
-    $ git clone https://github.com/dapphub/hevm
-    $ cd hevm && stack setup && stack install
+```
```
asymmetric: The rest of the README uses indentation to mark code blocks, so let's stick to that for now. We can change to code fences, if we want to, in a separate PR.

src/hevm/README.md
```diff
@@ -63,15 +63,47 @@ have GHC (the Haskell compiler) installed already, is to use
...
+$ brew install readline # for installing GNU readline in the system
+$ stack install readline --extra-include-dirs=/usr/local/opt/readline/include --extra-lib-dirs=/usr/local/opt/readline/lib # for installing the Haskell readline package
+$ git clone https://github.com/bitcoin-core/secp256k1.git
+$ cd secp256k1
+$ ./autogen.sh
+$ ./configure --enable-module-recovery # for generating secp256k1_recovery.h
+$ make
+$ sudo make install
+$ cd .. && rm -rf secp256k1 # optional (cleanup)
+```
```
asymmetric: But I'm fine if you leave this as a code-fenced block, as it's quite long :)

Jasagredo: I will use indentation as in the rest of the README

asymmetric:
We got a macOS machine for testing so we'll be looking at this shortly.

asymmetric:
The instructions seem to have helped [here](https://github.com/dapphub/dapptools/issues/74), merging.

PR #87, opened 2018-12-05
dapp: mention Bash 4 requirement for macOS

### LINKED ISSUE #78: `mapfile` not present on Mac OS
Mac OS is shipped using an older version of bash so it doesn't have some commands that are built-in in Bash 4. That's the case of `mapfile`.

This is the reason why `dapp init` terminates with a failure caused by `dapp test` that is using such function. The error it rises is this one:

```
/usr/local/libexec/dapp/dapp-build: line 19: mapfile: command not found
/usr/local/libexec/dapp/dapp---find-libraries: line 17: mapfile: command not found
```

On the other hand, running `hevm dapp-test` runs successfully all the tests without using `mapfile`.

Maybe this should be taken into account in future versions and use a different method if the system is Mac OS.

asymmetric:
Closes #78.

### DIFF
--- a/src/dapp/README.md
+++ b/src/dapp/README.md
@@ -11,6 +11,10 @@ Please make sure you have [`solc` installed in your CLI](https://solidity.readth
    make uninstall             uninstall dapp(1) from /usr/local
 ```
 
+### macOS
+
+On macOS, you need to install Bash 4.
+
 ## Usage
 
 ```


src/dapp/README.md
```diff
@@ -11,6 +11,10 @@ Please make sure you have [`solc` installed in your CLI](https://solidity.readth
    make uninstall             uninstall dapp(1) from /usr/local
 ```
 
+### macOS
+
+On macOS, you need to install Bash 4.
```
rainbreak: But only if they aren't installing via nix.

asymmetric: Isn’t that clear from context?  All of the section is tailored at non-nix users, so I’m wondering if we should add a separate nix one.

PR #88, opened 2018-12-10
Mention both nix and make installation methods

asymmetric:
* Mention that it can be installed with the dapp.tools installer 
* Not only macOS users might need to get Bash 4.

### DIFF
--- a/src/dapp/README.md
+++ b/src/dapp/README.md
@@ -5,15 +5,23 @@
 
 ## Installation
 
-Please make sure you have [`solc` installed in your CLI](https://solidity.readthedocs.io/en/develop/installing-solidity.html)
+`dapp` can be installed as part of the [dapp.tools suite](../../README.md).
+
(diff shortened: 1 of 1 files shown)


rainbreak:
Out of interest, is there a way to install with nix from within this directory?

asymmetric:
`nix-env -iA dapp -f ../../default.nix` should do it. Do you think it would make sense to add it to this README?

rainbreak:
I was meaning without referring to any of the directories above it :D

nanexcool:
I've also wondered if we can "decouple" the default.nix from dapp, seth, hevm, etc... from the main default.nix

asymmetric:
Hmm, no, don't see a good way to do it. The "upstream" `default.nix` ties all the pieces together: pinned nixpkgs, cached solc version. We could pass all of these things as inputs to the derivation, but we'd end up duplicating here what's already upstream - not sure it makes a lot of sense compared to `-f upstream`.

PR #89, opened 2018-12-10
enable hevm traces on passing unit tests

rainbreak:
With this change:

- `dapp test`: shows traces for failing tests
- `dapp test --verbose`: also show traces for passing tests

Also, emit raw event data in the case of anonymous events (e.g. ds-note).

@gbalabasquer @nanexcool :)

### DIFF
--- a/src/dapp/default.nix
+++ b/src/dapp/default.nix
@@ -3,7 +3,7 @@
 
 stdenv.mkDerivation rec {
   name = "dapp-${version}";
-  version = "0.9.0";
+  version = "0.9.1";
   src = ./.;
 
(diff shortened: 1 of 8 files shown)


nanexcool:
Works great! I'd modify the Changelog and up the version since @mbrock has been pretty consistent with that

asymmetric:
I guess we should upgrade both the `dapp` and `hevm` changelogs?

rainbreak:
dapp has a changelog?

nanexcool:
@asymmetric I'm merging this in an hour. We can merge the `dapp` Changelog as well

PR #90, opened 2018-12-10
--signature-only feature added for tx to return signature only

r001:
This request enables to use the feature `ethsign tx --signature-only ...` to only return the signature , and not the whole raw tx data. This enables us to use ethsign with Metamasks new external signing feature.


nanexcool:
I like it :) I agree `--sig` is enough

r001:
I have renamed --signature-only to --sig

PR #97, opened 2018-12-14
seth: remove -F argument from seth-sign usage

asymmetric:
It wasn't used.

I still think it would be better if we supported `-F`, but in the meantime, this takes care of the inconsistency.


nanexcool:
I don't understand. We do support `-F`, it's one of the flags available during the bootstrap phase in the `libexec/seth/seth` script

asymmetric:
We check here that `$# == 1` and exit otherwise. Are arguments handled in `seth`, rather than in the subcommands?

Issue #82, opened 2018-12-02
`seth` doesn't work with `solc 0.5.0`


reverendus:
```
# seth abi 'add(uint256)'
No input files given. If you wish to use the standard input please specify "-" explicitly.
```

reverendus:
```
--- a/src/seth/libexec/seth/seth-abi
+++ b/src/seth/libexec/seth/seth-abi
@@ -3,8 +3,8 @@
 ### Usage: seth abi <solidity-fragment>
 set -e
 [[ $1 ]] || seth --fail-usage "$0"
-input="pragma solidity ^0.4.8; contract A { $1 }"
-json=$(solc --combined-json=abi <<<"$input")
+input="contract A { $1 }"
+json=$(solc --combined-json=abi - <<<"$input")
 # Changed from "A" to "<stdin>:A" in solc 0.4.10:
 name=$(jshon <<<"$json" -e contracts -k)
 abi=$(jshon <<<"$json" -e contracts -e "$name" -e abi -u)
```

seems to solve the problem

rainbreak:
@reverendus did you get anywhere with this issue?

nanexcool:
This still causes things like `dapp create DSSpell` to fail, because of the `bytes` argument, for example:

```
dapp create DSSpell 0x6824b3909997c9d27e05e4bfa58054f3d497d6ae 0 0x
<stdin>:1:71: Error: Data location must be "memory" for parameter in function, but none was given.
pragma solidity ^0.5.6; contract A { function DSSpell(address,uint256,bytes) public; }
```

rainbreak:
Would be good to solve this by replacing `solc` with a js lib or a simple program.

reverendus:
> @reverendus did you get anywhere with this issue?

@rainbreak Looks like the problem is still there, I haven't tried to fix it.

asymmetric:
We found [this guy](https://github.com/ethers-io/ethers.js/blob/c2ce59f95e2b53cb4121c9e941a312913f4fc279/utils/abi-coder.js#L300), who can help us move away from using solc for this, which should allow us to close this issue.

Issue #84, opened 2018-12-04
Run functional tests when building


asymmetric:
To make sure that new versions of the dapptools work, we should run a subset of their functionality as part of the build process.

Examples:

- [x] `nix-build -A solc-versions.solc_$NEWVERSION` #101 
- [ ] spin up a dapp testnet
- [ ] Deploy a meaningful contract compiled with the last couple of solc versions (at least 0.4.24 and 0.5.0)
- [ ] Send txs to it, then read from it
- [ ] Contract should have a payable function, as well as functions that take in bytes and address[] to force solc 0.5.0 to force tagging them with memory or calldata

rainbreak:
`dapp selftest` already does some of this, so just extend that?

asymmetric:
`dapp selftest` runs `dapp test` on a bunch of repos.

I think we should test `seth` as well, and for that some kind of  test abstraction would be nice (e.g. `assert`).

MrChico:
Kinda addressed by #417 , closing

PR #99, opened 2018-12-14
seth: pass current solidity version to seth-abi

asymmetric:
This way, we don't have to update it manually whenever there are new,
backwards-incompatible solc releases.

See #82 and the fix in eee3859cc155e6e89e99bc16139ad72adbc3e5fc.

### DIFF
--- a/src/seth/CHANGELOG.md
+++ b/src/seth/CHANGELOG.md
@@ -1,0 +1,16 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [0.7.0] - 2018-12-14
+
+From this release on, changes, additions and removals will be documented in this
+changelog.
+
+### Changed
+- `seth abi`: use the current `solc` version in the `pragma` directive.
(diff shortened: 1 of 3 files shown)


asymmetric:
That's a good point. I guess it's a tradeoff between runtime performance and maintainability.

How often is seth run per second, and what's the overhead of running `solc --version`?

We could also pass the version to `seth` and down to `seth-abi`, but not sure if that works also when people don't use nix.

rainbreak:
Is there actually something that breaks this? You're thinking of `address payable` or similar?

nanexcool:
@rainbreak yeah that, and https://github.com/dapphub/dapptools/issues/82

asymmetric:
If this is a controversial change I think we can just close it. It was just meant to simplify maintenance, not to take too much brainpower away from more pressing matters :)

nanexcool:
I'd say this is a great opportunity to think of doing some kind of automated tests. I'm thinking...
- Spin up a `dapp testnet`
- Deploy a meaningful contract compiled with the last couple of solc versions (at least 0.4.24 and 0.5.0)
- Send txs to it, then read from it.

asymmetric:
Feel free to add your suggestions [here](https://github.com/dapphub/dapptools/issues/84).

rainbreak:
If it turns out to be annoying we can just change it again.

PR #101, opened 2018-12-17
nix: remove recursion from stable set

asymmetric:
`nix-build` does not evaluate sets recursively. This allows us to build all `solc` versions every time there's a change.

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -34,11 +34,12 @@ let
     inherit qrtx;
     inherit seth;
     inherit setzer;
-    inherit solc-versions;
     inherit token;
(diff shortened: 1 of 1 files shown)


asymmetric:
Basically what happens is that, by past experience, we have to restart the build once, cause the cache is warmed up on the first run.

I think that's a reasonable issue to deal with, but if you want we can also add the unstable (and actually I had it before), but i'm not sure if it's necessary.

nanexcool:
@asymmetric cool, then I'm fine with merging this.

PR #103, opened 2018-12-21
Add DAPP_SOLC_VERSION variable

nanexcool:
This is just a tiny bit hacky, but it works fine. Able to `export DAPP_SOLC_VERSION=0.5.0` for example, instead of having to `--use solc:0.5.0` with every command.

Command line arg takes precedence to the global var, so if the var is set to 0.5.0 but user types `--use solc:0.4.25` it'll correctly use `0.4.25`.

### DIFF
--- a/src/dapp/CHANGELOG.md
+++ b/src/dapp/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.9.2] - 2018-12-21
+- Introduce new global var `DAPP_SOLC_VERSION`. Overrides default `solc` used by `dapp`. Can be used with `export DAPP_SOLC_VERSION=0.5.0` instead of `dapp --use solc:0.5.0 [command]` for every command.
+
+## [0.9.1] - 2018-12-10
+- Pass `--verbose` flag to `hevm`
+
 ## [0.9.0] - 2018-12-04
 
 From this release on, we will document changes, additions and removals in this
--- a/src/dapp/CHANGELOG.md
+++ b/src/dapp/CHANGELOG.md
@@ -5,7 +5,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
 - Introduce a new environment variable `DAPP_SOLC_VERSION`. Overrides default `solc` used by `dapp`. Can be used with `export DAPP_SOLC_VERSION=0.5.0` instead of `dapp --use solc:0.5.0 [command]` for every command.
+
+## [0.9.1] - 2018-12-13
+### Added
 - Pass `--verbose` flag to `hevm`
 
 ## [0.9.0] - 2018-12-04
@@ -13,4 +17,5 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 From this release on, we will document changes, additions and removals in this
 changelog.
 
+[0.9.1]: https://github.com/dapphub/dapptools/tree/dapp/0.9.1
 [0.9.0]: https://github.com/dapphub/dapptools/tree/dapp/0.9.0
--- a/src/dapp/CHANGELOG.md
+++ b/src/dapp/CHANGELOG.md
@@ -4,10 +4,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.9.2] - 2018-12-21
-- Introduce new global var `DAPP_SOLC_VERSION`. Overrides default `solc` used by `dapp`. Can be used with `export DAPP_SOLC_VERSION=0.5.0` instead of `dapp --use solc:0.5.0 [command]` for every command.
-
-## [0.9.1] - 2018-12-10
+## [Unreleased]
+- Introduce a new environment variable `DAPP_SOLC_VERSION`. Overrides default `solc` used by `dapp`. Can be used with `export DAPP_SOLC_VERSION=0.5.0` instead of `dapp --use solc:0.5.0 [command]` for every command.
 - Pass `--verbose` flag to `hevm`
 
 ## [0.9.0] - 2018-12-04
--- a/src/dapp/default.nix
+++ b/src/dapp/default.nix
@@ -3,7 +3,7 @@
 
 stdenv.mkDerivation rec {
   name = "dapp-${version}";
-  version = "0.9.1";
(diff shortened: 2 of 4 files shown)


src/dapp/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.9.2] - 2018-12-21
+- Introduce new global var `DAPP_SOLC_VERSION`. Overrides default `solc` used by `dapp`. Can be used with `export DAPP_SOLC_VERSION=0.5.0` instead of `dapp --use solc:0.5.0 [command]` for every command.
```
asymmetric: Can we call it `environment variable`?

src/dapp/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.9.2] - 2018-12-21
```
asymmetric: We should add the actual link at the bottom. In order to to this, the version has to be tagged (we can do this after merging), and the tag pushed.

src/dapp/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
...
+- Introduce new global var `DAPP_SOLC_VERSION`. Overrides default `solc` used by `dapp`. Can be used with `export DAPP_SOLC_VERSION=0.5.0` instead of `dapp --use solc:0.5.0 [command]` for every command.
+
+## [0.9.1] - 2018-12-10
```
asymmetric: Same thing here.

src/dapp/CHANGELOG.md
asymmetric: Also, is this a bugfix? I think it's more like a backwards-compatible new feature, and so should bump the minor version, to 0.10.1.

If I read [SemVer](https://semver.org/) correctly.

src/dapp/libexec/dapp/dapp
```diff
@@ -25,6 +25,10 @@ fi
 
 if [[ $2 = --help ]]; then
   exec "${0##*/}" help -- "$1"
+elif [[ $DAPP_SOLC_VERSION ]]; then
+  v=$DAPP_SOLC_VERSION
+  unset DAPP_SOLC_VERSION
```
asymmetric: Why do we unset it here? I think that if someone exports the variable, they probably want this to persist, no?

src/dapp/CHANGELOG.md
nanexcool: yes, will fix

src/dapp/CHANGELOG.md
nanexcool: Yep, you're right. I'm used to only modifying the last number, but we should really aim for a 1.0 soon. I'll fix

src/dapp/CHANGELOG.md
nanexcool: yep

src/dapp/libexec/dapp/dapp
nanexcool: Ah, that's why I said this was hacky. When it detects the environment variable, it calls `dapp---use solc:$v [and everything else you put as commands]`. This in turn calls `dapp---nix-run` with an nix override that sets solc as the version you specified. `dapp --nix-run` can be used to alter more than `solc` if you wished. So if we had multiple versions of `hevm` as we do `solc`, and we specified a way to read that version, we could run `dapp` with a different `hevm` than the default.

src/dapp/CHANGELOG.md
nanexcool: Doesn't it change to `0.10.0` instead of `0.10.1`?

src/dapp/libexec/dapp/dapp
nanexcool: I unset because next time it runs, it should not go through that path, instead through the path below it

nanexcool:
@rainbreak yes. If installed via Nix, it'll be the default solc in overlay.nix. If installed via `make link` then it's whatever solc it finds in the `PATH`

src/dapp/CHANGELOG.md
asymmetric: Yes, of course, sorry :sweat_smile:

src/dapp/libexec/dapp/dapp
asymmetric: > I unset because next time it runs, it should not go through that path, instead through the path below it

Should it though? Wouldn't you expect the side-effects of the env var you set to persist until you unset it explicitly?

Basically:

* user runs `export DAPP_SOLC_VERSION=0.5.2`
* user runs `dapp test` - equivalent to `dapp --use solc:0.5.2 test`
* user runs `dapp test` again -- eq to `dapp --use solc:0.4.24 test`

Is this the behaviour you're after? If so, I don't think it's intuitive.

Or is the env var only being unset in a sub-shell?

nanexcool: Only in a sub-shell.
```
export DAPP_SOLC_VERSION=0.4.25
dapp --version
dapp --version
```
Will correctly use `0.4.25` both times.

src/dapp/CHANGELOG.md
asymmetric: Oh and another thing: we don't have to bump the version every time. We can bundle up a few changes under `Unreleased` and only release whenever we want. People will still get the `master` version 99.9% of the times anyway, the changelog is mostly to document what happened, less for actually shipping releases.

Whatever is easiest.

src/dapp/CHANGELOG.md
```diff
@@ -4,10 +4,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.9.2] - 2018-12-21
-- Introduce new global var `DAPP_SOLC_VERSION`. Overrides default `solc` used by `dapp`. Can be used with `export DAPP_SOLC_VERSION=0.5.0` instead of `dapp --use solc:0.5.0 [command]` for every command.
-
-## [0.9.1] - 2018-12-10
+## [Unreleased]
+- Introduce a new environment variable `DAPP_SOLC_VERSION`. Overrides default `solc` used by `dapp`. Can be used with `export DAPP_SOLC_VERSION=0.5.0` instead of `dapp --use solc:0.5.0 [command]` for every command.
```
asymmetric: Ah and one more thing: according to THE FORMAT, this should look like:

## [Unreleased]
### Added

- new environment variable ...

src/dapp/default.nix
```diff
@@ -3,7 +3,7 @@
 
 stdenv.mkDerivation rec {
   name = "dapp-${version}";
-  version = "0.9.1";
+  version = "0.9.2";
```
asymmetric: No bump necessary then.

PR #106, opened 2018-12-22
key management details

alexisgauba:
Added some nuances about key management that I didn't realize to the README in case it might be helpful to others developing with seth!


asymmetric:
Thanks!

PR #108, opened 2018-12-26
0.5.1 & 0.5.2

nanexcool:
Missing:

- [ ] Add lionello as maintainer
- [x] Put this in a CHANGELOG. (Where though? It has no effect unless we `dapp --use`)


src/dapp/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
+- Support for solc 0.5.1
```
asymmetric: And 0.5.2, right?

PR #110, opened 2019-01-11
set $SETH_STATUS variable true by default

gbalabasquer:


asymmetric:
Why do this?

nanexcool:
@asymmetric after Byzantium, every blockchain except for Ethereum Classic has this feature. It's easier to merge this, then figure out how to handle the edge case. When `SETH_STATUS` was created, it was to deal with the fact that Kovan already had it but not Mainnet.

gbalabasquer:
To add the for context, without this variable in true, a tx can fail deploying a contract but still the `seth` output will be that everything went fine. You will even get a contract address.

PR #117, opened 2019-01-27
Add constructor arguments to verification

travs:
Perhaps there is a more clever way to conditionally push elements to an array.
URL argument is spelt that way in [etherscan's API docs](https://etherscan.io/apis#contracts).

### DIFF
--- a/src/dapp/libexec/dapp/dapp-verify-contract
+++ b/src/dapp/libexec/dapp/dapp-verify-contract
@@ -49,6 +49,10 @@ params=(
   "apikey=$ETHERSCAN_API_KEY"
 )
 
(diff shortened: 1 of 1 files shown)


nanexcool:
Do we need these? Doesn't Etherscan not need constructor arguments anymore?

travs:
Oh is that true? I tried it a couple times the other day and it didn't seem to work without it, but perhaps I'm mistaken.
If you can confirm that we don't need it feel free to close this of course

rainbreak:
@nanexcool is this still necessary?

nanexcool:
nope, we can close this. The whole contract verification module needs to be rewritten

PR #124, opened 2019-02-03
Update geth, add goerli and kotti detection

nanexcool:
This will help for `dapp testnet` to have the latest constantinople updates


src/dapp/CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Support for solc 0.5.3
 
+### Changed
```
asymmetric: Is this to be added to 0.11.0 or to the next release? Cause the tag 0.11.0 has already been pushed.

nanexcool: can I unpush that tag?

nanexcool: gonna go with unreleased until we add some other things

Issue #91, opened 2018-12-11
hevm: format solidity revert messages


rainbreak:
Solidity treats the messages in `revert(msg)`and `require(bool, msg)` as string, not bytes. It would be ergonomic if `hevm` would decode these messages in the trace.

From the docs https://solidity.readthedocs.io/en/v0.5.0/control-structures.html#error-handling-assert-require-revert-and-exceptions (at the bottom)

> The provided string will be abi-encoded as if it were a call to a
> `function Error(string)`. In the above example, `revert("Not enough Ether provided.")`,
> will cause the following hexadecimal data be set as error return data:

```
0x08c379a0                                                         // Function selector for Error(string)
0x0000000000000000000000000000000000000000000000000000000000000020 // Data offset
0x000000000000000000000000000000000000000000000000000000000000001a // String length
0x4e6f7420656e6f7567682045746865722070726f76696465642e000000000000 // String data
```

PR #126, opened 2019-02-04
hevm/src/EVM/Format.hs: try to decode and display revert error string

livnev:
Trying to address #91


rainbreak:
:smile_cat: :chart_with_upwards_trend: :+1: fixes #91

PR #130, opened 2019-02-12
Add support for solc 0.5.4

nanexcool:


asymmetric:
Did you create the tag already?

nanexcool:
Just created it... tag is for commit hash right? should be the same on this branch or master

asymmetric:
Depends if we have a merge commit or not.

PR #131, opened 2019-02-12
bytecode view in hevm.el

livnev:
It works, but it gets a bit slow as soon as emacs saves some huge bytecodes into its cache. :cry:

### DIFF
--- a/src/hevm/hevm.el
+++ b/src/hevm/hevm.el
@@ -33,6 +33,9 @@
 (defvar hevm-stack-buffer nil
   "Buffer that displays the current VM's word stack.")
 
+(defvar hevm-waiting nil
+  "Are we waiting for the previous command to finish?")
+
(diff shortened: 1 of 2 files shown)


livnev:
Following @mbrock 's advice, factored to prepare the entire bytecode buffer in `Emacs.hs` before sending over, and `hevm.el` just looks up the correct line in the bytecode using `forward-search`. This is noticeably faster.

Bugs remaining:
(a) `hevm.el` doesn't seem to have the bytecode for the test contract, I think that's because `Emacs.hs` doesn't actually send it in the first setup interaction.
(b) queuing issue where consecutive presses of `n` cause a processing error.
(c) could be faster probably, there is still a perceptible delay at each step (around 0.5s) once some large bytecode has been loaded in...

livnev:
Can be merged IMO.

PR #134, opened 2019-02-13
[WIP] Add EXTCODEHASH and CREATE2 

livnev:
Testing against [these ad hoc solidity tests](https://github.com/dapphub/constantinople-tests/blob/master/src/ConstantinopleTests.t.sol). Hopefully we'll also be able to run those against e.g. `geth` with `dapp testnet`.

Please don't merge yet.

### DIFF
--- a/src/dapp-tests/constantinople.sol
+++ b/src/dapp-tests/constantinople.sol
@@ -8,29 +8,36 @@ contract DeadCode{
 
 contract ConstantinopleTests is DSTest {
 
+    ConstantinopleTests notmuch;
+    // this 5 byte-long initcode simply returns nothing
+    // PUSH1  00     PUSH1  00     RETURN
+    // 60     00     60     00     f3
+    bytes32 zerocode        = 0x60006000f3000000000000000000000000000000000000000000000000000000;
     // this 13 byte-long initcode simply returns 0xdeadbeef:
-    // PUSH4  de     ad     be     ef     PUSH1  00     MSTORE PUSH1  04     PUSH1  00     RETURN
-    // 63     de     ad     be     ef     60     00     52     60     04     60     00     f3
-    bytes32 deadcode        = 0x63deadbeef60005260046000f300000000000000000000000000000000000000;
+    // PUSH4  de     ad     be     ef     PUSH1  00     MSTORE PUSH1  32     PUSH1  00     RETURN
+    // 63     de     ad     be     ef     60     00     52     60     20     60     00     f3
+    bytes32 deadcode        = 0x63deadbeef60005260206000f300000000000000000000000000000000000000;
     // this 25 byte-long initcode returns deadcode (but without the padding)
     // PUSH1  0d     PUSH1  0c     PUSH1  00     CODECO PUSH1  0d     PUSH1  00     RETURN deadcode
     // 60     0d     60     0c     60     00     39     60     0d     60     00     f3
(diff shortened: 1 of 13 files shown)


rainbreak:
Should test CREATE as well, as you've touched that code and it isn't covered by VMTests

livnev:
Turns out the mistake was in the tests, all six tests are passing now (two for `CREATE2` and 4 for `EXTCODEHASH`). I want to add tests for:

(a) `EXTCODEHASH` on a precompiled contract should return `0x0`
(b) `EXTCODEHASH` on a contract selfdestructed earlier in the same transaction should return the same as if it hadn't been selfdestructed yet.
(c) `EXTCODEHASH` on a contract selfdestructed (in an earlier transaction) should return `0x0`: don't see how we can test this with `ds-test`.
(d) `EXTCODEHASH` on a contract during the execution of its initcode should return `0x0`, since it has no code. I'm concerned that `hevm` might actually fail this and similarly `EXTCODECOPY` and `EXTCODESIZE` could be buggy, due to how `hevm` uses `replaceCode`. Will investigate.
(e) `CREATE2` fails if contract with that address already exists (should be inherited from the existing behaviour of `CREATE`, but since this wasn't likely to happen with `CREATE` it's worth checking if the collision semantics were correct. 
(f) `CREATE2` fails if there was a contract with that address that selfdestructed earlier in the same transaction.
(g) `CREATE2` fails if it is called during execution of the initcode of another `CREATE2` generating a collision with the contract created by the parent `CREATE2`: in particular this tests that the target account nonce is being correctly incremented _before_ the execution of the initcode starts. 
(h) `CREATE2` succeeds if there was previously a contract at that address which selfdestructed in a previous transaction: again, don't know how in `ds-test`...

Also I need to finish the correct gas metering for `CREATE2`.

livnev:
There are also plenty of good test ideas [in this spreadsheet](https://docs.google.com/spreadsheets/d/1xat7UI8GtB4ZGVdlK5_XQSHJZaMThi4SrlcL8XMZb5Q/edit?pli=1#gid=1872446916).

mbrock:
@livnev Regarding (d), Hevm indeed treats the contract currently being created as having the code of its constructor, and later replaces it with the returned code. We probably need to make the `Contract` type distinguish between constructor and instance code. One way that would make the refactoring easy to get right is to change the type of `_bytecode` to be a new sum type `ContractCode` where

    data ContractCode = ConstructorCode ByteString | InstanceCode ByteString

and then fixing all the compilation errors. Contracts would start out with `ConstructorCode` and switch to `InstanceCode` in `replaceCode`.

livnev:
@mbrock I'm also going to take the opportunity to try getting rid of `codeSize`, which you marked is probably redundant, and is going to make it easier to not make a mistake in the external/internal initcode view distinction.

livnev:
@mbrock here it is. I experimented by adding two `Getter`s `bytecodeE` and `bytecodeI` that should replace `bytecode` everywhere: i.e. you have to decide if you want to get the bytecode as seen "externally" (e.g. through `EXTCODECOPY`, `CALL`, etc.) or "internally" (e.g. through `CODECOPY`, `CODESIZE`, etc.). Though `bytecodeI` might actually be made redundant if we change `CODECOPY` and `CODESIZE` to both just look at `the state code` instead of the former looking at `view bytecodeI this`. There is currently no other use for `bytecodeI`.

It would then be nice to somehow get `bytecodeE` to override the `Lens`-generated `bytecode`, but I'm not sure if that's possible.

livnev:
Closing in favour of #145, which includes these changes.

PR #137, opened 2019-02-20
solc: make z3 support optional, but defaulted true

asymmetric:
It's an [optional flag](https://solidity.readthedocs.io/en/latest/installing-solidity.html#smt-solvers), and we don't need it in the dss deployment scripts.


asymmetric:
This was [upstreamed](https://github.com/NixOS/nixpkgs/pull/56095).

PR #139, opened 2019-02-22
nix: fetch solc versions from pinned nixpkgs

asymmetric:
Advantages:

* reproducibility, as dependencies also match the exact version they had
in nixpkgs at the time
* cache-friendliness, as we can reuse cache.nixos.org instead of having to maintain
our own
* build times will be greatly reduced when bumping the pinned nixpkgs
version in the overlay, allowing us to bump it more often

Downsides:

* not all solc versions were in nixpkgs; we will lose the ability to
ship those (or would have to use the previous mechanism for them)


asymmetric:
Lol, the [failure](https://travis-ci.com/dapphub/dapptools/jobs/179800673#L4897) we see [was there](https://travis-ci.org/NixOS/nixpkgs/builds/250466114) with the original PR as well, but it still got merged.

asymmetric:
Another (big?) downside is that we'd have to wait for nixpkgs to merge a new version before we can use it.

asymmetric:
Fixed the build errors by separating linux and darwin versions.

@mbrock @icetan I'm sure this can be prettified, do you have any ideas?

asymmetric:
So with this, versions like 0.5.4 would be built with `solc-versions.unreleased.solc_0_5_4`.

When they're [merged](https://github.com/NixOS/nixpkgs/pull/56365) into upstream, we can move them to the normal set.

Is that ok @nanexcool?

nanexcool:
Yeah, I like that! We should really document some of this in the main README

asymmetric:
> We should really document some of this in the main README

Done.

PR #140, opened 2019-02-22
dapp: add gnumake as a dependency

### LINKED ISSUE #111: dapp init requires make
`dapp init` [requires `make`](https://github.com/dapphub/dapptools/blob/244a770df3345ac3520cb11e70ab11206fe24a47/src/dapp/libexec/dapp/dapp-init#L77), which is not installed by default on NixOS.

AFAICT, `dapp` does not declare `make` as a runtime dependency, which it should.

asymmetric:
it's used by dapp-init.

Closes #111.


asymmetric:
Tag pushed.

PR #145, opened 2019-03-04
run BlockchainTests/GeneralStateTests

### LINKED ISSUE #92: hevm: pass all General State Tests
Ethereum has a bunch of general client tests at https://github.com/ethereum/tests. Not all of these are relevant for `hevm`, e.g. consensus, rlp, difficulty, blockchain, as `hevm` is purely an EVM runner.

`hevm` is currently passing the VMTests and has some internal code to support this.

> This is specifically not meant to cover transaction, creation or call processing, or management of the state trie. Indeed at least one implementation tests the VM without calling into any Trie code at all.

> A VM test is based around the notion of executing a single piece of code as part of a transaction, described by the exec portion of the test. 

https://ethereum-tests.readthedocs.io/en/latest/test_types/vm_tests.html

`hevm` does not support running the GeneralStateTests, and it is not clear whether it would pass if it did. However, some opcodes only have coverage in the state tests, notably `CALL` and `CREATE`.

> A state test is based around the notion of executing a single transaction, described by the transaction portion of the test.

https://ethereum-tests.readthedocs.io/en/latest/test_types/state_tests.html

The vm tests do have some basic coverage of call / create: 

> Since these tests are meant only as a basic test of VM operation, the CALL and CREATE instructions are not actually executed. To provide the possibility of testing to guarantee they were actually run at all, a separate portion callcreates details each CALL or CREATE operation in the order they would have been executed. Furthermore, gas required is simply that of the VM execution: the gas cost for transaction processing is excluded.

TODO:
- [x] determine which subset of the state tests is relevant or desirable
- [x] add support for running individual state tests
- [ ] ensure passing all the relevant tests

livnev:
(please don't merge yet, this is still WIP)

Aim is to resolve #92.

### DIFF
--- a/src/dapp-tests/constantinople.sol
+++ b/src/dapp-tests/constantinople.sol
@@ -8,29 +8,36 @@ contract DeadCode{
 
 contract ConstantinopleTests is DSTest {
 
+    ConstantinopleTests notmuch;
+    // this 5 byte-long initcode simply returns nothing
+    // PUSH1  00     PUSH1  00     RETURN
+    // 60     00     60     00     f3
+    bytes32 zerocode        = 0x60006000f3000000000000000000000000000000000000000000000000000000;
     // this 13 byte-long initcode simply returns 0xdeadbeef:
-    // PUSH4  de     ad     be     ef     PUSH1  00     MSTORE PUSH1  04     PUSH1  00     RETURN
-    // 63     de     ad     be     ef     60     00     52     60     04     60     00     f3
-    bytes32 deadcode        = 0x63deadbeef60005260046000f300000000000000000000000000000000000000;
+    // PUSH4  de     ad     be     ef     PUSH1  00     MSTORE PUSH1  32     PUSH1  00     RETURN
+    // 63     de     ad     be     ef     60     00     52     60     20     60     00     f3
+    bytes32 deadcode        = 0x63deadbeef60005260206000f300000000000000000000000000000000000000;
     // this 25 byte-long initcode returns deadcode (but without the padding)
     // PUSH1  0d     PUSH1  0c     PUSH1  00     CODECO PUSH1  0d     PUSH1  00     RETURN deadcode
     // 60     0d     60     0c     60     00     39     60     0d     60     00     f3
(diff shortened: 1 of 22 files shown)


rainbreak:
The eip 161 compliance should be imperceptible within the vm right? It's just "technical correctness".

livnev:
@rainbreak I think you're right that it's not perceptible within the VM, i.e. there's no way to tell from within a program whether or not an empty account has been deleted (though this fact is at least a subtle one IMO). But unfortunately the existence of these empty accounts is checked in the GeneralStateTests, and thousands of tests fail if you get it wrong.

n.b. that empty accounts can have nonzero storage, meaning that conceivably certain types of analysis can give a different answer depending on whether or not you've correctly implemented empty account pruning.

livnev:
I just did a major manicuring of the commit history here to tidy things up and separate things. There are now a few distinct parts to this PR (they could be split into separate PRs but I think it would needlessly complicate the process tbh):

- **implementing `EXTCODEHASH` and `CREATE2`, making `hevm` up-to-date with Petersburg/ConstantinipleFix semantics, at least at the opcode level (the bitshift operators were already added earlier)** [first two commits]
- adding some basic solidity tests for the new opcodes [next two commits]
- fixing a bug in the chronology of code deployment [next commit]
- minor improvements to the workflow for the existing VMTest runner [next two commits]
- **adding support for running the `BlockchainTest` format from the Ethereum test suite** [next commit]
- a few improvements to the testing functionality for both the new and the old ones [next two commits]
- **fixing a bunch of semantic bugs that these new tests turned up in the past few weeks** [remaining commmits]

Here are the test results before any of the fixes in the last bullet point were applied:

screenshot2

And here are the results now:

screenshot1

The curious reader can rerun the `GeneralStateTests` for every commit in between to track the improvement:

```sh
$ export HEVM=path/to/hevm
$ ./run-blockchain-tests path/to/ethereum/tests > blockchain-tests.html
```

I've triaged some of the remaning failures into the following categories:

(i) subtle issues with gas metering, probably attributable to one or two mistakes (~4000 failures)
(ii) unimplemented precompiles (thousands of failures)
(iii) unknown (at most hundred failures)

rainbreak:
Will merge on ci passing.

PR #163, opened 2019-03-28
add solc 0.5.7, keep 0.5.6 as default

nanexcool:


rainbreak:
@asymmetric it's because ds-note, which is in basically everything, breaks with 0.5.7 because of the use of `callvalue` in an assembly block. ds-note needs updating.

PR #167, opened 2019-04-03
Update HEVM Makefile

cmditch:
use `cabal new-*` cmds (https://www.haskell.org/cabal/users-guide/nix-local-build-overview.html)
as well as `--pure` with `nix-shell` invocations (https://nixos.org/nix/manual/#options-1)

Note - Credit to @livnev for the proposed changes. Just snagged an opportunity to make this change/PR for ceremonial purposes.

### DIFF
--- a/src/hevm/Makefile
+++ b/src/hevm/Makefile
@@ -1,16 +1,16 @@
 all: build
 
 default.nix: hevm.cabal
-	nix-shell --command 'cabal2nix . > default.nix'
+	nix-shell --command 'cabal2nix . > default.nix' --pure
 .configured: default.nix shell.nix
-	nix-shell --command 'cabal configure --enable-tests'
+	nix-shell --command 'cabal new-configure --enable-tests' --pure
 	touch .configured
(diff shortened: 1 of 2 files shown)


livnev:
@mbrock do you think it's a good idea to switch to `cabal new-*` and to use `--pure`? The former is something I was advocating because I have issues with the old-style `cabal` commands on my nixOS machine (though they seem to work fine with `nix` on debian).

and `--pure` also seems like it would be a good idea?

cmditch:
Note: `make tests` currently fails, since `solc` does not exist within `nix-shell --pure`

livnev:
> Note: `make tests` currently fails, since `solc` does not exist within `nix-shell --pure`

adding `pkgs.solc` to `shell.nix` fixes that.

rainbreak:
Is this ready to go?

livnev:
@mbrock do you have any opinions about whether or not `--pure` is a good thing and whether or not `cabal new-*` is a good thing?

PR #169, opened 2019-04-15
Get `stack build` working again

cmditch:


rainbreak:
Thanks! Was there a simple way to populate the extra-deps in stack.yaml?

cmditch:
Ideally, I'd like to get hevm compiling with LTS friendly library versions, and there will be no need for very many extra-deps.

We could also switch to stack's package.yaml, which will generate hevm.cabal for us, and make it so we don't have two sources of versioning truth, e.g. hevm.cabal and stack.yaml, rather package.yaml will be the one source of truth.

rainbreak:
Just going to merge this as it actually makes stack work again.

Issue #170, opened 2019-04-19
Support for signed negative numbers


eordano:
I'm trying to run:

```
seth --to-hex '-1'
```

Which clearly fails

```
seth---to-hex: error: bad number: `-1'
```

Because https://github.com/dapphub/dapptools/blob/master/src/seth/libexec/seth/seth---to-hex#L8 doesn't think the `-` is a valid part of a number.

My usecase is:
```
seth call 0xf87e31492faf9a91b02ee0deaad50d51d56d5d4d 'ownerOfLand(int256,int256)' 123 '-143'
```

Which should tell me who is the owner of the parcel 123,-143 in Decentraland... I know it's fairly uncommon to use signed integers in contracts, but I wonder if you would ever support something like this (my bash foo is not strong enough as to propose a PR)

Thanks for making this awesome tool!

asymmetric:
Closed by #172.

rainbreak:
Does #172 actually close this, or just partially address it?

asymmetric:
True:

```
❯ seth --to-hex (seth --to-int256 -9000)
seth---to-hex: error: missing `0x' prefix: `ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffdcd8'
```

rainbreak:
I think this is something that seth is inconsistent on. I think that all hex output should be given a `0x` prefix.

asymmetric:
The error message is actually incorrect. There's a [regexp](
https://github.com/dapphub/dapptools/tree/cb9cbcdcbe3d7449d122c4543013fa884517113a/src/seth/libexec/seth/seth---to-hex#L8-L9) to match against the hex digit character class, which [seems](https://www.regular-expressions.info/posixbrackets.html) to mean `[A-Fa-f0-9]`.

So a) the error message is wrong, and b) I'm not sure why -9000 (`ff...ff`) doesn't match the regexp.

asymmetric:
I had another look.

The usecase mentioned in the issue is satisfied by `seth---to-int256` - no need to pass it to `seth---to-hex`, as it's already in hex.

So I think the issue can actually be closed.

I can open another one to deal with the `0x` prefix.

eordano:
Thank you!

PR #172, opened 2019-04-26
add seth---to-int256 macro

MrChico:
addresses https://github.com/dapphub/dapptools/issues/170

### DIFF
--- a/src/seth/libexec/seth/seth---to-int256
+++ b/src/seth/libexec/seth/seth---to-int256
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+### seth---to-int256 -- convert an integer into int256 value
+### Usage: seth --to-int256 <value>
+set -e
+export BC_LINE_LENGTH=100
+seth --to-uint256 "$(bc <<< "(2^256 * ($1 < 0)) + $1")"


MrChico:
Does anyone want to review before I merge @asymmetric @rainbreak ?

src/seth/libexec/seth/seth---to-int256
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+### seth---to-int256 -- convert an integer into int256 value
+### Usage: seth --to-int256 <value>
+set -e
+export BC_LINE_LENGTH=5000
+seth --to-uint256 $(bc <<< "(2^256 * ($1 < 0)) + $1")
```
asymmetric: [`shellcheck`](http://shellcheck.net/) gives a warning about missing quotes around the process substitution - could you do that?

I can recommend installing a `shellcheck` plugin locally btw, and I plan to add it to the CI pipeline.

src/seth/libexec/seth/seth---to-int256
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+### seth---to-int256 -- convert an integer into int256 value
+### Usage: seth --to-int256 <value>
+set -e
+export BC_LINE_LENGTH=5000
```
asymmetric: Is the actual max number of characters not `2 * bytes(word_size) = 64`?

src/seth/libexec/seth/seth---to-int256
livnev: 2nded, @MrChico I recommend installing `shellcheck` and `(add-hook 'sh-mode-hook 'flycheck-mode)`.

asymmetric: It's now run in Travis https://github.com/dapphub/dapptools/pull/175

MrChico: Thanks, should be fixed now

src/seth/libexec/seth/seth---to-int256
MrChico: This is just to force `bc` to not break lines. The actual length is however many digits 2^256 (in decimal) has. But 5000 suffices

rainbreak:
```
$ BC_LINE_LENGTH=5000 bc <<< 2^256 | wc -c
79
```

PR #173, opened 2019-04-30
Minor readme updates

MrChico:


README.md
```diff
@@ -8,7 +8,21 @@ This repository contains the source code of several dapp tools
...
 
-To install tools from this repository follow these instructions:
+Contents:
+
+- [Dapp](./src/dapp)
+- [Seth](./src/seth)
+- [Hevm](./src/hevm)
+- [Ethsign](./src/hevm)
+
+We use `Nix` to for packaging and managing our tools.
```
asymmetric: "to for"

README.md
```diff
@@ -8,7 +8,21 @@ This repository contains the source code of several dapp tools
...
+
+All dependencies and tools can be automaticaly installed by running:
+```sh
+curl dapp.tools/install | sh
```
asymmetric: I think we should have an `https://` here, just in case our server-side redirect goes away at any point in the future.

README.md
```diff
@@ -8,7 +8,21 @@ This repository contains the source code of several dapp tools
...
+```
+
+For a manual install, please follow the following instructions.
```
asymmetric: "follow the following" sounds bad but I don't know how to improve it 😢

src/seth/libexec/seth/seth---to-int256
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
```
asymmetric: This file got here by accident I reckon.

README.md
asymmetric: Also I think this is already mentioned above, where it says

> along with various Nix expressions which are all bundled up into a Nix overlay.

WDYT? Maybe one of these can be removed.

README.md
MrChico: Maybe "heed" the instructions 😏

asymmetric: - "have a look at"
- "check out"
- "peruse"
- "enjoy"
- "scope"

asymmetric: “follow the instructions below”

MrChico:
Thanks for comments @asymmetric should be good now

PR #175, opened 2019-05-02
dapp 0.18.0 + seth 0.8.0

asymmetric:
Two new releases, brought about by the fact that this PR also adds `shellcheck` checking on Travis, which helped identify bugs and issues.


asymmetric:
@icetan @nanexcool wdyt?

PR #177, opened 2019-05-04
More control over hevm test output

### LINKED ISSUE #159: hevm: summarised dapp-test output
Would be neat to make `hevm dapp-test` more readable by:

- making test output for failing tests not show the full trace by default, but only the listing of assertion failures, with the full trace available with `dapp test -v`
- giving a summary of the number of passing, failing, and bailing tests at the end of the output.

@rainbreak

rainbreak:
Modifies the `dapp test` output, hiding the trace by default and allowing showing the trace for passing tests.

- `dapp test`: only assertion failures for failing tests
- `dapp test -v`: full trace for failing tests
- `dapp test -vv`: full trace for all tests

Also, display unknown event output in `logn` binary form, including topics.

Fixes #159

### DIFF
--- a/src/dapp/libexec/dapp/dapp-test-hevm
+++ b/src/dapp/libexec/dapp/dapp-test-hevm
@@ -9,8 +9,11 @@ while [[ $# -gt 0 ]]; do
     -r|--regex)
       opts+=(--match "$1"); shift
       ;;
+    -vv)
+      opts+=(--verbose 2);
+      ;;
     -v|--verbose)
-      opts+=(--verbose);
+      opts+=(--verbose 1);
       ;;
     *)
       echo >&2 "dapp-test-hevm: unknown argument: $opt"
--- a/src/hevm/src/EVM/UnitTest.hs
+++ b/src/hevm/src/EVM/UnitTest.hs
@@ -59,7 +59,7 @@ import qualified Data.Vector as Vector
 data UnitTestOptions = UnitTestOptions
   {
     oracle     :: Query -> IO (EVM ())
-  , verbose    :: Bool
+  , verbose    :: Maybe Int
   , match      :: Text
   , vmModifier :: VM -> VM
   , testParams :: TestVMParams
@@ -427,13 +427,16 @@ runUnitTestContract
                      (fromIntegral gasSpent :: Integer)
                in
                  pure
-                   ( "PASS " <> testName <> " (gas: " <> gasText <> ")"
+                   ("\x1b[32m[PASS]\x1b[0m "
+                   <> testName <> " (gas: " <> gasText <> ")"
                    , Right (passOutput vm dapp opts testName)
                    )
              (Right False, vm) ->
-               pure ("FAIL " <> testName, Left (failOutput vm dapp opts testName))
+               pure ("\x1b[31m[FAIL]\x1b[0m "
+               <> testName, Left (failOutput vm dapp opts testName))
              (Left _, _)       ->
-               pure ("OOPS " <> testName, Left ("VM error for " <> testName))
+               pure ("\x1b[33m[OOPS]\x1b[0m "
+               <> testName, Left ("VM error for " <> testName))
 
       let inform = \(x, y) -> Text.putStrLn x >> pure y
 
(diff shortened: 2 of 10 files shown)


src/hevm/src/EVM/UnitTest.hs
```diff
@@ -427,28 +427,31 @@ runUnitTestContract
...
       let inform = \(x, y) -> Text.putStrLn x >> pure y
 
       -- Run all the test cases and print their status updates
       details <-
         mapM (\x -> runOne x >>= inform) testNames
 
-      let mails = [x | Right x <- details]
-      let fails = [x | Left x <- details]
+      let running = [x | Right x <- details]
+      let bailing = [x | Left x <- details]
```
asymmetric: Surprised by this change.

src/dapp/libexec/dapp/dapp-test-hevm
```diff
@@ -9,8 +9,11 @@ while [[ $# -gt 0 ]]; do
     -r|--regex)
       opts+=(--match "$1"); shift
       ;;
+    -vv)
+      opts+=(--verbose=2);
+      ;;
     -v|--verbose)
-      opts+=(--verbose);
+      opts+=(--verbose=1);
```
asymmetric: This needs to be quoted, according to shellcheck.

src/hevm/src/EVM/UnitTest.hs
rainbreak: `fails` was actually misleading. The `running` tests contain both passing and failing tests, where the `bailing` tests are those with a vm error.

src/dapp/libexec/dapp/dapp-test-hevm
rainbreak: Or don't use the `=`

asymmetric:
I wonder if it's useful for the default to show no trace for failures. I expect many users to have to run the tests again with `-v`.

I think the behaviour of `-v` would make for a better default.

* `dapp test`: trace for failing tests
* `dapp test -v`: trace for all tests
* `dapp test --compact`: no traces

src/hevm/src/EVM/UnitTest.hs
```diff
@@ -427,28 +427,31 @@ runUnitTestContract
                      (fromIntegral gasSpent :: Integer)
                in
                  pure
-                   ( "PASS " <> testName <> " (gas: " <> gasText <> ")"
+                   ("\x1b[32m[PASS]\x1b[0m "
```
asymmetric: 😍

rainbreak:
@asymmetric the default `dapp test` is to show output like

```
Failure: test_add
  
  "Error: Wrong `uint' value"
    Expected: 1
      Actual: 0
```

Which comes from the various `assertEq` statements. This could be improved with line numbers etc, but are you saying that this is less useful than the full trace? The full trace contains a lot of extra stuff, e.g. the `setUp`, making it hard to focus on the tested lines.

(btw this is how it used to work, before switching to hevm)

gbalabasquer:
> I wonder if it's useful for the default to show no trace for failures. I expect many users to have to run the tests again with `-v`.
> 
> I think the behaviour of `-v` would make for a better default.
> 
> * `dapp test`: trace for failing tests
> * `dapp test -v`: trace for all tests
> * `dapp test --compact`: no traces

I'd prefer to not have any trace if the failure is a from a test assertion (as we have the standard message showing the expected and got values which in most of cases is enough). However if there is an evm error IMO @asymmetric is right. Probably the best is to show the trace as the default option.
Could we do some differentiation between these two types of failures?

asymmetric:
@rainbreak you're right, the output of `dapp test` according to your PR is intuitive (modulo the missing line numbers).

Does this PR address #159?

rainbreak:
@asymmetric yes, it does. There are subsequent improvements to make with line numbers etc.

asymmetric:
I'm fine with merging this and creating a separate issue for line numbers.

asymmetric:
This should have probably included a bump in the minor version in `dapp`'s changelog, due to the change in `dapp test`'s interface.

PR #182, opened 2019-05-07
seth: fix SC2207 by using the read builtin

### LINKED ISSUE #180: mapfile breaks seth-send
See https://github.com/dapphub/dapptools/pull/179.

It's possible that the problem only presents itself when some env variables [read by `seth---send-params`](https://github.com/dapphub/dapptools/pull/179) are not set.

asymmetric:
The output of `seth---send-params` is a space-separated list of words on
one line, rather than on multiple lines, so mapfile followed by `$params[@]` was
outputting a quoted string, which broke json parsing.

Closes #180.

/cc @nanexcool @icetan @kmbarry1


nanexcool:
Works on my machine! :D

PR #184, opened 2019-05-09
hevm: add version subcommand

rainbreak:
```
$ hevm version
0.30
```

Not sure if `--version` is preferable, or even how to get optparse-generic to do that.


rainbreak:
@asymmetric opinion about version flags?

asymmetric:
dapp and seth only support `--version`, and I think consistency is the thing to optimize for in this case.

rainbreak:
I think `--version` isn't actually possible without replacing the whole cli parser (currently optparse-applicative, could use optparse-generic). Sticking with `version` for now.

Issue #112, opened 2019-01-16
dapp install with no arguments has ugly error message


asymmetric:
Instead of showing git's failure at cloning an empty string, it could print a usage blob (also considering there's 3 different syntaxes for adding libraries)

kmbarry1:
Taking this as a warm-up bug; should have something soon^TM.

kmbarry1:
Also, what's the third syntax? I know one can do the following, which I assume are the first two:
dapp install ds-foo
dapp install https://github.com/dapphub/ds-foo

The third seems to be some sort of relative path thing, but I don't see how it's intended to work.

rainbreak:
@kmbarry1 the third syntax is `dapp install github_org_name/ds-foo`, e.g. `dapp install makerdao/dss`.

PR #195, opened 2019-05-09
no-arg 'dapp install' prints usage

kmbarry1:
Also fleshed out the usage blurb a bit to cover the three syntaxes.


kmbarry1:
Closes #112

asymmetric:
Thanks!

PR #196, opened 2019-05-10
Add `solc --allow-paths` flag for nix store paths in `solidityPackage` builder

icetan:
To fix `Source “<filename>” not found: File outside of allowed directories.` problem when using `solidityPackage` with `deps`.


asymmetric:
Is there a test case we can run this against?

icetan:
> Is there a test case we can run this against?

I put together a nix expression that illustrates the issue: https://gist.github.com/icetan/9f60e70494d4d53aa2725f324cf86f93

PR #197, opened 2019-05-10
Add `solc` argument to `solidityPackage`

icetan:
To allow for inversion of control of which version of `solc` to compile with.

This is useful when you need to provide a version of `solc` other then the current default one so that you can comply with the pragma of the contracts you want to compile.

### DIFF
--- a/nix/solidity-package.nix
+++ b/nix/solidity-package.nix
@@ -15,9 +15,9 @@
          xs);
 in
   pkgs.lib.makeOverridable (
-    attrs @ { test ? true, deps ? [], ... }:
+    attrs @ { test ? true, deps ? [], solc ? pkgs.solc, ... }:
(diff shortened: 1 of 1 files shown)


asymmetric:
Can you provide some context on where this is useful?

I'm also thinking that it could be more intuitive if the `solc` attribute were mandatory. WDYT?

icetan:
> Can you provide some context on where this is useful?

I updated my initial comment to give more context.

> I'm also thinking that it could be more intuitive if the `solc` attribute were mandatory. WDYT?

Yes, that might be a good idea. It would create a bit more awareness of which `solc` is actually being used.

On the other hand it would break eventual backwards compatibility this is why I chose this solution.

asymmetric:
Not sure why this isn't triggering a CI build.

Can you amend the commit, adding the rest of the context, so we see if the build gets triggered?

PR #199, opened 2019-05-10
hevm: regular integer notation in test output

### LINKED ISSUE #198: Error message rounding values [dapp]
When a test fails in an `assertEq` statement, if the values are very large, they will be rounded returning errors like this:
```
"Error: Wrong `uint' value"
    Expected: ~5.25000000e19
      Actual: ~5.25000000e19
```

It would be nice that no rounding is applied so we can get accurate results.

rainbreak:
fixes #198 

```
  "Error: Wrong `uint' value"
    Expected: 1250000000000000000
      Actual: 1250000000000000001
  "Error: Wrong `int' value"
    Expected: -1250000000000000000
      Actual: -1250000000000000001
  "Error: Wrong `uint' value"
    Expected: 1250000000000000000
      Actual: 1
```

Could consider right justifying.

### DIFF
--- a/src/hevm/src/EVM/Format.hs
+++ b/src/hevm/src/EVM/Format.hs
@@ -34,7 +34,6 @@ import Numeric (showHex)
 import qualified Data.ByteString as BS
 import qualified Data.Char as Char
 import qualified Data.Map as Map
-import qualified Data.Scientific as Scientific
 import qualified Data.Text as Text
 
 data Signedness = Signed | Unsigned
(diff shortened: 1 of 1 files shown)


rainbreak:
This will affect the tty output as well, e.g. in interactive mode.

asymmetric:
Right justifying would look better, I think.

Will the output fit in interactive mode, or are there issues there?

rainbreak:
The maximum width is fairly large:

```
  "Error: Wrong `uint' value"
    Expected:                                                            1250000000000000000
      Actual: 115792089237316195423570985008687907853269984665640564039457584007913129639935
```

livnev:
@rain don't we want this change only for test output, and keep the `hevm interactive` display as it is?

rainbreak:
@livnev which change do you mean, the right justifying (not implemented yet) or the regular, non-scientific notation?

livnev:
> @livnev which change do you mean, the right justifying (not implemented yet) or the regular, non-scientific notation?

Displaying a full word in decimal, "unscientific notation" I think already takes up too much space on most screens, so I think `hevm interactive` should keep the truncated view with scientific notation?

rainbreak:
I think by far the most common "long" numbers are `max_(u)int{256,128,64}` etc., so I'll special case those and display a symbol instead. We can retain the current format for the tty, but perhaps longer term the tty should be choosing how to render based on available space.

rainbreak:
Have just special cased max_uint.

PR #200, opened 2019-05-10
IDENTITY, SHA256, RIPEMD160 precompile

cmditch:
The identity, sha256, and ripemd160 precompiles now pass the eth tests.

This begins a work-in-progress refactor for handling precompiles which aims to address the gas cost calculation issues present on the old implementation.

More tests are _correctly_ failing now. For example a bunch of ecmul and modexp precompiles were passing despite not being implemented. 

**Test results**
Before:  5080 passed, 3999 bad-balance, 1533 failed
After: ~4425 passed, 3728 bad-balance, 2459 failed~  need to re-run since sha256/ripemd are done

- Prettify balance/state output for easier mole whacking of eth tests
- Start refactor of handling precompiles
- implement identity, sha256, ripemd160
- fix costOfCall bug for new accounts with no tx value
- increased vmTest timeout from 1 to 10 seconds

### DIFF
--- a/src/hevm/src/EVM/VMTest.hs
+++ b/src/hevm/src/EVM/VMTest.hs
@@ -115,16 +115,21 @@ checkExpectation diff execmode x vm =
   case (execmode, testExpectation x, view EVM.result vm) of
     (EVM.ExecuteAsBlockchainTest, Just expectation, _) -> do
       let
+        printField :: (v -> String) -> Map Addr v -> IO ()
+        printField f d = putStrLn $ Map.foldrWithKey (\k v acc ->
+          acc ++ showAddrWith0x k ++ " : " ++ f v ++ "\n") "" d
+
         ((s1, s1'), (b1, b1')) = (("bad-state", "bad-balance"), checkExpectedContracts vm (expectedContracts expectation))
         ss = map fst (filter (not . snd) [(s1, b1 || b1'), (s1', b1 || not b1')])
+        expected = expectedContracts expectation
+        actual = view (EVM.env . EVM.contracts . to (fmap clearZeroStorage)) vm
       putStr (intercalate " " ss)
       if diff && not b1 then do
-        putStrLn ""
-        putStrLn "Expected postState: "
-        putStrLn $ show (expectedContracts expectation)
-        putStrLn "Actual postState: "
-        putStrLn $ show (vm ^. EVM.env . EVM.contracts . to (fmap clearZeroStorage))
-        else return ()
+        putStrLn "\nExpected balance/state: "
+        printField (\v -> (show . toInteger $ _balance v) ++ " " ++ (show . Map.toList $ _storage v) ) expected
+        putStrLn "\nActual balance/state: "
+        printField (\v -> (show . toInteger $ EVM._balance v) ++ " " ++ (show . Map.toList $ EVM._storage v)) actual
+      else return ()
       return b1
     (_, Just expectation, Just (EVM.VMSuccess output)) -> do
       let
--- a/src/hevm/hevm-cli/hevm-cli.hs
+++ b/src/hevm/hevm-cli/hevm-cli.hs
@@ -368,7 +368,7 @@ runVMTest diffmode execmode mode (name, x) = do
     action <- async $
       case mode of
         Run ->
-          timeout (1e6) . evaluate $ do
+          timeout (1e7) . evaluate $ do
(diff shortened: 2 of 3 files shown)


livnev:
@cmditch I rebased this on top of `master` to make the history clearer, without merges from `master`, and pushed to `cmditch:identity-precompile-rebased`. Do you mind resetting this branch to that?

src/hevm/src/EVM/VMTest.hs
```diff
@@ -115,16 +115,21 @@ checkExpectation diff execmode x vm =
...
-        putStrLn "Expected postState: "
-        putStrLn $ show (expectedContracts expectation)
-        putStrLn "Actual postState: "
-        putStrLn $ show (vm ^. EVM.env . EVM.contracts . to (fmap clearZeroStorage))
-        else return ()
+        putStrLn "\nExpected balance/state: "
+        printField (\v -> (show . toInteger $ _balance v) ++ " " ++ (show . Map.toList $ _storage v) ) expected
+        putStrLn "\nActual balance/state: "
+        printField (\v -> (show . toInteger $ EVM._balance v) ++ " " ++ (show . Map.toList $ EVM._storage v)) actual
+      else return ()
```
livnev: I love this, indeed it makes debugging the tests so much easier!

livnev:
@cmditch any idea why `SHA256` and `RIPMD` are so slow? I noticed when the tests were running that it was noticeably hanging on them, and I assume that's why you bumped the timeout. Are these Haskell implementations just crap?

livnev:
> More tests are correctly failing now. For example a bunch of ecmul and modexp precompiles were passing despite not being implemented.

I believe that's because we were able to get the right behaviour anyway: for example `ecmul_0-0_0_21000_0_d0g0v0.json` is one such test, where the execution reverts since by the time we reach the precompile call there isn't enough gas to do the call (40k is required for `ECMUL`). I agree with you that it seems misleading to mark those as passing, so for now while we work on them we can keep the exceptions at gas-measurement time as you propose.

livnev:
I'm trying to get `ECADD` working on [the ec-precompiles branch](https://github.com/dapphub/dapptools/tree/ec-precompiles).

src/hevm/src/EVM/VMTest.hs
cmditch: Excellent, glad to hear it!

cmditch:
@livnev :

> @cmditch I rebased this on top of `master` to make the history clearer, without merges from `master`, and pushed to `cmditch:identity-precompile-rebased`. Do you mind resetting this branch to that?

Wasn't able to change what branch this PR was pointing at, but just force pushed your `identity-precompile-rebased` over `identity-precompile`. So should be all good now. Thanks for that.

>@cmditch any idea why SHA256 and RIPMD are so slow? I noticed when the tests were running that it was noticeably hanging on them, and I assume that's why you bumped the timeout. Are these Haskell implementations just crap?

I don't think it's the haskell code, but the eth tests themselves. Most of the SHA256/RIPEMD160 tests are blazing fast, but a handful of them are turtles. (For example, try `/BlockchainTests/GeneralStateTests/stQuadraticComplexityTest/Call50000_d0g1v0.json` in debug mode, it starts with ~250 million gas.) Also, I did a comparison of the test times for this branch vs master for only `GeneralStateTests/stPreCompiledContracts2` tests, and they're both around 3 seconds on my machine. (This branch is oddly consistently 100ms faster, despite the new features)

>I believe that's because we were able to get the right behaviour anyway...
I figured they we're false positive "reverts" as well.

livnev:
@cmditch I see, the tests are probably just huge then.

I'm going to try another pass over your precompile handling this week, to see if we can tighten the code up a bit and reduce the amount of duplication.

cmditch:
@livnev Sounds good. Agreed, the code "tightness" could use some love. 

I've held back on making the code pretty while figuring out why the gas calculations in some of the tests fail. My intuition points to how precompiles are handling the 2300 gas call_stipend.

@rainbreak You'd asked in the chat what was wrong with the previous precompile implementation, and the above is part of the issue. Also, how the state is changed on value transfers, and all the variants of zero or non-zero xfers within a CALL, CALLCODE, STATICCALL, or DELEGATECALL. In short, what I saw in the previous implementation "worked" for most cases, but would fail, due to state or gas reasons, for many eth tests.

livnev:
absorbed by #213

PR #202, opened 2019-05-13
Injection of `test-hevm`

icetan:
To be able to build `solidityPackage`s with specific requirements on HEVM version to run tests.

### DIFF
--- a/nix/solidity-package.nix
+++ b/nix/solidity-package.nix
@@ -15,15 +15,18 @@
          xs);
 in
   pkgs.lib.makeOverridable (
-    attrs @ { test ? true, deps ? [], solc ? pkgs.solc, ... }:
+    { doCheck ? true
+    , deps ? []
+    , solc ? pkgs.solc
+    , ...
+    } @ attrs:
       pkgs.stdenv.mkDerivation (rec {
         buildInputs = [pkgs.dapp2.test-hevm solc];
+        inherit doCheck;
         passthru = {
           remappings = remappings deps;
           libPaths = libPaths deps;
         };
-
-        TEST = test;
 
         REMAPPINGS =
           pkgs.lib.mapAttrsToList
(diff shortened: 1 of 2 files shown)


nix/solidity-package.nix
```diff
@@ -15,16 +15,20 @@
...
+    { doCheck ? true
+    , deps ? []
+    , solc ? pkgs.solc
+    , test-hevm ? pkgs.dapp2.test-hevm
+    , ...
+    } @ attrs:
       pkgs.stdenv.mkDerivation (rec {
-        buildInputs = [pkgs.dapp2.test-hevm solc];
+        inherit doCheck;
+        buildInputs = [test-hevm solc];
```
asymmetric: Can you add space after the brackets please?

nix/solidity-package.nix
```diff
@@ -15,16 +15,20 @@
...
         passthru = {
           remappings = remappings deps;
           libPaths = libPaths deps;
         };
 
-        TEST = test;
```
asymmetric: Where was this being read?

nix/solidity-package.sh
```diff
@@ -20,7 +20,10 @@ mkdir lib
 echo "$LIBSCRIPT" > setup.sh
 source setup.sh
 export DAPP_LIB=lib
-dapp2-test-hevm
+
+if [ "$doCheck" == 1 ]; then
```
asymmetric: Just to confirm: `true` is translated to 1? Do you know where this happens?

nix/solidity-package.nix
icetan: Couldn't find it being read anywhere.

nix/solidity-package.nix
icetan: Yes

nix/solidity-package.sh
icetan: Yes, because `doCheck` gets set as an environment variable it's interpolated with `toString` which is `"1"` for `true`.

PR #205, opened 2019-05-13
travis: only push when building upstream

asymmetric:
Avoids trying to push on external PR builds.

Also changes from a second invocation of `cachix push` to a `sleep 10`, mostly because it works with a pipe and a conditional.

The real fix would be if the daemon would [accept signals](https://github.com/cachix/cachix/issues/53).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,8 @@ script:
   - |
     nix-env -i ShellCheck
     grep -RHL --exclude-dir node_modules node src/*/libexec | xargs shellcheck
-  - cachix push --watch-store dapp &
+  - |
+    [[ -z "$TRAVIS_PULL_REQUEST_SLUG" || "$TRAVIS_PULL_REQUEST_SLUG" == "dapphub/dapptools" ]] && cachix push --watch-store dapp &
   - |
     PLATFORM=linux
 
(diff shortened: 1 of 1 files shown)


xwvvvvwx:
What is the difference between this and #206 ?

.travis.yml
```diff
@@ -17,27 +17,29 @@ script:
...
+    [[ "$TRAVIS_OS_NAME" == "osx" ]] && PLATFORM=darwin
 
     # the additional call to `cachix push` is there to prevent a race condition
     # that happens when the above daemon is killed mid-push at the end of this
     # script.
     nix-build -j auto release.nix -A dapphub.$PLATFORM.stable | cachix push dapp
 after_success:
   - |
-    COMMIT_HASH=$(echo $TRAVIS_COMMIT | cut -c 1-8)
+    if [[ "$TRAVIS_REPO_SLUG" != "dapphub/dapptools" ]]; then
```
xwvvvvwx: Is this check the right way round? I feel like it should match the check above? Or am I misunderstanding?
```suggestion
    if [[ "$TRAVIS_REPO_SLUG" -= "dapphub/dapptools" ]]; then
```

.travis.yml
```diff
@@ -17,27 +17,29 @@ script:
...
   - |
     PLATFORM=linux
 
-    if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then PLATFORM=darwin; fi
+    [[ "$TRAVIS_OS_NAME" == "osx" ]] && PLATFORM=darwin
 
     # the additional call to `cachix push` is there to prevent a race condition
     # that happens when the above daemon is killed mid-push at the end of this
     # script.
     nix-build -j auto release.nix -A dapphub.$PLATFORM.stable | cachix push dapp
```
xwvvvvwx: Do we need to check the `REPO_SLUG` for this push as well?

.travis.yml
asymmetric: You're right, was distracted when I did this :)

.travis.yml
asymmetric: Yep.

asymmetric:
I don't understand why these commits aren't being built.

asymmetric:
Oh I see, there are issues with the `.travis.yml`: https://travis-ci.com/dapphub/dapptools/requests

Issue #121, opened 2019-02-02
ethsign needs updating


nanexcool:
1. There's a lot of repeated code, refactor some into functions
2. Update to more recent go-ethereum libraries

PR #212, opened 2019-05-18
ethsign: factor out repetitions

MrChico:
Just a minor clean up. Addresses (part of) https://github.com/dapphub/dapptools/issues/121

### DIFF
--- a/src/ethsign/ethsign.go
+++ b/src/ethsign/ethsign.go
@@ -41,6 +41,109 @@ func rawSignHash(data []byte) []byte {
   return crypto.Keccak256([]byte(msg))
 }
 
+func getWallets(c *cli.Context, defaultKeyStores cli.StringSlice) []accounts.Wallet {
+  backends := []accounts.Backend{}
+
+  var paths []string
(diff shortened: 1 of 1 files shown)


MrChico:
cc @nanexcool

asymmetric:
Have you tested manually that it works as intended?

We should have unit tests for this obviously.

MrChico:
Tried the error propagation and the actual signing with some files in `keystore` which works. Did not try with any hardware wallets, but it should not be a problem

nanexcool:
This is not working for me on Ubuntu 18 with a Ledger. Then again... neither is master, so will have to investigate further.

nanexcool:
I'd suggest merging this, then figure out why Ledger doesn't work. Was looking at go-ethereum code yesterday, and I'm pretty sure they straight up do not support message signing using Ledger. 

https://github.com/ethereum/go-ethereum/blob/7c91038bff21b9085f61108d75716e2d57cdca8b/accounts/usbwallet/wallet.go#L514

PR #213, opened 2019-05-19
hevm: add SHA256, RIPEMD, IDENTITY, MODEXP, ECADD, ECMUL, ECPAIRING precompiles

livnev:
and also fixes some precompile-related bugs, refactors precompile handling.

**GeneralStateTests status**: 6015 passed, 3782 bad-balance, 815 failed.

Thanks @cmditch 😊.  This is now ready to merge IMO.

### DIFF
--- a/src/hevm/hevm.cabal
+++ b/src/hevm/hevm.cabal
@@ -1,3 +1,4 @@
+cabal-version: 2.2
 name:
   hevm
 version:
@@ -12,7 +13,7 @@ description:
 homepage:
   https://github.com/dapphub/dapptools
 license:
-  AGPL-3
+  AGPL-3.0-only
 license-file:
   COPYING
 author:
@@ -23,8 +24,6 @@ category:
   Ethereum
 build-type:
   Simple
-cabal-version:
-  >=1.10
 data-files:
   run-consensus-tests
 extra-source-files:
@@ -64,9 +63,11 @@ library
   ghc-options:
     -Wall -Wno-deprecations
   extra-libraries:
-    secp256k1
+    secp256k1, ff
   c-sources:
     ethjet/tinykeccak.c, ethjet/ethjet.c
+  cxx-sources:
+    ethjet/ethjet-ff.cc
   install-includes:
     ethjet/tinykeccak.h, ethjet/ethjet.h
   build-depends:
--- a/src/hevm/hevm.cabal
+++ b/src/hevm/hevm.cabal
@@ -68,6 +68,8 @@ library
     ethjet/tinykeccak.c, ethjet/ethjet.c
   cxx-sources:
     ethjet/ethjet-ff.cc
+  cxx-options:
+    -std=c++0x
   install-includes:
     ethjet/tinykeccak.h, ethjet/ethjet.h, ethjet/ethjet-ff.h
   build-depends:
--- a/src/hevm/hevm.cabal
+++ b/src/hevm/hevm.cabal
@@ -69,7 +69,7 @@ library
   cxx-sources:
     ethjet/ethjet-ff.cc
   install-includes:
-    ethjet/tinykeccak.h, ethjet/ethjet.h
+    ethjet/tinykeccak.h, ethjet/ethjet.h, ethjet/ethjet-ff.h
   build-depends:
     QuickCheck                        >= 2.9,
     abstract-par                      >= 0.3,
--- a/src/hevm/src/EVM.hs
+++ b/src/hevm/src/EVM.hs
@@ -1062,8 +1062,7 @@ precompiledContract vm fees gasCap precompileAddr recipient xValue inOffset inSi
                 case stk of
                   (0:_) ->
                     return ()
-                  (1:_) ->
-                    touchAccount recipient $ \_ -> do
+                  (1:_) -> do
                     zoom (env . contracts) $ do
                       ix self . balance -= xValue
                       ix recipient  . balance += xValue
--- a/src/hevm/src/EVM.hs
+++ b/src/hevm/src/EVM.hs
@@ -1406,7 +1406,11 @@ finalize txmode = do
               (Map.delete createe)
           Just (VMSuccess output) -> do
             createe <- use (state . contract)
-            replaceCode createe (RuntimeCode output)
+            createeExists <- (Map.member createe) <$> use (env . contracts)
+            if createeExists then
+              replaceCode createe (RuntimeCode output)
+              -- don't deploy code when createe selfdestructed
+              else return ()
           Nothing -> error "Finalising an unfinished tx."
       modifying (env . contracts)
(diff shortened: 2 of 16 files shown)


src/hevm/hevm.cabal
```diff
@@ -12,7 +13,7 @@ description:
 homepage:
   https://github.com/dapphub/dapptools
 license:
-  AGPL-3
+  AGPL-3.0-only
```
rainbreak: Is this automated?

src/hevm/hevm.cabal
```diff
@@ -1,3 +1,4 @@
+cabal-version: 2.2
```
rainbreak: What about 2.2 specifically? And why no >=?

src/hevm/src/EVM.hs
```diff
@@ -1027,53 +1030,174 @@ exec1 = do
...
+
+        -- IDENTITY
+        0x4 -> do
+            assign (state . stack) (1 : xs)
+            assign (state . returndata) input
+            copyBytesToMemory input inSize 0 outOffset
+            next
+
+        -- EXPMOD
+        0x5 -> notImplemented
```
rainbreak: When implemented?

src/hevm/src/EVM.hs
```diff
@@ -1895,19 +2023,21 @@ mkCodeOps bytes = RegularVector.fromList . toList $ go 0 bytes
 -- Gas cost function for CALL, transliterated from the Yellow Paper.
 costOfCall
   :: FeeSchedule Word
-  -> Maybe a -> Word -> Word -> Word
+  -> Bool -> Word -> Word -> Word
   -> (Word, Word)
-costOfCall (FeeSchedule {..}) recipient xValue availableGas xGas =
+costOfCall (FeeSchedule {..}) recipientExists xValue availableGas xGas =
```
rainbreak: What was wrong with Maybe?

rainbreak:
Will wait for +1 from @cmditch. Also would be nice if Travis wasn't broken, but that's unrelated.

src/hevm/hevm.cabal
livnev: not sure, I guess the cabal file was regenerated in commit 27dbc97c0bbba34907c115abf5e9cea4821fa019

src/hevm/hevm.cabal
livnev: see 27dbc97c0bbba34907c115abf5e9cea4821fa019. We could probably revert the cabal file to how it was, leaving just the `libff` related changes.

src/hevm/src/EVM.hs
livnev: `costOfCall` used to take a `Maybe Contract` but only checked `isNothing`, I think this was sloppy (and sometimes burdensome to actually produce a term of that type) so I changed it to take a `Bool`.

livnev:
> What's the before/after test suite score?

@rainbreak the last change that affected the VM was merged in #145. Note that this PR not only implements the missing precompiles, but also:

(i) @cmditch fixed a bug with call gas costs
(ii) I fixed the handling of selfdestruct during initcode
(iii) There were some `BlockchainTests` which we were failing to convert into actual test cases. For instance, some of them had no transactions to test. I enquired about the meaning of these, and determined that they were safe to skip, so they are now marked as `ok`.

So we have gone from:
5080 passed, 4000 bad balance, 1532 failed
to
6015 passed, 3782 bad-balance, 815 failed

src/hevm/src/EVM.hs
livnev: next weekend 😹

livnev:
> Will wait for +1 from @cmditch. Also would be nice if Travis wasn't broken, but that's unrelated.

Actually Travis was reporting a relevant failure: one of the commits that changed `haskell.nix` was bad. I have fixed it and force pushed.

livnev:
@rainbreak also worth warning that [this issue](https://github.com/scipr-lab/libff/issues/15) will mean that it won't build on older MacOS. Do you think we should do something about that?

livnev:
```
  ecrecover
    Example 1:                Query <EVM.Query: fetch contract 0000000000000000000000000000000000000001>
FAIL (0.03s)
      test/test.hs:206:
      bytes32 h = keccak256(abi.encodePacked("\x19Ethereum Signed Message:\n32", d));
      x = ecrecover(h, a, b, c);
      
      expected: Just [0,0,0,0,0,0,0,0,0,0,0,0,45,94,86,212,92,99,21,13,147,127,33,130,83,138,15,24,81,12,177,31]
       but got: Nothing
```
curious, it looks like this has started to fail for some reason, I'll investigate. I wasn't running these tests locally before.

src/hevm/src/EVM.hs
```diff
@@ -1027,53 +1030,173 @@ exec1 = do
...
+  => VM
+  -> FeeSchedule Word
+  -> Word
+  -> Addr
+  -> Addr
+  -> Word
+  -> Word -> Word -> Word -> Word
+  -> [Word]
+  -> EVM ()
+precompiledContract vm fees gasCap precompileAddr recipient xValue inOffset inSize outOffset outSize xs =
```
cmditch: Good call parameterizing everything, better without that previous case statement for the call type imo.

cmditch:
@livnev Getting some build errors on the C parts in macOS.
Error output: https://gist.github.com/cmditch/bff97daf634e5f1c6b0ac75d4bd6775a

livnev:
@cmditch I think I finally got it to build on macOS (on travis). The secret was to use `cxx-options` I think. Let me know if it now builds on your machine as well.

cmditch:
Compilation works!

I noticed a regression in the tests that send eth to the precompile address, e.g. this test used to pass:
image

As you said above, you changed this here https://github.com/dapphub/dapptools/commit/b9a6025f68b6a8a2a95cf9554fca504211fbc42a in order to get the Haskell tests working.

livnev:
@cmditch Good to hear that it builds.

I think it's right that it's marked as `bad-balance`: `bad-balance` is for when the states are equal but only if you ignore the balances. Thanks for spotting this regression, of course `touchAccount` is actually the right thing to do, otherwise the ether sent just goes missing since the recipient account never gets initialised. I am considering putting it back in and instead just removing that Haskell test, since I haven't been able to figure out how to properly initialise the state in those tests yet.

livnev:
@rainbreak imo this can be merged now.

nanexcool:
This PR is amazing :heart:

rainbreak:
@cmditch does `stack build` still work for you with this?

livnev:
This now includes the last missing precompile, `MODEXP`.

rainbreak:
Merging. Thanks everyone :) 

@livnev can you make a new issue for the remaining modexp failures?

PR #216, opened 2019-05-27
dapp: faster builds with single solc pass

rainbreak:
N.B. breaking change as .abi .bin and .bin-runtime artefacts are no longer generated.

Pass all of the source files to solc at once for much faster compilation. Previously we iterated through each file, which isn't necessary with the solc json output.

A positive side effect is no more test duplication in `dapp test`, which can occur with more complex projects (see e.g. https://github.com/makerdao/dss).

TODO:
  - [x] fix shellcheck issue
  - [x] if needed, extract .abi .bin .bin-runtime from the json
  - [x] ~decide on `--legacy` vs `--fast` flag~ --extract as optional
  - [x] update `dapp debug`

### DIFF
--- a/src/dapp/CHANGELOG.md
+++ b/src/dapp/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.23.0] - 2019-07-08
+### Changed
+- Faster `dapp build`, with a single compilation pass. Consolidates solc
+  output into a single `out/dapp.sol.json`, which may break some
+  workflows. `dapp build --extract` can be used to extract .abi, .bin,
+  .bin-runtime files from the json into `out/`.
+- The version of `hevm` is now shown in `dapp --version`.
+
 ## [0.22.0] - 2019-06-21
 ### Changed
 - geth upgraded to 1.8.27
@@ -112,3 +120,4 @@ changelog.
 [0.20.0]: https://github.com/dapphub/dapptools/tree/dapp/0.20.0
 [0.21.0]: https://github.com/dapphub/dapptools/tree/dapp/0.21.0
 [0.22.0]: https://github.com/dapphub/dapptools/tree/dapp/0.22.0
+[0.23.0]: https://github.com/dapphub/dapptools/tree/dapp/0.23.0
--- a/src/dapp/libexec/dapp/dapp-build
+++ b/src/dapp/libexec/dapp/dapp-build
@@ -1,10 +1,43 @@
 #!/usr/bin/env bash
 ### dapp-build -- compile the source code
-### Usage: dapp build
+### Usage: dapp build [--extract]
+###
+###  --extract: After building, write the .abi, .bin and .bin-runtime
+###             files from the solc json into $DAPP_OUT. Beware of contract
+###             name collisions. This is provided for compatibility with older
+###             workflows.
 set -e
 
+while [[ $# -gt 0 ]]; do
+  opt=$1; shift
+  case $opt in
+    --extract)
+      DAPP_BUILD_EXTRACT=1
+      ;;
+    *)
+      info "unknown argument: $opt"
+      exit 1
+      ;;
+  esac
+done
+
+info() {
+  while [[ $# -gt 0 ]]; do
+    opt=$1; shift
+    case $opt in
+      -*)
+        FLAGS=$opt;
+        ;;
+      *)
+        ARGS=$opt;
+        ;;
+    esac
+  done
+  echo >&2 "$FLAGS" "${0##*/}: $ARGS"
+}
+
 IFS=" " read -r -a opts <<<"$SOLC_FLAGS"
-json_opts=--combined-json=abi,bin,bin-runtime,srcmap,srcmap-runtime,ast
+json_opts=--combined-json=abi,bin,bin-runtime,srcmap,srcmap-runtime,ast,metadata
 solc --help | grep -q -- --overwrite && opts+=(--overwrite)
 
 remappings=$(dapp remappings)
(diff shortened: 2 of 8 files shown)


rainbreak:
@gbalabasquer any thoughts on the .bin .bin-runtime .abi files? Do you know if these being relied on?

MrChico:
`seth send --create` usually requires the `.bin`, it might be nice to do the `jq` wrapping needed if  `--create` would rely on the `sol.json` instead

gbalabasquer:
@rainbreak we use the `.abi` in the deployments scripts for example. But I think eventually, if this changes is gonna happen, we could adapt to get the abis from this unique file.
Apart from the most important case (the `seth send --create`) I remember we have the etherscan verify functionality which I think relies in these files. cc @nanexcool 
I think, once we take care to fix all the internal dapptools commands which rely on them, we could have a new `dapp` command that generates these files on demand when needed for external applications.

gbalabasquer:
Maybe not even really needed to have the command. Just copying and pasting the `jq` filter to extract the needed content could be enough for external application.
However if we want to be backward compatible what we could do is keeping all the internal commands getting the content from this unique file but also generating the other ones (excepting if you use some flag like `--fast`). In the future we can deprecate the automatic generation of the extra files.

nanexcool:
This is really fast compared to the previous version. We'll need to revisit anything that uses `bin` and `bin-runtime` and metadata files.

As for verification, we can just add `metadata` to the `json_opts` so:

```
json_opts=--combined-json=abi,bin,bin-runtime,srcmap,srcmap-runtime,ast,metadata
```
and verification can grab data from there just as it would every other file.

rainbreak:
I think I'd prefer if data were extracted from the json by the programs that require it, possibly with some new dapp commands to help, but one option is to do the extraction inside `dapp build` and leave this data in files as-is.

rainbreak:
@nanexcool can you link to somewhere that the metadata is actually used?

nanexcool:
I use it to verify contracts, although I haven't used the functionality in a while. Don't know if Etherscan changed the API

https://github.com/dapphub/dapptools/blob/master/src/dapp/libexec/dapp/dapp-verify-contract#L31

nanexcool:
If it adds too much cruft to the json file, we can just omit the metadata, and I'll rewrite Etherscan contract verification to extract metadata on demand, it's no biggie and it shouldn't block this PR.

rainbreak:
I added a `--legacy` flag to `dapp build` to enable the extraction of the data files. Thoughts?

@nanexcool I added a fix for `dapp verify-contract`, but didn't test it fully. One difference is that now you'd have to provide the arguments like `dapp verify-contract "src/vat.sol:Vat" 0xabc...`.

rainbreak:
`--legacy` flag was changed to `--extract`. The default is to not extract the build data.

rainbreak:
@gbalabasquer pls review.

gbalabasquer:
LGTM

asymmetric:
Pls add to `CHANGELOG`.

src/dapp/libexec/dapp/dapp-build
```diff
@@ -31,12 +59,24 @@ else
...
+    info -ne "Extracting build data... [$count/$total]\\r"
+    ((count++))
+    name="${path#*:}"
+    contract=$(echo "$data" | jq '.[''"'"$path"'"'']')
+    echo "$contract" | jq '.["abi"]' -r > "$DAPP_OUT/$name.abi"
+    echo "$contract" | jq '.["bin"]' -r > "$DAPP_OUT/$name.bin"
+    echo "$contract" | jq '.["bin-runtime"]' -r > "$DAPP_OUT/$name.bin-runtime"
+  done
+  echo
+  info "Used --extract, build data can be extracted from $DAPP_JSON for faster builds!\\n"
```
asymmetric: I don't understand what this means: `--extract` extracts right? So what does it mean that one can further extract for faster builds?

src/dapp/libexec/dapp/dapp-build
```diff
@@ -3,8 +3,36 @@
...
+      DAPP_BUILD_EXTRACT=1
+      ;;
+    *)
+      info "unknown argument: $opt"
+      exit 1
+      ;;
+  esac
+done
+
+info() {
```
asymmetric: Nice!

src/dapp/libexec/dapp/dapp-verify-contract
```diff
@@ -28,14 +28,17 @@ case "$chain" in
     exit 1
 esac
 
-meta=$(cat "$DAPP_OUT"/"${1?contractname}"_meta.json)
+path=${1?contractname}
```
asymmetric: What are the changes in this file doing?

src/dapp/CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
...
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.23.0] - 2019-07-05
+### Changed
+- Faster `dapp build`, with a single compilation pass. Consolidates solc
+  output into a single `out/dapp.sol.json`, which may break some
+  workflows. `dapp build --extract` can be used to extract .abi, .bin,
+  .bin-runtime files from the json into `out/`.
+- The version of `hevm` is now shown in `dapp --version`.
+
```
asymmetric:
```suggestion
### Removed
- Support for solc 0.4.24 on macOS, since it wasn't working
```

:)

asymmetric:
Can you document `--extract` in `dapp-build`?

src/dapp/libexec/dapp/dapp-build
rainbreak: It means that it's faster to not use `--extract`. Rather, users would ideally get the specific things that they need from the .sol.json.

src/dapp/libexec/dapp/dapp-verify-contract
rainbreak: According to @nanexcool this is broken anyway. This is just a best effort try to update it to this new build format. It should be properly fixed in another pr.

src/dapp/libexec/dapp/dapp-build
asymmetric: I think this message doesn't make a lot of sense. Users have to opt-in to the slower behaviour, so will probably have executed without `--extract` before trying it, and will be able to see the speed difference by themselves.

rainbreak: So just have no message?

nanexcool:
Enough Gandalfing, let's merge this!! :heart:

PR #217, opened 2019-05-27
[WIP] implement MODEXP precompile

### LINKED ISSUE #189: hevm: implement MODEXP (precompile no. 5)

livnev:
resolves #189 

TODOs: 
- [x] use more generic bytes conversions
- [x] improve performance
- [ ]  fix ~86~ 33 failing tests
- [ ] fix 7 tests which run out of memory


MrChico:
I think we should just add a changelog entry and then we should be all good

MrChico:
Actually, it would be great if `modexp` actually returned the correct result as well

livnev:
subsumed into #213

PR #218, opened 2019-05-28
hevm: add "e" for stepping to end in interactive

livnev:
By popular demand: also made is so that advancing past the end of execution just blocks instead of halting `hevm`. To exit `hevm` you have to press `Esc`.


livnev:
cc @lucasvo

lucasvo:
👍

PR #219, opened 2019-05-28
Add `solcFlags` argument to `solidityPackage` that passes flags to `solc`

icetan:
To be able to set e.g. `--optimize` flag when using `solidityPackage` Nix expression to build solidity contracts.

This is equivalent to setting `SOLC_FLAGS` for `dapp build`.

### DIFF
--- a/nix/solidity-package.sh
+++ b/nix/solidity-package.sh
@@ -1,6 +1,7 @@
 source $stdenv/setup
 unpackPhase
 
+IFS=" " read -r -a opts <<<"$solcFlags"
 jsonopts=--combined-json=abi,bin,bin-runtime,srcmap,srcmap-runtime,ast
 
 export DAPP_SRC=$src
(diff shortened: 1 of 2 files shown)


nix/solidity-package.sh
```diff
@@ -1,6 +1,7 @@
 source $stdenv/setup
 unpackPhase
 
+IFS=" " read -r -a opts <<<"$solcFlags"
```
asymmetric: Who defines `solcFlags`? Does it make sense to convert it to an array? We could simply pass it to `solc` as a string.

icetan: It would be set as an implicit argument to `solidityPackage`. Maybe add `solcFlags ? ""` to make it clear that it could be set.

It could definitely be just a regular string.

nix/solidity-package.sh
```diff
@@ -1,6 +1,7 @@
 source $stdenv/setup
 unpackPhase
 
+IFS=" " read -r -a opts <<<""
```
asymmetric: What's this about?

icetan: fixed

PR #220, opened 2019-05-31
Fix stack.yaml to build with GHC 8.4

japesinator:
I would have updated to build with GHC 8.6, but that has problems with quickcheck's `Gen` not having a `MonadFail` instance

### DIFF
--- a/src/hevm/stack.yaml
+++ b/src/hevm/stack.yaml
@@ -1,41 +1,27 @@
-resolver: lts-8.15
+resolver: lts-12.10
 packages:
 - .
 extra-deps:
-  - HSH-2.1.3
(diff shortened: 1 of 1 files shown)


rainbreak:
Can you explain how you're able to drop a bunch of dependencies? Does the resolver figure them out somehow?

ggrieco-tob:
I'm not an expert in the stack resolver, but I believe that the use of a more recent resolver (`12.10`) avoids to set the version of the dependencies (that otherwise are too old in `8.15`)

japesinator:
`extra-deps` aren't dependencies, they're packages that aren't handled by the resolver by default. The actual dependencies are tracked in `hevm.cabal`

PR #225, opened 2019-06-15
Use release-19.03 as base

asymmetric:
This means we're using `release-19.03` at commit [`f1707d8875276cfa110139435a7e8998b4c2a4fd`](https://github.com/NixOS/nixpkgs/tree/f1707d8875276cfa110139435a7e8998b4c2a4fd) when pushing to cachix.


rainbreak:
Lgtm

PR #228, opened 2019-06-20
launch geth with nohup

asymmetric:
Allows the script to be sent to the background, since geth needs nohup
in order to be sent to the background (see [here](0)).

[0]: https://github.com/ethereum/go-ethereum/issues/16307


asymmetric:
WIP: this still doesn't handle geth's stdout properly. It's currently being redirected to a file (nohup's default behaviour), whereas we want it to be printed to the screen.

asymmetric:
Needs to be investigated further: `dapp testnet` running in the background fails to be recognized by the setup scritps, and therefore `ETH_FROM` does not get set, breaking everything.

PR #230, opened 2019-06-21
nix: clean up solc-versions.nix

asymmetric:
Instead of adding versions for all platforms to `x86_64-darwin`, they can now be added to `common`.


rainbreak:
This is much clearer!

PR #237, opened 2019-07-03
dapp: show hevm version in --version

rainbreak:


asymmetric:
Could've actually been nice to add it to the `CHANGELOG`.

rainbreak:
> Could've actually been nice to add it to the `CHANGELOG`.

done in #216

PR #240, opened 2019-07-05
use ethers.js for JSON ABI blob generation

### LINKED ISSUE #82: `seth` doesn't work with `solc 0.5.0`

asymmetric:
How to test:
```
nix-build -A seth
result/bin/seth abi 'execute(bytes)'
result/bin/seth --abi-function-json "DSSpell(address,uint256,bytes)"
result/bin/seth --abi-decode "foo(uint256)(address)" 0000000000000000000000000000000000000000000000000000000000000001
```

Issues:

* ~ethers.js produces annoying warnings we don't care about~
* ~6.9 MB of stuff in `src/seth/libexec/seth/node_modules`~

Closes #82.


src/seth/CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.8.2] - 2019-07-05
+### Fixed
+ - ABI JSON blob creation would fail when function arguments required specifying
```
asymmetric: This could be improved, looking for poetic help.

livnev: How about: 

> ABI JSON blob creation used to fail when function arguments had to be specified

asymmetric: The sentence wasn't finished :)

An improvement could be "when function arguments were values of reference type`

rainbreak: - ABI JSON blob creation would fail for some solc reference types, e.g. `bytes memory`, since solc 0.5.0. solc was removed as a dependency and replaced with ethers.js

asymmetric:
The test cases in the first comment all behave as expected, but I'd be happy if others could explore other paths through which `seth---abi` gets called, in case I've missed any.

The main difference from solc's output (which is also what ethabi expects) is that ethers.js wasn't enclosing the JSON object in an array.

rainbreak:
@reverendus maybe interested

PR #241, opened 2019-07-08
dapp-address: remove js dependency

rainbreak:


src/dapp/libexec/dapp/dapp-address
```diff
@@ -1,19 +1,25 @@
-#!/usr/bin/env node
-/*
+#!/usr/bin/env bash
 ### dapp-address -- determine address of newly generated contract
-### Usage: dapp address
-*/
+### Usage: dapp address <creator> <nonce>
+set -ex
```
asymmetric: Do we need the `x`?

src/dapp/libexec/dapp/dapp-address
```diff
@@ -1,19 +1,25 @@
...
+[[ $# -ne 2 ]] && echo "Usage: dapp address <creator> <nonce>" >&2 && exit 1
```
asymmetric: Think we use the `dapp-address` format, i.e. with a hyphen.

PR #247, opened 2019-07-21
Hevmfixes

MrChico:
Increases coverage by a whole lot. If Mr. @rainbreak can tell me what `--skip` pattern will cause GeneralStateTests not to crash I could tell you the exact figures.


rainbreak:
--skip modexp

PR #250, opened 2019-08-07
hevm: correctly handle c++ exceptions on darwin - fixes #248

### LINKED ISSUE #248: hevm: some MODEXP calls fail with libc++abi.dylib: terminating with uncaught exception of type int
Running any code that triggers the `modexp` precompile leads to:
```
hevm bc-test --file ~/dapphub/klab/evm-semantics/tests/ethereum-tests/BlockchainTests/GeneralStateTests/stCreate2/create2callPrecompiles_d5g0v0.json
create2callPrecompiles_d5g0v0_ConstantinopleFix libc++abi.dylib: terminating with uncaught exception of type int
[1]    83906 abort      hevm bc-test --file
```

desaperados:


src/hevm/hevm.cabal
```diff
@@ -150,6 +150,8 @@ executable hevm
     -Wall -threaded -with-rtsopts=-N
   other-modules:
     Paths_hevm
+  if os(darwin)
+    ld-options: -Wl,-keep_dwarf_unwind
```
rainbreak: Can you explain a bit more?

desaperados: Yes, how to summarise - I spent a bit of time chasing this around :-) The issue turned out to be with some broken c++ exception handling on darwin, triggered by the try/catch behaviour in ethjet-ff.cc. These options are link time optimisation flags which allow fallback to dwarf exception tables, overriding the default GHC -no_compact_unwind which seems to cause the broken behaviour.

rainbreak: Wow ok.

rainbreak:
Tests aren't being rerun, as hevm-compliance is being taken from the cache. Maybe the change to hevm.cabal isn't getting picked up by nix somehow. @desaperados did you rebuild with nix or just with cabal when you tested locally?

rainbreak:
Oh I think it's fine because all the darwin tests are passing now.

tabathap:
👌

PR #251, opened 2019-08-08
ethsign: add import command

mbrock:
This command reads a hexadecimal private key and a desired passphrase
from stdin and saves a corresponding keyfile into the keystore using
Geth's standard options and file name scheme.


asymmetric:
Won't be able to test this anytime soon so can't really 👍, but I do have one req: can you add the changes to the CHANGELOG?

mbrock:
@asymmetric cool I forgot there was such a file, will do!

PR #254, opened 2019-08-16
hevm: tty memory pager - fixes #164

### LINKED ISSUE #164: hevm: scrollable memory view
AFAIK it's not currently possible to scroll the memory view.

/cc @rainbreak

desaperados:
added key-bindings help (h)
added memory pager view (M)
moved event handlers to a function

### DIFF
--- a/src/hevm/CHANGELOG.md
+++ b/src/hevm/CHANGELOG.md
@@ -5,6 +5,8 @@
  - `hevm exec` ergonomics:
   - Allow code/calldata prefixed with 0x
   - create transactions with specific caller nonce
+  - interactive help pane
+  - memory pane scrolling
 
 ## 0.33 - 2019-08-06
  - Full compliance with the [General State Tests][245] (with the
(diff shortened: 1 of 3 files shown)


rainbreak:
Does the pager persist the location of the view when toggled? So that in a large view I don't lose my scrolling position.

rainbreak:
Hmm good point. I was thinking of the basic case of within a frame. In a new frame it should reset, but when you pop back to a frame it could persist. However that's likely be quite annoying to implement?

desaperados:
This sounds like an argument for implementing scrolling inside the memory pane itself rather than an expanded view where scrolling position becomes an issue. I was thinking that using an expanded view would be cleaner, and allow sharing of event handlers etc. but maintaining context could actually be less problematic.

desaperados:
I've pushed a version which implements scrolling in the memory pane itself - if we later find that that we later need scrolling in other panes can do that too - would just need to implement some kind of scrolling context switcher.

rainbreak:
Yes, let's play around with it and see. This is a huge improvement :)

PR #256, opened 2019-08-19
hevm: fix dapp-test VM errors

desaperados:


desaperados:
Skipping the entire `finalize` function avoids any VM errors when running dapp-test, although it's actually only the [revert case](https://github.com/dapphub/dapptools/blob/hevm-fix-dapptest/src/hevm/src/EVM.hs#L1393) which causes the errors.  

Given that `exec` shares the same branching logic I wonder if we can leave as is or need to consider an `executeAsDappTest` mode.

PR #259, opened 2019-08-20
Support ABIv2 in HEVM

incertia:
I am porting over the code used to support ABIv2 from [echidna](https://github.com/crytic/echidna/blob/master/lib/Echidna/ABIv2.hs). If possible, I would like some guidance on how to better write `readJSON` without passing around an extra `Maybe` parameter in its subroutines.

### DIFF
--- a/src/hevm/src/EVM/Solidity.hs
+++ b/src/hevm/src/EVM/Solidity.hs
@@ -305,7 +305,7 @@ signature abi =
         ")"
       ]
 
--- Helper funcntion to convert the fields to the desired type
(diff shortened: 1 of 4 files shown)


src/hevm/src/EVM/Solidity.hs
```diff
@@ -305,11 +305,19 @@ signature abi =
         ")"
       ]
 
+-- Helper funcntion to convert the fields to the desired type
```
MrChico: typo

PR #261, opened 2019-08-22
hevm: update metadata hash to sol >=0.5.9

MrChico:
Updates the metadata format expected by hevm to be what's in 0.5.9 and later.

This is a quickfix, which will BREAK THE SOURCEMAPPING FOR OLDER SOLC VERSIONS IN THE SAME WAY THAT CURRENT HEVM IS BROKEN FOR NEW SOLC VERSIONS.

### DIFF
--- a/src/hevm/src/EVM/Solidity.hs
+++ b/src/hevm/src/EVM/Solidity.hs
@@ -345,8 +345,8 @@ stripBytecodeMetadata bs =
 
 bzzrPrefix :: ByteString
 bzzrPrefix =
-  -- a1 65 "bzzr0" 0x58 0x20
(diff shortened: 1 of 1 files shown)


livnev:
Great find, shall we meditate on this to see if we want to try come up with a more general solution?

rainbreak:
I think we need the more general solution, shouldn't be too hard.

MrChico:
We should. 

We should also investigate why it doesn't show `<unknown contract/source>` instead of just showing the wrong sourcemapping

livnev:
For now I am proposing this backwards-compatible but not future-proof solution, where we simply iterate over the known bzzr prefixes. I considered a more principled approach but it seems trickier than expected and would likely still require us to make some assumptions that might be violated in the future, see the commit message there for details.

PR #262, opened 2019-08-23
hevm: handle rpc fetch in tty - fixes #257

### LINKED ISSUE #257: the --rpc flag in hevm interactive
From what I can tell, RPC fetching doesn't currently work in `hevm interactive`. For example, using [this dapp](https://github.com/makerdao/mkr-authority/tree/hevm) and running

```
dapp build && DAPP_TEST_TIMESTAMP=$(seth block latest timestamp) DAPP_TEST_NUMBER=$(seth block latest number) DAPP_TEST_ADDRESS=0x8EE7D9235e01e6B42345120b5d270bdB763624C7 hevm dapp-test --rpc=$ETH_RPC_URL --json-file=out/dapp.sol.json
```

sees all tests pass (as you would expect if the state is being fetched), however replacing `dapp-test` with `interactive` and inspecting `testChangeOwners()` makes the test fail, and the cause of the failure is an attempted call into a non-existent contract, so it looks like the fetcher is not working in this case. A cursory examination of `hevm-cli.hs` and `Fetch.hs` didn't turn up an obvious reason why the RPC isn't being called.

cc @rainbreak

desaperados:
A minimal fix for #257 where remote contracts are now correctly
fetched when running with the --rpc flag. Attempts to fetch
non-existant contracts will result in a VMFailure.

The fix is somewhat unsatisfactory because the tty interpreter
is not suited to gracefully handling operational failures without a
more extensive refactor of the stepper and other interpreters.

In the meantime rpc fetch is functional, with more robust and
informative error handling to follow.

### DIFF
--- a/src/hevm/hevm-cli/hevm-cli.hs
+++ b/src/hevm/hevm-cli/hevm-cli.hs
@@ -160,10 +160,16 @@ unitTestOptions cmd = do
 
   params <- getParametersFromEnvironmentVariables
 
+  let
+    testn = testNumber params
+    block = if (0 ==) testn
+       then EVM.Fetch.Latest
+       else EVM.Fetch.BlockNumber testn
+
   pure EVM.UnitTest.UnitTestOptions
     { EVM.UnitTest.oracle =
         case rpc cmd of
-         Just url -> EVM.Fetch.http (EVM.Fetch.BlockNumber (testNumber params)) url
+         Just url -> EVM.Fetch.http block url
          Nothing  -> EVM.Fetch.zero
     , EVM.UnitTest.verbose = verbose cmd
     , EVM.UnitTest.match   = pack $ fromMaybe "^test" (match cmd)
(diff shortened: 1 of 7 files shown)


src/hevm/src/EVM.hs
```diff
@@ -81,10 +81,7 @@ data Error
   | MaxCodeSizeExceeded Word Word
   | PrecompileFailure
 
-instance Show Error
-  where
-    show (Revert msg) = "Revert " ++ showByteStringWith0x msg
```
rainbreak: I think we need this so that revert messages are rendered in hexadecimal, rather than literally?

desaperados: Ah, ok. I reverted to the derived version becasue this was causing an infinite loop. I'll reinstate a working version.

src/hevm/hevm-cli/hevm-cli.hs
```diff
@@ -160,10 +160,14 @@ unitTestOptions cmd = do
 
   params <- getParametersFromEnvironmentVariables
 
+  let block = if (512 ==) (testNumber params)
```
rainbreak: Not sure if we need to persist 512 as the magic number here. Could just use 0?

desaperados: Ah yes - zero is probably a less magic default - wasn't sure if 512 had some significance :-)

PR #264, opened 2019-08-24
hevm: Solidity.hs: strip using old and new bzzr prefixes

livnev:
This version just iterates over the two known bzzr prefixes, and correctly displays source maps for both new and old  `solc` versions, unlike #261. See commit message for details.

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -19,7 +19,7 @@ let
     export PATH=${x.pkgs.hevm}/bin:${x.pkgs.jq}/bin:$PATH
     ${x.pkgs.hevm}/bin/hevm compliance \
       --tests ${ethereum-test-suite x} \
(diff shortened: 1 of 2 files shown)


MrChico:
This doesn't resolve the issue where the sourcemap display doesn't display `<unknown contract >` instead of just the current source, right? I would like to solve that as well

livnev:
> This doesn't resolve the issue where the sourcemap display doesn't display `<unknown contract >` instead of just the current source, right? I would like to solve that as well

@MrChico  I believe this might have fixed it, since previously when the contract didn't contain the expected BZZR string, the result of `stripBytecodeMetadata` was simply `""` and so the computed codehash was `keccak ""` which caused collisions, which I think is the phenomenon you are describing. If so, this is now fixed (see the added `case` expression in this commit). If not, could you find an example where this happens (that isn't fixed by this PR)?

livnev:
@MrChico @rainbreak can we merge?

PR #270, opened 2019-09-01
dapp: correctly load testnet with multiple accounts

### LINKED ISSUE #269: save|load is broken when running testnet with more than 1 account
How to reproduce:

run testnet & save snapshot after finishing:  `dapp testnet --accounts=2 --save=name`
try to load the snapshot later: `dapp testnet --load=name`

-->  throws an error `Fatal: Could not list accounts: invalid account address or index`🙂

desaperados:
fixes #269


manuelpolzhofer:
@desaperados Thanks for the quick fix. 👍  @ilinzweilin and I had that problem on Friday

PR #274, opened 2019-09-13
hevm: remove (arbitrary?) memory bound

MrChico:
Not sure what the reasoning behind this memory bound was in the first place, but it causes hevm to have incorrect behaviour. Here's an example execution where `hevm` behaves differently to geth:
```
hevm exec --code 600063ffffffff52 --gas 351847752663100
OutOfGas 0x14000f0500033 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
```
```
evm --code 600063ffffffff52 --gas 351847752663100 --prestate customgenesis.json  --debug --nomemory run
0x
```

where `customgenesis.json` makes sure that the geth evm operates according to the Petersburg fork:

```
{
  "alloc": {
  },
  "nonce": "0x000000000000002a",
  "difficulty": "0x020000",
  "mixhash": "0x0000000000000000000000000000000000000000000000000000000000000000",
  "coinbase": "0x0000000000000000000000000000000000000000",
  "timestamp": "0x00",
  "parentHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
  "extraData": "0x",
  "gasLimit": "0x2fefd8",
  "config" : { "chainId": 56312,
               "homesteadBlock": 0,
               "EIP150Block": 0,
               "byzantiumBlock": 0,
               "constantinopleBlock": 0,
               "petersburgBlock": 0
             }
}
```

### DIFF
--- a/src/hevm/src/EVM.hs
+++ b/src/hevm/src/EVM.hs
@@ -2303,9 +2303,7 @@ memoryCost FeeSchedule{..} byteCount =
     linearCost = g_memory * wordCount
     quadraticCost = div (wordCount * wordCount) 512
   in
-    if byteCount > exponentiate 2 32
-    then maxBound
-    else linearCost + quadraticCost
+    linearCost + quadraticCost
 
 -- * Arithmetic
 
--- a/src/hevm/src/EVM/Concrete.hs
+++ b/src/hevm/src/EVM/Concrete.hs
@@ -197,10 +197,6 @@ instance FiniteBits Word where
   countLeadingZeros (C _ x) = countLeadingZeros x
   countTrailingZeros (C _ x) = countTrailingZeros x
 
-instance Bounded Word where
-  minBound = w256 minBound
-  maxBound = w256 maxBound
-
 instance Eq Word where
   (C _ x) == (C _ y) = x == y
 
--- a/src/hevm/src/EVM/Types.hs
+++ b/src/hevm/src/EVM/Types.hs
@@ -40,7 +40,7 @@ mkUnpackedDoubleWord "Word512" ''Word256 "Int512" ''Int256 ''Word256
 newtype W256 = W256 Word256
   deriving
     ( Num, Integral, Real, Ord, Enum, Eq
-    , Bits, FiniteBits, Bounded, Generic
+    , Bits, FiniteBits, Generic
     )
 
 newtype Addr = Addr { addressWord160 :: Word160 }


livnev:
@MrChico  Hmm, yeah this bound does indeed seem very arbitrary. We do have the constraint that the gas cost must fit into a `Word`, otherwise we will get an overflow and absurd results, but that kicks in somewhere between 2^137 and 2^138 bytes, much more than where it is currently bounded at. I am assuming that this bound check was just a miscalculation. I think replacing `if byteCount > exponentiate 2 32` with `if byteCount > exponentiate 2 137` might be a good option, even though it is not the perfect bound.

MrChico:
Not sure I follow your math, do you mean `2^126` and `2^127`? 
Having word arithmetic for gas seem pretty scary. I was able to trigger some overflows indeed on this branch. 
Can we not keep gasUsed as an `Int` in intermediate calculations and truncate it to a word if it fits?
I.e. changing the signature of `burn` to:
```
burn :: Int -> EVM () -> EVM ()
```
and throw overflow error if gas doesn't fit into a word?

MrChico:
Arranging things in this way would also make it more convenient if we ever want to change the `gas` variable to a `uint64`, which I suggest we should. Following geth and [this suggested PR](https://eips.ethereum.org/EIPS/eip-1985)

livnev:
> Not sure I follow your math, do you mean `2^126` and `2^127`?
> Having word arithmetic for gas seem pretty scary. I was able to trigger some overflows indeed on this branch.
> Can we not keep gasUsed as an `Int` in intermediate calculations and truncate it to a word if it fits?
> I.e. changing the signature of `burn` to:
> 
> ```
> burn :: Int -> EVM () -> EVM ()
> ```
> 
> and throw overflow error if gas doesn't fit into a word?

@MrChico  No, I did mean between 2^137 and 2^138. The largest number of bytes for which `memoryCost` fits into a word is `246390752428036619822114583017088058268768`.

I'm not sure if making gas an `Int` is appropriate, those are architecture dependent and are only guaranteed to go up to 2^32 - 1 I think, which is far from sufficient. We could use `Integer` but I'm guessing this could have serious performance penalties.

As I see it, the options are:

(i) explicitly check for overflow in the `memoryCost` calculation, e.g. by comparing against this upper bound I calculated above.
(ii) actually using a `uint64` for gas, as in that linked EIP, at the cost of sacrificing "technically correct" semantics, at least until that EIP is merged (though our semantics are already technically incorrect because of the overflow we are discussing, even though it is incorrect only for much larger values.

As I see it, the only canonical constraint on gas is currently the word size (since this is implied by the existence of opcodes which take/produce gas values from the stack), so (i) is the more correct option. Thoughts?

MrChico:
I'd actually lean towards (2) and check for overflows for uint64s. Although the YP does not specify this explicitly, I think we should follow geth (and the proposed EIP) on this one.

I guess the only "real" scenario this could occur would be in a modified testnet. If that testnet would be a geth driven testnet it would check for uint64 overflow.

What is the opinion if Mr @rainbreak

rainbreak:
I like the uint64 gas option. The rest of that EIP seems sensible too.

When updating the GSTs it did appear that > int64 memory sizes only had cursory coverage. `hevm` likely still has errors in this regard, but testing them is somewhat non-physical. For example, `ByteString` is in principle limited to the machine memory limit (though this didn't appear to be exactly the case in testing), so perhaps `LazyByteString` or some derivative should be used for full generality.

MrChico:
Yeah, there might be some real optimizations to be realized here. The code above, `0x600063ffffffffff52`, takes about 5 seconds to execute in both geth and my modified hevm, whereas it runs in milliseconds with parity. I suspect rust might automatically be lazy in this memory allocation. I wonder if it can be engineered to work the same way in geth as well...

MrChico:
Merging this as it's an improvement over the current semantics, making an issue out of the size of the gas instead

PR #279, opened 2019-09-23
Fix dynamic tuples

incertia:
This fixes a mistake I made earlier in the calculation of the size of dynamic tuples.


incertia:
Any progress on this? It's a super quick change.

PR #284, opened 2019-10-09
dapp verify-contract updates/fixes

brianmcmichael:
Etherscan verification with dapptools does not appear to be working at the moment. This PR updates the `dapp-create` and `dapp-verify` scripts to successfully verify a contract at etherscan.

Notable Changes:

* In `dapp-create`, pass the path of the code, instead of the class name, to `dapp-verify`
* Add leading `v` to version string to match etherscan list
* Get the list of available compilers from etherscan and compare with the build compiler, due to [occasional discrepancies](https://github.com/ethereum/solidity/issues/7512) between the C solc release and the JS solc release, we add a block of code to select the best option from the list of available etherscan-compatible compilers.
* fixes bug in contract address string
* cleans up some of the loops related to checking results, since there is always a delay between when the transaction is confirmed and when it is available for verification at etherscan, sleep timeout is moved to the beginning of retry loop so that it should succeed on the first try.
* Similar handling of checkverify status. Loop was removed here as a design decision to keep the script moving, we can replace it if necessary.

### DIFF
--- a/src/dapp/libexec/dapp/dapp-create
+++ b/src/dapp/libexec/dapp/dapp-create
@@ -19,13 +19,12 @@ done
 bin=$DAPP_OUT/${1?missing contract name}.bin
 type=$(seth --abi-constructor <"$DAPP_OUT/$1.abi")
 address=$(set -x; seth send --create "$bin" "${type/constructor/${1##*/}}" "${@:2}")
-# TODO(cmooney): parse this out of $DAPP_JSON file
-path="src/value.sol"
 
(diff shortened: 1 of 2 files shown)


godsflaw:
@livnev ^^^

livnev:
@brianmcmichael @godsflaw Thanks for the PR. This looks good and appears to work. Could you squash the commit history into one or more clean commits, to get rid of development artifacts, etc.?

brianmcmichael:
Hi @livnev

Sorry about the mess. You should be able to do this directly from the github UI.

squash

PR #291, opened 2019-10-31
Nix flatten solidity source

icetan:
- Add `flatten` argument to `solidityPackage` to pass source through `hevm flatten`
- Fix to allow non-`solidity` `pragmas` being passed through `hevm flatten`
- Rename top scoped contracts and structs so not to cause naming conflicts on flatten

### DIFF
--- a/nix/solidity-package.nix
+++ b/nix/solidity-package.nix
@@ -26,7 +26,7 @@ in
     } @ attrs:
       pkgs.stdenv.mkDerivation (rec {
         inherit doCheck;
-        buildInputs = [ test-hevm solc ] ++ (if flatten then [hevm] else []);
+        buildInputs = [ test-hevm solc ] ++ pkgs.lib.optional flatten hevm;
         passthru = {
           remappings = remappings deps;
           libPaths = libPaths deps;
--- a/nix/solidity-package.nix
+++ b/nix/solidity-package.nix
@@ -19,12 +19,14 @@ in
     , deps ? []
     , solc ? pkgs.solc
     , test-hevm ? pkgs.dapp2.test-hevm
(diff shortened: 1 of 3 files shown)


nix/solidity-package.nix
```diff
@@ -19,12 +19,14 @@ in
...
     , test-hevm ? pkgs.dapp2.test-hevm
+    , hevm ? pkgs.hevm
     , solcFlags ? ""
+    , flatten ? false
     , ...
     } @ attrs:
       pkgs.stdenv.mkDerivation (rec {
         inherit doCheck;
-        buildInputs = [ test-hevm solc ];
+        buildInputs = [ test-hevm solc ] ++ (if flatten then [hevm] else []);
```
asymmetric:
```suggestion
        buildInputs = [ test-hevm solc ] ++ pkgs.lib.optional flatten hevm;
```

MrChico:
What's the status of this? @icetan

icetan:
Working but needs a cleanup before proper review.

brianmcmichael:
+1 this update now handles most of the usecases for the Maker contracts, including support for `pragma experimental abiencoderv2`

There is still an issue with some contracts where an imported contract includes contract name conflicts with other imported contracts ([dss-deploy](https://github.com/makerdao/dss-deploy) is the offender I'm talking about), but none of the other flatteners on the market can handle this properly either. The version on etherscan now was manually flattened.

The benefits here are worth adding in now and then tackling the issues we're facing in dss-deploy later, most users aren't going to be verifying contracts that complex right now anyways.

asymmetric:
@icetan what are the main benefits of this PR?

Issue #281, opened 2019-09-26
control derivation path in ethsign


livnev:
It would be great if `rain-hd-path` could be merged.

MrChico:
What's missing?

livnev:
cc @rainbreak

rainbreak:
Hi, just needs rebasing onto current master and perhaps another consideration of the syntax.

livnev:
@mrchico could you help me with this rebase? There is a major conflict in merging in your changes, and I am not familiar at all with them, or with Go.

PR #293, opened 2019-11-06
Martin hd path

MrChico:
https://github.com/dapphub/dapptools/issues/281

### DIFF
--- a/src/ethsign/ethsign.go
+++ b/src/ethsign/ethsign.go
@@ -90,7 +90,7 @@ Scan:
     } else if x.URL().Scheme == "ledger" {
       x.Open("")
       for j := 0; j <= 3; j++ {
-        pathstr := fmt.Sprintf("m/44'/60'/0'/%d", j)
+        pathstr := fmt.Sprintf(c.String("hd-path") + "/%d", j)
         path, _ := accounts.ParseDerivationPath(pathstr)
         y, err := x.Derive(path, true)
         if err != nil {
@@ -170,6 +170,7 @@ func recover(data []byte, sig hexutil.Bytes, noPrefix bool) (common.Address, err
 }
 
 func main() {
+  var defaultHDPath = "m/44'/60'/0'" // aka "ledger legacy"
   var defaultKeyStores cli.StringSlice
   if runtime.GOOS == "darwin" {
     defaultKeyStores = []string{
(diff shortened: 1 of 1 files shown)


MrChico:
@livnev this is essentially what you would get if one were to have a more intelligent diff algo while rebasing rain-hd-path on top of master. Please try it out

livnev:
Tested and it worked for me. @rainbreak you want to test?

rainbreak:
Shall we take this opportunity to relicense to AGPL?

MrChico:
Yeah, AGPL header and a changelog entry + version bump and this looks good to me.

src/ethsign/ethsign.go
```diff
@@ -161,19 +157,15 @@ func recover(data []byte, sig hexutil.Bytes, noPrefix bool) (common.Address, err
...
-  if err != nil {
-    return common.Address{}, fmt.Errorf("invalid public key")
-  }
-  recoveredAddr := crypto.PubkeyToAddress(*pubKey)
-  return recoveredAddr, nil
+  return crypto.PubkeyToAddress(*rpk), nil
 }
 
 func main() {
+  var defaultHDPath = "m/44'/60'/0'" // aka "ledger legacy"
```
rainbreak: What should the default actually be? I'd like to use something more canonical, but changing this would be a breaking change.

livnev: It would be nice to extend this to also support setting the derivation path scheme by name, e.g. `ethsign --hd-style ledgerlive`. Right now I have to do things like `ethsign ls --hd-path "m/44'/60'/n'/0"` where `n` is the address index (this causes the selected index to be the first address), and I have not been able to find an hd path which corresponds to Ledger Live, i.e. `m/44'/60'/n'` at all. Have you?

MrChico: > Ledger Live uses : m/44'/60'/0'/0
According to this reddit post https://www.reddit.com/r/ledgerwallet/comments/9saceu/confused_at_the_hd_derivation_path_when/e8nvuqo?utm_source=share&utm_medium=web2x

livnev:
FYI this is currently somewhat broken, in that in order to access the Ledger Live derivation path `m/44'/60'/n'/0/0`, I need to do `ETH_HDPATH="m/44'/60'/n'/0" ethsign ls`, and `ETH_HDPATH="m/44'/60'/0/0" ethsign ls` _does not_ give me the right accounts (nor does any other variation on `ETH_HDPATH`). So I think you need to be incrementing the `/0'/`, not the last /0`.

Issue #277, opened 2019-09-16
hevm: possible inaccuracy in gas readouts


realisation:
I have a contract with a method that deploys another contract and calling this method costs me 5 millionish gas even though the other contract is very simple (inherits from a 3 small open zepellin ERC20 contracts). 

It deploys it via `MyContract myContract = new MyContract(arrayOf2to20Addresses);` and that contract writes the array to storage.

When I deploy that contract on Ropsten with Remix it only costs 1.5 million gas.

I'd report what it costs to call the method on my main contract to deploy it but I have to rewrite it to make it an appropriate length first.

realisation:
The deployed bytecode of the contract being deployed is 7169, so that should be about 1433800 gas plus the opcodes' cost to be deployed from a contract, right?

realisation:
Deploying the contract by itself that the factory contract was deploying reads out to 3041261 gas from hevm.

MrChico:
Gas accounting should be pretty accurate now as we are passing the GeneralStateTests. If you can provide an isolated example we can look at it. Sounds like you might be looking at the cost for deploying both contracts

realisation:
@MrChico @livnev 

I just tested again with a completely simplified version of the contract and see the same behaviors. To make it easy for you guys I swapped out the open zeppelin ERC20 I was using and used DSToken instead. 

Also here is a gist with the relevant files with remapped paths so you can copy/paste into remix for verification: https://gist.github.com/realisation/bab71e261ba847e21aaeaabbca569ff6
Just copy/paste, compile "sandbox", deploy it (you have to make sure the sandbox contract is selected when you deploy it - it will give a dropdown containing every contract from the inheritance tree), and run "create".


Using dapp test, the gas readout is 5708669 gas.

Using remix on Rinkeby and Kovan, it is 1910260 gas.

Here's the sandbox contract so you don't have to sort through that gist above if you just trust my remix figures:  

```
pragma solidity ^0.5.6;

import "../ERC20Token.sol";

contract Sandbox {

    ERC20Token token;

    function create () public {
        token = new ERC20Token("test a", "TESTA", 18, 100 * (10 ** 18));
    }

}
```

MrChico:
Closed via #294

PR #294, opened 2019-11-12
Hevm gas

MrChico:
Addresses https://github.com/dapphub/dapptools/issues/277.

Two changes were made to correct gas readouts for `dapp test`:
- The gas for the `failure()` call to check whether assertions were violated is now not included in the calculation
- Use `initialGas - usedGas` instead of `burned`. I'm not sure why this doesn't give the same result as `burned`, but it gives the right result.

Case splittings look a little ugly, but preserves the semantics.

### DIFF
--- a/src/hevm/src/EVM.hs
+++ b/src/hevm/src/EVM.hs
@@ -2037,28 +2037,16 @@ stackOp3 cost f =
 checkJump :: (Integral n) => n -> [Word] -> EVM ()
 checkJump x xs = do
   theCode <- use (state . code)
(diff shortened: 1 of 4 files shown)


MrChico:
Added
- Make hevm debug not break if you press 'N' if source is not present
- Simplify `checkJump` logic, making it rely on the `codeOps` vector rather than the `opIxMap` vector. 
The Instructions vector should not be relevant to execution, only to sourcemapping display

MrChico:
ping @livnev

PR #295, opened 2019-11-13
Missing dependency in stack.yaml to build hevm

ggrieco-tob:
This small PR adds a missing dependency to build hevm using stack.


MrChico:
Thanks!

Issue #299, opened 2019-11-20
poor performance in `hevm interactive` when backstepping


livnev:
Currently, `hevm` has to step all the way from zero when performing a backstep. This can be very slow.

livnev:
@MrChico @mbrock any ideas of how to speed it up? Now that we have `e` for jump to end, debugging any failure would be pretty quick, if only hitting `e` and backstepping wasn't cripplingly slow. Right now it's still faster to step from the beginning :|

mbrock:
Aha, well, the evaluator could save state snapshots (for every 50 states or whatever?) and then backstepping would replay from the latest snapshot. This should be easy in principle...

mbrock:
Because of how Haskell stores "immutable" data structures, this wouldn't exactly make the space consumption quadratic, it should be fine. We probably don't even have to implement any "copying" functions or whatever, just store a bunch of pointers to the `VM` structure.

mbrock:
Another fact that makes this easy is that execution is deterministic, so we can consider the snapshot list a cache that we don't need to invalidate. This would change if the debugger gains the ability to alter the VM state.

mbrock:
Specifically we would 

  - change the field `uiVmFirstState` in `EVM.TTY.UiVmState`  into something like

        uiVmSnapshots :: Vector UiVmState

  - add a constant

        snapshotInterval = 50

  - change `stepOneOpcode` to handle insertion to the vector according to `uiVmStepCount`; and 
  - change the `p` key handler to look up in the vector with some modulo arithmetic and step the remaining steps.

At a first glance that seems to be enough.

PR #301, opened 2019-11-22
hevm: TTY.hs: add periodic state snapshots for faster backstepping

livnev:
Implements @mbrock's ideas for fixing #299.


livnev:
Unfortunately this is occasionally hitting a crash which looks like:

```
hevm: internal error: can't replace code of deployed contract
  error, called at src/EVM.hs:1721:11 in hevm-0.35-7qCi2CCySav6gxUy152IKb:EVM
```

Reproduced by going to some big test, like one of the tests in `dss`, stepping to end with `e`, and then holding down `p` to step back, after taking a few dozen steps it crashes.

Cause currently not understood..

mbrock:
Hmm, the control flow is a bit wonky, but I think I see what's happening.

The `UiVmState`'s stepping progress is defined not only by the EVM state, but also by a field `uiVmNextStep :: Stepper`. A `Stepper` is a kind of script in a monadic DSL which we use to define the unit test setup/run/check logic (in `EVM.UnitTest`, see `initializeUnitTest` and `runUnitTest`). 

When you hit `n`, Hevm not only executes the next EVM opcode, but after that also continues the `Stepper` script right until it wants to execute another opcode. For example, if the executed opcode is the last opcode in a unit test method, the `runUnitTest` script will proceed to prepare the VM for running the `failed()` method. After the `n` step, the VM is now prepared to start executing `failed()`.

Now in your pull request, when we save the `UiVmState` inside `stepOneOpcode`, an EVM opcode has just been executed and the step count has been increased. However, the `uiVmNextStep` field is still unchanged, and we have not yet proceeded with the remainder of the current `Stepper` step. So for example, the unit test method has returned, but we have not yet prepared to start executing `failed()`.

I think this is what causes the weirdness. If we happen to save a snapshot in this example situation, when we later revert to it and take another step, the `uiVmNextStep` field will be as if we are continuing to run the unit test method, while in fact we have returned from that and need to start executing `failed()` and so on.

So it's just an inconsistent save state. I made an attempt to fix it here:

https://github.com/mbrock/dapptools/commit/65d5241f8e68e2fedbcd545b8e123e0d0862d7af

I think now the step count arithmetic is off because after my change, the snapshot saving logic is done *after* the step count is increased, while in your original code it was *before*. I briefly tried to fix it but didn't get it and I'm out of time for tonight!

livnev:
@mbrock thanks for your efforts in looking into this. Your description of the problem as an inconsistent save state makes sense to me, though I suspect there is still something I'm not understanding.

With your WIP version, there is a `(!)` out of bounds crash whenever I have stepped anywhere past the second snapshot and then press `p` twice. From thinking about it, I was not able to determine what error in the snapshot arithmetic would cause this.

I then tried to redo the snapshot lookup, where instead of the `mod` arithmetic we just look up the last snapshot in the vector, and compare its step count with the current step count to decide how many more steps to take on top of it. This gets rid of the out of bounds crash, and everything seems to work, except that the `can't replace code of deployed contract` crash at the end of a test returns! Which doesn't make sense to me, since you said that was caused by an inconsistent save state, which should now be fixed...

I pushed your commit and this attempted fix onto this branch, in case you want to try.

MrChico:
@livnev I think I got rid of those pesky `can't replace code of deployed contract` errors, so I think this should work now. Will you give it a run for me?

MrChico:
Closing in favor of #469

Issue #178, opened 2019-05-04
Hex inconsistencies


asymmetric:
As mentioned in https://github.com/dapphub/dapptools/issues/170#issuecomment-489329022, there is some inconsistency in how seth outputs hexadecimal numbers.

`seth---to-int256` and `seth---to-uint256` [remove the `0x`](
https://github.com/dapphub/dapptools/tree/cb9cbcdcbe3d7449d122c4543013fa884517113a/src/seth/libexec/seth/seth---to-uint256#L6), whereas `seth---to-hex` [adds it](
https://github.com/dapphub/dapptools/tree/cb9cbcdcbe3d7449d122c4543013fa884517113a/src/seth/libexec/seth/seth---to-hex#L5).

Are there any reasons for this inconsistency? e.g. are there any tools that expect the number with/without the `0x` prefix?

If we decide not to always prepend the `0x`, we should add a mention of the output base in the help of  the commands.

rainbreak:
I think all hex should have the `0x` prefix. If any tools are relying on this not being there then I think they are broken.

MrChico:
Closed by https://github.com/dapphub/dapptools/pull/326

PR #309, opened 2019-12-04
Return the checksummed address after contract creation

brianmcmichael:
On dapp create, checksum and return the contract address.

Checksummed addressed are increasingly required by consuming services, and in the way that our scripts are using the dapp output it probably makes sense to put this here prior to handing it back to the user.

### DIFF
--- a/src/dapp/libexec/dapp/dapp-address
+++ b/src/dapp/libexec/dapp/dapp-address
@@ -29,4 +29,5 @@ else
   exit 1
 fi
 
-seth keccak "$rlp" | cut -b 27-66
+raw_address=$(seth keccak "$rlp" | cut -b 27-66)
+seth --to-checksum-address "$raw_address" | cut -b 3-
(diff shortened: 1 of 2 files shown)


MrChico:
Can you add the same behavior to `dapp address` ?

brianmcmichael:
Sure thing, will add now.

brianmcmichael:
Take a look at that. It looks like `seth address` does not currently return the `0x` prefix. I retained that output here (no `0x`) to avoid breaking something else, but wasn't sure if that's intended.

nanexcool:
`seth --to-address` already does `--to-checksum-address`, handles both `0x` prefix or no prefix.

What I don't remember is when `dapp address` stopped returning with `0x` prefix. Pretty sure it did?

MrChico:
I'm not sure it did, actually. The previous implementation was using`ethereumjs-util.generateAddress` which not return `0x` unless given with `hexPrefix`. I do think that it _should_ give the `0x` prefix though, and would be happy to change the behaviour to that in this PR. Related issue https://github.com/dapphub/dapptools/issues/178

brianmcmichael:
Simple enough solution. I will just remove that final `cut` command and we will return the `0x`.

brianmcmichael:
Fixed. We'll just need to make sure it gets into the changelog to alert anyone who was adding the `0x` as a workaround in their code.

PR #313, opened 2019-12-17
Property based testing for dapps

MrChico:
Adds support for property based testing using `dapp`. By default, any test function which takes more than 0 arguments is interpreted as a property test, and it's arguments are randomly generated. 

The number of tests that are run can be modified with the `--fuzz-runs` flag, and defaults to `100`.

### DIFF
--- a/src/dapp/CHANGELOG.md
+++ b/src/dapp/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
 ### Added
- - Integrated `hevm`s new property based testing functionality. Any test with nonzero arguments will be interpreted as a property test whose arguments are randomly generated and run `--fuzz-runs` number of times.
+- Integration with `hevm`s new property based testing functionality. Any test with nonzero arguments will be interpreted as a property test whose arguments are randomly generated and run `--fuzz-runs` number of times.
 
 ## [0.26.0]
 ### Added
--- a/src/dapp/CHANGELOG.md
+++ b/src/dapp/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [unreleased]
+### Added
+ - Integrated `hevm`s new property based testing functionality. Any test with nonzero arguments will be interpreted as a property test whose arguments are randomly generated and run `--fuzz-runs` number of times.
+
 ## [0.26.0]
 ### Added
 - Support for solc 0.5.15
--- a/src/dapp/libexec/dapp/dapp-test
+++ b/src/dapp/libexec/dapp/dapp-test
@@ -5,6 +5,7 @@
 ### Options:
 ###   -v, --verbose         trace ouput for failing tests
 ###   -vv                   trace output for all tests including passes
+###   --fuzz-runs=<number>   number of times to run fuzzing tests
 ###   -m, --match=<string>  only run test methods matching regex
 ###
 ### RPC options:
--- a/src/dapp/libexec/dapp/dapp-test
+++ b/src/dapp/libexec/dapp/dapp-test
@@ -5,7 +5,8 @@
 ### Options:
 ###   -v, --verbose         trace ouput for failing tests
 ###   -vv                   trace output for all tests including passes
-###   --fuzz-runs=<number>   number of times to run fuzzing tests
+###   --fuzz-runs=<number>  number of times to run fuzzing tests
+###   --replay=<string>     rerun a particular test case
 ###   -m, --match=<string>  only run test methods matching regex
 ###
 ### RPC options:
--- a/src/dapp/libexec/dapp/dapp-test
+++ b/src/dapp/libexec/dapp/dapp-test
@@ -5,14 +5,14 @@
 ### Options:
 ###   -v, --verbose         trace ouput for failing tests
 ###   -vv                   trace output for all tests including passes
-###   --fuzz-runs=<number>  number of times to run fuzzing tests
-###   --replay=<string>     rerun a particular test case
-###   -m, --match=<string>  only run test methods matching regex
+###   --fuzz-runs <number>  number of times to run fuzzing tests
+###   --replay <string>     rerun a particular test case
+###   -m, --match <string>  only run test methods matching regex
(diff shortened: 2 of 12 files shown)


MrChico:
There's some intricacies wrt how to deal with randomly generated tests in `TTY`... Current behaviour of this PR allows you to enter property based test, but with empty arguments, meaning that your tests will always fail...

I'm working on adding a `replay` functionality, so that a failing fuzz test will give you a nice prompt ```
[FAIL] test_deposit(bytes,bytes,bytes,bytes32). Found counterexample. Run 
dapp debug --replay [test_deposit ["(0x1c717f9035a080fdd28210840948be20e02a97c8909e23c16f6bcc687f502454f325a0, 0xd7b435ea2f5504, 0x1ea091, 0x8b18b1cc15518dd4882dc054e772b946e94288c1de2ed3aa429f2560b81f7420)"]
to debug
```

**Comment by MrChico** [comment]
I decided to hide the fuzzed tests from the TTY completely unless they have been explicitly instantiated with `--replay`. Verbosity of quickcheck follows the `verbose` flag, but I've added some helpful print statements either case on how to deal with `--replay` when tests fail. 

This is now in a good state to review.

**Comment by MrChico** [comment]
I'd like to merge this, and unless there are any vocalized concerns I'll probably do so before the end of the week

**Comment by livnev** [comment]
@MrChico can you suggest a good example dapp to test this on?

**Comment by MrChico** [comment]
@livnev here's something you can try out https://github.com/makerdao/dss/compare/master...MrChico:prop
(also, just force pushed the repo, please make sure you are on latest)

**Comment by livnev** [comment]
@MrChico using `--fuzz-runs=X` gives me an error:
```
$ dapp test --fuzz-runs=1000 -m testTransferLinearity
dapp-test: error: unknown argument: --fuzz-runs=1000

Usage: dapp test [<options>]

Options:
  -v, --verbose         trace ouput for failing tests
  -vv                   trace output for all tests including passes
  --fuzz-runs=<number>  number of times to run fuzzing tests
  --replay=<string>     rerun a particular test case
  -m, --match=<string>  only run test methods matching regex

RPC options:
  --rpc                 fetch remote state via ETH_RPC_URL
  --rpc-url=<url>       fetch remote state via <url>
  --rpc-block=<number>  block number (latest if not specified)
```
while `--fuzz-runs 1000` works. `--match=FOO` likewise fails, so I think we just need to remove the `-foo=` syntax from the help text, unless someone wants to fix the parsing of those...

**Comment by livnev** [comment]
@MrChico I am trying to test this by introducing the possibility of an overflow bug for `testTransferLinearity`: see `livnev/dss` branch `prop`, I made this commit to add a bug: https://github.com/livnev/dss/commit/5ff70f32571805e477b5dbab317d59f756eb8063

(I also had to make the math in the test check safe so that it doesn't overflow)

How come quickcheck never tries any values which cause a revert due to safemath overflow?

**Comment by livnev** [comment]
n.b. I was able to find a test failure and a counterexample by introducing the possibility of an underflow bug, but not overflow, for some reason.

**Comment by MrChico** [comment]
@livnev the only way you could get an overflow would be if the address `them` already had a nonzero balance, which, together with `amount` would exceed `2^256`. But of course `them` starts with a zero balance in the overwhelming majority of cases. N.b. that the overflow safe math in the test would cause the test to revert and the test would fail because of this, and not a failed assert

**Comment by livnev** [comment]
ah yes, good point. So it probably doesn't make much sense to have `address them` in the test arguments without also initialising that balance. I made the following test and was able to find the overflow case:

```
    function testTransferLinearity(address them, uint256 preMe, uint256 preThem, uint amount) public {
      token.burn(address(this), initialBalanceThis);
      token.mint(address(this), preMe);
      token.mint(them, preThem);
      if (preMe >= amount) {
         token.transfer(address(them), amount);
      }
      uint postMe   = token.balanceOf(address(this));
      uint postThem = token.balanceOf(address(them));
      assertEq(add(preMe, preThem), add(postMe, postThem));
    }
```

However, this also made it clear that we need to be able to make a new kind of test in addition to `test` and `testFail`: one that either reverts or all asserts succeed, since otherwise it is very hard to turn the test above into something that will pass without reverting. We could call it `testInvariant` or `testHoldsOrReverts` or something like that. @MrChico thoughts?

File: src/dapp/CHANGELOG.md [hunk 1]
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [unreleased]
+### Added
+ - Integrated `hevm`s new property based testing functionality. Any test with nonzero arguments will be interpreted as a property test whose arguments are randomly generated and run `--fuzz-runs` number of times.
```
asymmetric:
```suggestion
 - Integration with `hevm`s new property based testing functionality. Any test with nonzero arguments will be interpreted as a property test whose arguments are randomly generated and run `--fuzz-runs` number of times.
```

The verb is already there 😛

src/dapp/libexec/dapp/dapp-test
```diff
@@ -5,6 +5,8 @@
 ### Options:
 ###   -v, --verbose         trace ouput for failing tests
 ###   -vv                   trace output for all tests including passes
+###   --fuzz-runs=<number>  number of times to run fuzzing tests
+###   --replay=<string>     rerun a particular test case
```
asymmetric: What format should the string here have? Could you fit a very shot description in here?

MrChico: When tests are rejected, I output a string showing exactly how to replay it

MrChico: But, long story short the syntax is `(f(uint,uint),0xdeadbeef` where the second argument is the abiencoding of the function arguments

livnev: Can the suggestion of `=` be removed here, or the parsing fixed for that case?

asymmetric: @livnev what do you mean?

livnev: @asymmetric see [here](https://github.com/dapphub/dapptools/pull/313#issuecomment-570622166)

MrChico: > Can the suggestion of `=` be removed here, or the parsing fixed for that case?

yeah, the `=` can definitely be removed. I just followed the syntax that was already given for `match`. I don't know what the syntax of bash help is supposed to look like

MrChico: so lets remove it

PR #314, opened 2019-12-23
solc 0.5.15

asymmetric:
* Bumps solc to 0.5.15
* Disable the building of shared libs in Linux, which was difficult to maintain cause it required a patch that had to be kept in sync with upstream, for very little gain.
* bumps dapp to 0.26.0

## TODO

when this is merged, create a `dapp/0.26.0` tag.

### DIFF
--- a/nix/solc-updates.md
+++ b/nix/solc-updates.md
@@ -15,7 +15,7 @@
         1. add a new `solc_X.Y.Z` in the `unreleased` section
         1. update `rev` and `sha256` based on the output from `nix-prefetch-git
            git@github.com:dapphub/nixpkgs refs/heads/solc-X.Y.Z`
-    1. bump the version number in `dapp---version`, `default.nix` and the changelog
+    1. bump the version number in `dapp---version`, `src/dapp/default.nix` and the changelog
     1. commit the changes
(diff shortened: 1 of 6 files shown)


asymmetric:
The failures on macOS are due to the fact that on Travis we use the latest master nixpkgs, which seems to be broken.

MrChico:
is there a fix for that nixpkgs? Or is the issue being tracked somewhere? Having a false negative on ci is not that fun

asymmetric:
No issue AFAIK, I brought it up on IRC but nothing yet.

[This PR](https://github.com/NixOS/nixpkgs/pull/76646) is where poetry2nix is introduced. Maybe this is the right place? I'm not sure if the problem lies there or in the Travis builder.

Either way, we shouldn't really be building PRs with the latest unstable nixpkgs, that's clear. But there's no "stable" NixOS we could set and forget - if we pin a version, we have to remember to update it when the next stable comes out. That would make our Travis builds marginally slower (a few seconds for 30+ minute builds), so I think we should.

By the way, for reassurance, have a look at [this build](https://travis-ci.com/dapphub/dapptools/jobs/270462100) - no code changes were introduced since it passed.

asymmetric:
@MrChico https://github.com/NixOS/nixpkgs/pull/76646#issuecomment-570817266

PR #315, opened 2019-12-23
seth: add 0x to --to-uint256 and friends

MrChico:
Addresses https://github.com/dapphub/dapptools/issues/178


livnev:
Afaik one of the main uses of `seth --to-uint256` is to pass to `seth calldata` or `seth call` or `seth send`, and I think none of those mind if an argument comes with or without `0x`. So this sounds good to me, of course there is probably some script somewhere that will now be doing `0x0xdeadbeef...` 😢

MrChico:
subsumed by #326

PR #320, opened 2020-01-02
Vmtestsfix

MrChico:
Unfortunately, I introduced a bug in https://github.com/dapphub/dapptools/commit/f1b9b3485fa297a8ad5cd99c38e5ac6194970d11, wherein the rlp encoding of transactions wrapped some fields twice. I.e. a transaction:
```
(Transaction {txData = "", txGasLimit = 0x61a80, txGasPrice = 0x1, txNonce = 0x0, txR = 0xe94818d1f3b0c69eb37720145a5ead7fbf6f8d80139dd53953b4a782301050a3, txS = 0x1fcf46908c01576715411be0857e30027d6be3250a3653f049b3ff8d74d2540c, txToAddr = Just 0x95e7baea6a6c7c4c2dfeb977efac326af552d87, txV = 0x1c, txValue = 0x186a0})
```
was encoded as:
```
[0x80,0x01,0x83061a80,0x94095e7baea6a6c7c4c2dfeb977efac326af552d87,0x830186a0,0x80]
```
instead of:
```
[0x,0x01,0x061a80,0x095e7baea6a6c7c4c2dfeb977efac326af552d87,0x0186a0,0x]
```

This error caused blockchain-tests to not be able to decode any "valid tests", since all transactions were malformed. But because `InvalidTx` was not treated as a fatal error, no failures were reported and we happily ignored all tests. 

This fixes the tx encoding bug, and makes `InvalidTx` a fatal error. 
Let's see how the ci does on this one

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -12,14 +12,15 @@ let
     sha256 = "06h3hcsm09kp4hzq5sm9vqkmvx2nvgbh5i788qnqh5iiz9fpaa9k";
   };
 
-  # run all General State Tests, skipping modexp as it is
-  # problematic for darwin. todo: don't skip for linux
+  # run all General State Tests, skipping tests that deal with "anomalies on the main network"
+  # (see section K.1 of https://ethereum.github.io/yellowpaper/paper.pdf), and some performance
+  # heavy ones.
   hevmCompliance = x: x.runCommand "hevm-compliance" {} ''
     mkdir "$out"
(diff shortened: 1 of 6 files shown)


MrChico:
I have also removed `modexp` from skipped tests, as its darwin problems were fixed by https://github.com/dapphub/dapptools/pull/250

Issue #265, opened 2019-08-24
hevm: ABIEncoderV2 breaks sourcemaps and stepping


livnev:
to reproduce, build the `ABIEncoderV2` branch of [this dapp](https://github.com/livnev/multipleCalls/tree/ABIEncoderV2), run `dapp debug` and observe that:

- `<no source map>` a lot of the time
- if you hold down `C-n` from the beginning, you eventually get stuck on the missing sourcemap section, and have to advance out of it with `n` or `N` until you can use `C-n` again.

Neither of these things occur without the `ABIEncoderV2` enabled

livnev:
n.b. that the example above does not actually make use of any of the new V2 ABI types, and #259 does not resolve this problem.

MrChico:
If I understand correctly, `ABIEncoderV2` is written almost entirely in yul and therefore uses the "new code generator" of solidity (i.e. one that uses yul as an ir). I believe `<no source map>` is correct here since 
> The compiler sometimes provides source links to source files internal to the compiler. In that case, the source index is -1. [1]

Perhaps the behaviour on what happens when you press `C-n` for missing sourcemap sections is the only thing that is really an issue here.
[1] https://github.com/ethereum/solidity/issues/6119#issuecomment-467797635

livnev:
@MrChico last time I reproduced this, I think I didn't see any sourcemaps at all for most of the execution, i.e. there were important parts of the code that had no sourcemap. I would have to check again to make sure that this is the case. And yes, the `C-n` is also an issue.

MrChico:
We can extract "utility" files from the compiler which contains the relevant yul code and display them in srcmaps

PR #321, opened 2020-01-03
hevm: metadata: search for bzzr1 as well as bzzr0

xwvvvvwx:
Just a little thing I noticed while investigating the embedded metadata hash. Solidity embeds `bzzr1` as part of the metadata identifier since solc 0.5.11 ([ref](https://github.com/ethereum/solidity/blob/385e1bf70a0389009af0da2f2f6054cd5b246619/Changelog.md#0511-2019-08-12))

Not sure what to test here, I ran: `nix build -f release.nix dapphub.linux.stable` and everything passes, but I guess there's something more I need to do?

### DIFF
--- a/src/hevm/src/EVM/Solidity.hs
+++ b/src/hevm/src/EVM/Solidity.hs
@@ -360,7 +360,9 @@ knownBzzrPrefixes = [
   -- a1 65 "bzzr0" 0x58 0x20 (solc <= 0.5.8)
   BS.pack [0xa1, 0x65, 98, 122, 122, 114, 48, 0x58, 0x20],
   -- a2 65 "bzzr0" 0x58 0x20 (solc >= 0.5.9)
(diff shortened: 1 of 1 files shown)


livnev:
Thanks! To properly test this would be to see it unbreak a sourcemap, since that was the original reason we introduced support for multiple metadata prefixes in #264. So `solc` `0.5.6`, `0.5.9`, and `0.5.11` should now all have working sourcemaps (and before applying your commit, `0.5.11` source maps are expected to be broken, if you can confirm).

xwvvvvwx:
in what way should I expect the sourcemaps to break with the incorrect prefix?

livnev:
@david.terry completely :P

MrChico:
cf. https://github.com/dapphub/dapptools/issues/265

PR #330, opened 2020-01-24
solc: add unreleased 0.5.16

desaperados:


desaperados:
Hmmm, this isn't available via `dapp --use solc:0.5.16` - is unreleased compatible with use when not the default?

desaperados:
What's up with the CI - max job length exceeded?

asymmetric:
Our CI setup seems to be [broken](https://github.com/dapphub/dapptools/pull/326#issuecomment-576463781), not sure why it's trying to rebuild the world.

desaperados:
ugh. How about merging regardless? Would be nice to have this available on the klab CI.

gbalabasquer:
@asymmetric we had the same problem with `dss`. I've been checking it with @icetan and seems to be some issue with cachix. Updating to this has fixed the problem: https://github.com/makerdao/dss/blob/master/.travis.yml#L5

asymmetric:
@icetan I think what was happening in dss is that you were trying to build latest cachix on [this](https://github.com/cachix/cachix/blob/master/nix/sources.json#L43) version of nixpkgs, which is quite old.

PR #331, opened 2020-01-27
travis: install cachix from latest nixpkgs

asymmetric:
This could solve the CI issues we were having, cf. https://github.com/dapphub/dapptools/pull/330#issuecomment-578716213.

I think the problem was that the pinned version of nixpkgs we were using to build/fetch cachix is too old.

By using the version in the nixpkgs Travis is already using (which I think is the latest unstable) we lose some reproducibility, but we leverage Hydra tests.


xwvvvvwx:
lgtm

asymmetric:
I think the CI pass is a false positive.

asymmetric:
[This](https://travis-ci.com/dapphub/dapptools/jobs/280513636) looks better.

Issue #333, opened 2020-01-28
`stripBytecodeMetadata` does not work as advertised


incertia:
taken from an interactive ghci session:

```
> import qualified EVM.Solidity as S
> import qualified Data.ByteString.Base16 as BS16
> BS16.encode . S.stripBytecodeMetadata . fst . BS16.decode $ "608060405234801561001057600080fd5b50600436106100365760003560e01c806317bf8bac1461003b578063acffee6b1461005d575b600080fd5b610043610067565b604051808215151515815260200191505060405180910390f35b610065610073565b005b60008060015414905090565b6000809054906101000a900473ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff1663f8a8fd6d6040518163ffffffff1660e01b815260040160206040518083038186803b1580156100da57600080fd5b505afa1580156100ee573d6000803e3d6000fd5b505050506040513d602081101561010457600080fd5b810190808051906020019092919050505060018190555056fea265627a7a723158205d775f914dcb471365a430b5f5b2cfe819e615cbbb5b2f1ccc7da1fd802e43c364736f6c634300050b0032"
"608060405234801561001057600080fd5b50600436106100365760003560e01c806317bf8bac1461003b578063acffee6b1461005d575b600080fd5b610043610067565b604051808215151515815260200191505060405180910390f35b610065610073565b005b60008060015414905090565b6000809054906101000a900473ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff1663f8a8fd6d6040518163ffffffff1660e01b815260040160206040518083038186803b1580156100da57600080fd5b505afa1580156100ee573d6000803e3d6000fd5b505050506040513d602081101561010457600080fd5b810190808051906020019092919050505060018190555056fea265627a7a723158205d775f914dcb471365a430b5f5b2cfe819e615cbbb5b2f1ccc7da1fd802e43c364736f6c634300050b0032"
```

if you are willing to introduce a `cborg` dependency, then the following snippet works
```haskell
stripBytecodeMetadata :: ByteString -> ByteString
stripBytecodeMetadata bc = if BS.length cl /= 2
                           then bc
                           else if BS.length h >= cl' && (isRight . deserialiseFromBytes decodeTerm $ fromStrict cbor)
                                then bc'
                                else bc
  where l = BS.length bc
        (h, cl) = BS.splitAt (l - 2) bc
[...]

PR #334, opened 2020-01-28
fix `stripBytecodeMetadata`

incertia:
This introduces the fix suggested in #333. `solc-0.6.2` also uses `ipfs` instead of `bzzr` so this works a bit better.


incertia:
ok so travis is failing with

```
Setup: Encountered missing dependencies:
cborg >=0.2
```

but [cborg-0.2.2.0](https://www.stackage.org/lts-14.14/package/cborg-0.2.2.0) is in `lts-14.14`

MrChico:
Can you run `make default.nix` to update the nix build process? Might be why ci is failing

incertia:
looks like everything is ok now

MrChico:
Thanks @incertia :pray:

PR #339, opened 2020-02-19
seth-calldata: multiple args support

MrChico:
Adds support for array arguments in `seth calldata`.

### DIFF
--- a/src/seth/libexec/seth/seth-calldata
+++ b/src/seth/libexec/seth/seth-calldata
@@ -14,10 +14,10 @@ if (process.argv.length >= 3 &&
     const funcs = new ethers.utils.Interface(["function " + sig.replace(')(', ') returns (')]).functions;
     const args = process.argv.slice(3).map(s => s.match(/\[.*\]/) ? JSON.parse(s) : s)
     console.log(funcs[Object.keys(funcs)[0]].encode(args))
-    } catch (e) {
-      console.error(e.toString())
-      process.exit(1)
-    }
+  } catch (e) {
+    console.error(e.toString())
+    process.exit(1)
+  }
 } else if (process.argv.length == 3) {
   const {execFileSync}   = require('child_process');
   try {
--- a/src/seth/libexec/seth/seth-calldata
+++ b/src/seth/libexec/seth/seth-calldata
@@ -12,7 +12,7 @@ if (process.argv.length >= 3 &&
   const sig = process.argv[2];
   try {
     const funcs = new ethers.utils.Interface(["function " + sig.replace(')(', ') returns (')]).functions;
-    const args = process.argv.slice(3).map(JSON.parse)
+    const args = process.argv.slice(3).map(s => s.match(/\[.*\]/) ? JSON.parse(s) : s)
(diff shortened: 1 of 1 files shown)


src/seth/libexec/seth/seth-calldata
```diff
@@ -12,7 +12,8 @@ if (process.argv.length >= 3 &&
   const sig = process.argv[2];
   try {
     const funcs = new ethers.utils.Interface(["function " + sig.replace(')(', ') returns (')]).functions;
-      console.log(funcs[Object.keys(funcs)[0]].encode(process.argv.slice(3)))
+    const args = process.argv.slice(3).map(JSON.parse)
```
asymmetric: This seems like it could fail, because [`JSON.parse`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse) takes an additional optional argument, and `Array.map` passes 3 args to the callback -  cf. [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map), section *Tricky use case*.

Better be explicit IMO:

```suggestion
    const args = process.argv.slice(3).map(arg => JSON.parse(arg))
```

src/seth/libexec/seth/seth-calldata
```diff
@@ -12,7 +12,8 @@ if (process.argv.length >= 3 &&
   const sig = process.argv[2];
   try {
     const funcs = new ethers.utils.Interface(["function " + sig.replace(')(', ') returns (')]).functions;
-      console.log(funcs[Object.keys(funcs)[0]].encode(process.argv.slice(3)))
+    const args = process.argv.slice(3).map(s => typeof(s) === 'string' ? JSON.parse(s) : s)
```
asymmetric: To make sure I understand: Why are we doing `JSON.parse`? I think [`process.argv`](https://nodejs.org/api/all.html#process_process_argv) returns an array of `string`s. Have you had issues without it?

travs: hey @asymmetric , the changes in this PR are meant to address arrays which aren't parsed properly with the trunk version of `seth`.

e.g. `seth calldata 'multicall(bytes[])' '["0x00","0x01"]'`

src/seth/libexec/seth/seth-calldata
```diff
@@ -12,7 +12,8 @@ if (process.argv.length >= 3 &&
   const sig = process.argv[2];
   try {
     const funcs = new ethers.utils.Interface(["function " + sig.replace(')(', ') returns (')]).functions;
-      console.log(funcs[Object.keys(funcs)[0]].encode(process.argv.slice(3)))
+    const args = process.argv.slice(3).map(JSON.parse)
+    console.log(funcs[Object.keys(funcs)[0]].encode(args))
```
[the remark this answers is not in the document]
MrChico: fixed

asymmetric:
@mrchico I think we could add this to the changelog, either under unreleased or as a new release.

MrChico:
Yeah, lets make a release

PR #340, opened 2020-02-20
Updated readme to include --to- address int256 uint256 and bytes32

realisation:
I tried out --to-bytes32 on a few things and it seems to only work with numbers. Is that right?

### DIFF
--- a/src/seth/README.md
+++ b/src/seth/README.md
@@ -59,9 +59,10 @@ Contents
       * [`seth --from-bin`]
       * [`seth --from-wei`]
       * [`seth --to-wei`]
+      * [`seth --to-int256`]
       * [`seth --to-uint256`]
       * [`seth --to-bytes32`]
-      * [`seth --to-checksum-address`]
+      * [`seth --to-address`]
       * [`seth age`]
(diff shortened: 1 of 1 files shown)


src/seth/README.md
```diff
@@ -329,6 +333,30 @@ Convert an ETH amount into wei.
...
+
+### `seth --to-uint256`
+
+Convert a number into uint256 hex string with 0x prefix.
+
+    seth --to-uint256 <value>
+
+### `seth --to-bytes32`
+
+Convert data (what types, just numbers?) into bytes32 hex string with 0x prefix.
```
MrChico:
```suggestion
Right pads a hex string with zeros to 32 bytes.
```

src/seth/README.md
```diff
@@ -329,6 +333,30 @@ Convert an ETH amount into wei.
...
+
+    seth --to-bytes32 <value>
+
+### `seth --to-address
+
+Convert address into checksum address.
```
asymmetric:
```suggestion
Convert an address into a checksummed address.
```

PR #344, opened 2020-02-21
seth: 0.8.4 release

MrChico:


asymmetric:
@MrChico you forgot [this](https://github.com/dapphub/dapptools/blob/master/src/seth/libexec/seth/seth---version).

PR #345, opened 2020-02-23
Seth readme - added gas-price method

realisation:


src/seth/README.md
```diff
@@ -63,6 +63,7 @@ Contents
       * [`seth --to-uint256`]
       * [`seth --to-bytes32`]
       * [`seth --to-address`]
+      * [`seth gas-price`]
```
asymmetric: Can you please add it in alphabetical order?

realisation: Indeed

src/seth/README.md
```diff
@@ -357,6 +358,10 @@ Convert an address into a checksummed address.
 
     seth --to-address <address>
 
+### `seth gas-price`
```
asymmetric: This should be sprted alphabetically too :)

Thanks! :pray:

PR #351, opened 2020-03-18
dapp create without need of --extract

nanexcool:
Uses dapp.sol.json data for dapp create, without needing `--extract`

If there are multiple contracts with the same name (on different files) it fails. I believe this is the right thing to do, but wondering how to fix so that it's backwards compatible

```
$ dapp create MyContract
dapp-create: error: MyContract found in multiple files
src/MyContract.sol:MyContract
src/MyContract2.sol:MyContract
```

### DIFF
--- a/src/dapp/libexec/dapp/dapp-create
+++ b/src/dapp/libexec/dapp/dapp-create
@@ -16,8 +16,31 @@ for i in "$@"; do
   esac
 done
 
-bin=$DAPP_OUT/${1?missing contract name}.bin
-type=$(seth --abi-constructor <"$DAPP_OUT/$1.abi")
+name=${1?missing contract name}
+
+mapfile -t contracts < <(<"$DAPP_JSON" jq '.contracts|keys[]' -r)
+
+name=()
+for path in "${contracts[@]}"; do
+  [[ "${path#*:}" = "$1" ]] && name+=("$path")
+done
+
+[[ "${#name[@]}" -eq 0 ]] && {
+  echo >&2 "${0##*/}: error: $1 not found"
+  exit 1
+}
+
+[[ "${#name[@]}" -ge 2 ]] && {
+  echo >&2 "${0##*/}: error: $1 found in multiple files"
+  printf '%s\n' "${name[@]}"
+  exit 1
+}
+
+data=$(<"$DAPP_JSON" jq '.contracts' -r)
+contract=$(echo "$data" | jq '.[''"'"$path"'"'']')
+abi=$(jq '.["abi"]' -r <<< "$contract")
+bin=$(jq '.["bin"]' -r <<< "$contract")
+type=$(seth --abi-constructor <<< "$abi")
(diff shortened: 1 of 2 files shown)


src/dapp/libexec/dapp/dapp-create
```diff
@@ -16,8 +16,36 @@ for i in "$@"; do
...
-bin=$DAPP_OUT/${1?missing contract name}.bin
-type=$(seth --abi-constructor <"$DAPP_OUT/$1.abi")
+[[ -e "${DAPP_JSON}" ]] || {
+  echo >&2 "${0##*/}: error: ${DAPP_JSON} file not found. Did you run ``dapp build``?"
+  exit 1
+}
+
+name=${1?missing contract name}
+
+mapfile -t contracts < <(<"$DAPP_JSON" jq '.contracts|keys[]' -r)
```
asymmetric: We could maybe ensure that the jq selector has matched something by using the `-e` flag and checking for its exit status. Otherwise we're trusting that the file has the structure we expect, which might not necessarily be the case (also wrt backwards/forward-compatibility).

WDYT?

src/dapp/libexec/dapp/dapp-create
```diff
@@ -16,8 +16,36 @@ for i in "$@"; do
...
+  exit 1
+}
+
+[[ "${#name[@]}" -ge 2 ]] && {
+  echo >&2 "${0##*/}: error: $1 found in multiple files"
+  printf '%s\n' "${name[@]}"
+  exit 1
+}
+
+data=$(<"$DAPP_JSON" jq '.contracts' -r)
```
asymmetric: We could maybe ensure that the jq selector has matched something by using the -e flag and checking for its exit status. Otherwise we're trusting that the file has the structure we expect, which might not necessarily be the case (also wrt backwards/forward-compatibility).

WDYT?

nanexcool:
> Does this mean we can remove `dapp-build --extract`?

well, I think `--extract` is useful if you want to separate abis, metadata, bytecode, etc, into different files. But yes, you wouldn't need the `--extract` flag to call `dapp create` since it will read from `dapp.sol.json` directly

src/dapp/libexec/dapp/dapp-create
nanexcool: yep, good catch! I'll add it

nanexcool: Actually... `mapfile -t contracts < <(<"$DAPP_JSON" jq '.contracts|keys[]' -r)` already checks this right? And if it didn't find anything `contracts` will be length 0, and it will fail.

nanexcool:
I've forgone the use of a flag, and if multiple files are found you can pass the full path:
```
$ dapp create MyContract
dapp-create: error: MyContract found in multiple files. Use full path:
dapp create src/Create.sol:MyContract
dapp create src/Create2.sol:MyContract
### then
$ dapp create src/Create2.sol:MyContract -G 5000
# works
```

MrChico:
LGTM

Issue #305, opened 2019-11-29
add --rpc flag to `hevm exec`


livnev:
This would basically allow one to give `eth_call`s to `hevm`.

Issue #273, opened 2019-09-09
Hevm exec precompile support


MrChico:
When calling a precompile inside another contract, `hevm exec` breaks with:
`Query <EVM.Query: fetch contract 0000000000000000000000000000000000000002>`

However, when running with `--debug` it succeeds.

Here's some example code that triggers this:
```
740100000000000000000000000000000000000000006020526f7fffffffffffffffffffffffffffffff6040527fffffffffffffffffffffffffffffffff8000000000000000000000000000000060605274012a05f1fffffffffffffffffffffffffdabf41c006080527ffffffffffffffffffffffffed5fa0e000000000000000000000000000000000060a052341561009857600080fd5b6101406000601f818352015b600061014051602081106100b757600080fd5b600260c052602060c020015460208261016001015260208101905061014051602081106100e357600080fd5b600260c052602060c020015460208261016001015260208101905080610160526101609050602060c0825160208401600060025af100```

PR #353, opened 2020-04-05
Symbolic execution for HEVM; `hevm symbolic` and `hevm equivalence`

MrChico:
This draft PR is ~a WIP~ extends `hevm` with symbolic execution capabilities, allowing it to be used as a proving engine for formal verification, as well as admitting more intelligent fuzzing strategies.

~It's not particularly close to being ready for merging yet, but I would like to share my progress to get some early feedback and discuss some of my findings so far.~
It is now ready for review!

## On a high level, here's what's going on:

Most of the magic here comes from the SMT-based verification library [sbv](https://hackage.haskell.org/package/sbv-8.6/docs/Data-SBV.html), which provides symbolic variants of common basetypes like booleans, words and integers. These variants are strict generalizations of their concrete counterparts and functions over them will just evaluate normally when given concrete arguments. Using these we can generalize various parts of the EVM:

- The EVM `stack` is changed from a list of words, `[Word 256]`, to a list of symbolic words, `[SWord 256]`, and stack operations are adapted accordingly.
- The buffers `memory`, `returndata` and `calldata` are changed from `ByteString` to `[SWord 8]`. This is a (concrete) list of symbolic values, not a symbolic list, which means that while reads and writes can refer to symbolic values, they must always happen at concrete locations. This restriction means that we are unable to deal with dynamic data types for the moment. More on this later.

~The main part of symbolic execution is handled by the function [`symExec`](https://github.com/dapphub/dapptools/blob/symbolic/src/hevm/src/EVM/Symbolic.hs#L87), which executes the EVM with symbolic values while keeping track of an accumulated list of constraints (path conditions), until reaching a JUMPI opcode. At this point, it invokes z3 to check which branches are possible, and recurses with the path condition updated accordingly until reaching a list of possible final VM states. The resulting VM states can be checked against a post condition.~

EDIT (04/07/2020):

Symbolic execution is now done with the "operational monad pattern" as seen in `TTY`, `VMTests` `Dev`, etc. Similarly to how instances of `SLOAD` or `CALL` causes the execution to end with a `Query` error, `JUMPI` introduces a new type of query (`PleaseAskSMT`), which will need to be performed in an `IO` context. The actual smt query is performed in a [new fetcher](https://github.com/dapphub/dapptools/blob/symbolic/src/hevm/src/EVM/Fetch.hs#L161), which is used in `TTY.interpret` (which (as before) only keeps track of one VM at a time) and the new [`SymExec.interpret`](https://github.com/dapphub/dapptools/blob/symbolic/src/hevm/src/EVM/SymExec.hs#L126), which is essentially a function `VM -> [VM]`, but crucially does the necessary lifting in case we encounter any blocking queries.

Some example proofs are given in [test.hs](https://github.com/dapphub/dapptools/blob/symbolic/src/hevm/test/test.hs#L169). 

While defining arbitrary post conditions is probably better handled with something like [act](https://github.com/ethereum/act), this PR provides an easy way to check for asserts (executions ending with an `INVALID` opcode (`0xfe`)) using the cli endpoint `hevm symbolic`.

Example usage:
```sh
echo "pragma solidity >=0.5.15; 
           contract A {
               function factor(uint x, uint y) public pure {
                   require(1 < x && x < 973013 && 1 < y && y < 973013);
                   assert(x*y != 973013);}
}" \
| solc --combined-json=bin-runtime - \
| jq '."contracts"."<stdin>:A"."bin-runtime"' \
| tr -d '"' > fun.txt

hevm symbolic --code $(<fun.txt) --abi $(seth --abi-function-json "factor(uint256,uint256)")
"Assertion violation:"
(1021, 953)
```

## Obvious TODOs

- [x] Allow `storage` to be symbolic as well.

- [x] Ensure that the conversion did not mess up any of the general state tests. The semantics for concrete execution should remain unchanged after this.

## More interesting TODOs:

- [x] EVM words are a little too large for `z3` to be good at dealing with them. As a result, it hangs on some simple proofs like [this tautological SafeAdd proof](https://github.com/dapphub/dapptools/blob/symbolic/src/hevm/test/test.hs#L140). This has also been noted as a problem for @leonardoalt, which recommended to use unbounded integers instead. This would require adapting all arithmetic opcodes to operate modulo 2^256 instead. But there also might be other solvers than `z3` that are better at dealing with large bitvectors.

> EDIT 08 June:
> There is now support for cvc4 as well, which in certain situations deals a lot better with large bitvectors. Large bitvectors seems preferable to modulo integer arithmetic, especially since most contracts use quite a lot of bitvector operations to do masking and other encoding gymnastics.

- [x] It should be possible to generalize byte buffers like memory and calldata to deal with symbolic locations. SBV offers two options here, `SArray` which translates directly to an smt `array`, and `SFunArray`, which is "implemented internally, without translating to SMT-Lib functions". I'm guessing the latter option is preferable.

> EDIT 08 June:
> I have generalized calldata slightly with a slight of hand: calldata is now a tuple `(SWord 32, [SWord 8])`, where the first argument is understood as the calldata length. Since EVM just returns zeros when reading out of bounds calldata, this allows us to fully abstract calldata without having to deal with arrays.

- [x] The `Show` instance for symbolic values, is quite boring as it just displays `<symbolic>` for symbolic variables. For a more insightful `hevm --debug` experience, we might want to accompany symbolic words with a pretty-print string, at least when function ABIs are available. Opcodes would need be adapted to update the pretty-print string as well as updating the values.

> EDIT 13 July:
> Basic scaffolding for displaying symbolic values with a richer AST is given, but the full implementation is decided to be out of scope of this first PR.

- [x] Quite understandably, the current execution strategy fails miserably when dealing with loops whose condition depends on a symbolic value. Depending on the setting, one could go in several directions to account for this:
1. Keep track of which pc values have been visited, and restrict the number of times we are allowed to return to a particular pc value, aka "bounded model checking".
2. Maintain a set of "trusted specs", i.e. `VM -> VM` functions that can only be applied under certain conditions. These "trusted specs" can themselves be proven by symbolic execution and coinduction.
3. Try to infer loop invariants automagically using some CHC style solution or whatever [Data.SBV.Tools.Induction](https://hackage.haskell.org/package/sbv-8.6/docs/src/Data.SBV.Tools.Induction.html) uses. I'm doubtful this would work, but it's worth exploring.

> EDIT 08 June:
> 1. is now implemented, and the number of times revisiting the same branching point is allowed can be configured using the flag `--max-iterations`.
> 2 can be done once better act integration is here, and 3 remains interesting to explore in the future.

This type of symbolic execution would probably be able to generate a lot more interesting test cases for tools like Echidna. Do you guys (@japesinator, @incertia) have any specific requests for functions that would allow you to incorporate this into your workflow?

What are the effects on performance for concrete execution? Hopefully it does not change drastically, but I think a *slight* performance decrease would still be acceptable.

> EDIT: July 13
> Performance is actually better on this PR than on master, see the measurements done by @xwvvvvwx further down this thread.


leonardoalt:
> It should be possible to generalize byte buffers like memory and calldata to deal with symbolic locations. SBV offers two options here, SArray which translates directly to an smt array, and SFunArray, which is "implemented internally, without translating to SMT-Lib functions". I'm guessing the latter option is preferable.

@MrChico What's `SFunArray`? If it's not too hard to implement, maybe give SMT arrays a try as well at some point. It is one of the most studied/implemented SMT theories, so I wouldn't be surprised if it's performance is good.

> Try to infer loop invariants automagically using some CHC style solution

I'd be interested in trying something like that

MrChico:
> @leonardoalt  What's SFunArray?

From what I understand, when using `SFunArray`, sbv handles the "internals" of the array structure manually – it maintains symbolic variables for all used keys and values, and translates reads and writes to big nested `ite` expressions over these. It also does caching, apparently.
As a result, one cannot compare two `SFunArray`s.

It seems appropriate for memory and storage, where we care mainly about the values at specific keys, rather than the structure as a whole. 
For buffers like `returndata` and `calldata` where we want to make claims about the entire structure, maybe SMT arrays are more appropriate

MrChico:
Also implements https://github.com/dapphub/dapptools/issues/305 as of https://github.com/dapphub/dapptools/pull/353/commits/d9f8741a1a68e2dc098fe8555d5a51cd7ccecef2, allowing previewing of transactions or debugging of past transactions.

MrChico:
Incidentally also solves https://github.com/dapphub/dapptools/issues/273

livnev:
> Also implements #305 as of [3575269](https://github.com/dapphub/dapptools/commit/3575269f0689446de755ab6512de7114031ada8f), allowing previewing of transactions or debugging of past transactions.

Wow amazing!

MrChico:
I now consider this effort feature complete for a first release. I'll be adding some more tests and docs (please let me know whats unclear!) as well as writing a blog post over the next couple of days, but that shouldnt be blocking review.

I'd be happy for experiments all over, but in particular the equivalence checking and the symbolic execution against remote state (using rpc) would benefit from some extra hands trying things out 

Concrete execution and all old cli endpoints should remain their current functionality (modulo some minor improvements), so please lmk if you find any regressions here. That said, many library functions have changed type signature though so this is a very breaking release 

Finally, I suggest we should not squash these commits when merging as that would make the git history fairly opaque.

src/hevm/src/EVM/TTY.hs
```diff
@@ -214,9 +214,14 @@ interpret mode =
...
 
+        -- Stepper is waiting for user input from a query
+        Stepper.Option _ -> do
+                  -- pause & await user.
+                  pure (Stepped (Operational.singleton action >>= k))
+
         -- Stepper wants to make a query and wait for the results?
         Stepper.Wait q ->
-          -- Tell the TTY to run an I/O action to produce the next stepper.
+          -- Tell the TTY to run a SBV.query action to produce the next stepper.
```
MrChico, line 224:
```suggestion
          -- Tell the TTY to run an I/O action to produce the next stepper.
```

src/hevm/src/EVM/TTY.hs
```diff
@@ -352,14 +364,15 @@ takeStep ui policy mode =
             StepNormally StepNone
 
         StepTimidly ->
-          error "step blocked unexpectedly"
+          error $ "step blocked unexpectedly" ++ show (view (uiVm . result) ui')
+--                  ++ "blocker:" ++ show blocker
```
MrChico, line 368:
```suggestion
```

haskell.nix
```diff
@@ -14,7 +14,7 @@ in self-hs: super-hs:
         (self-hs.callPackage x {});
     hevmBinPath = lib.makeBinPath (with pkgs; [bash coreutils git]);
     sbv_8_4_prepatch = self-hs.callCabal2nix "sbv" (builtins.fetchGit {
-        url = "https://github.com/LeventErkok/sbv/";
+        url = "https://github.com/LeventErkok/sbv";
```
MrChico, line 17:
```suggestion
        url = "https://github.com/LeventErkok/sbv/";
```

src/hevm/src/EVM/TTY.hs
```diff
@@ -917,17 +957,27 @@ withHighlight :: Bool -> Widget n -> Widget n
...
+  Just lits -> prettyHex 40 lits
+  Nothing -> show x
+
 drawTracePane :: UiVmState -> UiWidget
 drawTracePane s =
   case view uiVmShowMemory s of
     True ->
       hBorderWithLabel (txt "Calldata")
-      <=> str (prettyHex 40 (view (uiVm . state . calldata) s))
+--      <=> str (prettyHex 40 (view (uiVm . state . calldata) s))
```
MrChico, line 970:
```suggestion
```

src/hevm/src/EVM/UnitTest.hs
```diff
@@ -254,17 +228,23 @@ currentOpLocation vm =
...
   case view result vm0 of
     Nothing -> do
       vm1 <- zoom _1 (State.state (runState exec1) >> get)
       zoom _2 (modify (MultiSet.insert (currentOpLocation vm1)))
-      execWithCoverage
-    Just r ->
-      pure r
+      runWithCoverage
+    Just _ -> pure vm0
+--    Just _ -> id
```
MrChico, line 246:
```suggestion
```

src/hevm/src/EVM/VMTest.hs
```diff
@@ -205,21 +202,26 @@ checkExpectedOut output ex = case ex of
 
 checkExpectedContracts :: EVM.VM -> Map Addr Contract -> (Bool, Bool, Bool, Bool, Bool)
 checkExpectedContracts vm expected =
-  let cs = vm ^. EVM.env . EVM.contracts . to (fmap clearZeroStorage)
-      expectedCs = realizeContracts expected
+  let cs = vm ^. EVM.env . EVM.contracts . to (fmap (clearZeroStorage.clearOrigStorage))
+      expectedCs = fmap clearOrigStorage $ realizeContracts expected
```
MrChico, line 206:
```suggestion
      expectedCs = clearOrigStorage <$> realizeContracts expected
```

src/hevm/src/EVM/VMTest.hs
```diff
@@ -355,7 +357,12 @@ realizeContract x =
   EVM.initialContract (x ^. code)
     & EVM.balance .~ EVM.w256 (x ^. balance)
     & EVM.nonce   .~ EVM.w256 (x ^. nonce)
-    & EVM.storage .~ (
+    & EVM.storage .~ EVM.Concrete (
+        Map.fromList .
+        map (\(k, v) -> (EVM.w256 k, EVM.w256 v)) .
```
MrChico, line 362:
```suggestion
        map (bimap EVM.w256 EVM.w256) .
```

MrChico:
The failing test is confirmed to be due to an upstream error in sbv, for which a fix is being developed. Currently, storage updates are forgotten causing us to explore inconsistent branches. I suggest we wait with merging until the fix has arrived

endorphin:
Running the above example crashes for me:
```
(nix-shell) [~/repos/cyberbrain/dapptools] > cat fun.txt
6080604052348015600f57600080fd5b506004361060285760003560e01c8063d5a2424914602d575b600080fd5b606060048036036040811015604157600080fd5b8101908080359060200190929190803590602001909291905050506062565b005b81600110801560735750620ed8d582105b8015607e5750806001105b8015608b5750620ed8d581105b609357600080fd5b620ed8d5818302141560a157fe5b505056fea265627a7a72315820ef21c2220ea85dd4936a620a351bbd11caa5465867fa7d1def47c8c53ad14e1164736f6c634300050f0032
(nix-shell) [~/repos/cyberbrain/dapptools] > ./result/bin/hevm symbolic --code $(<fun.txt) --abi $(seth --abi-function-json "factor(uint256,uint256)")
hevm: (^?!): empty Fold
  error, called at src/Control/Lens/Fold.hs:1285:28 in lens-4.17.1-H6wZkUUks109osGIy1GNSc:Control.Lens.Fold
  ^?!, called at hevm-cli/hevm-cli.hs:635:82 in main:Main
```
Perhaps I haven't set everything up properly?

MrChico:
@endorphin did you reinstall seth before running the example? This PR slightly changes to output of seth --abi-function-json

xwvvvvwx:
Some rough benchmarks re: performance impact vs master:

I ran `hevm compliance` against `ethereum/tests@6af0621522dd0274525457741291d391c10002be` 5 times for both the `symbolic` and `master` branches on my laptop (Intel Core i7-8550U @ 1.80GHz // 16 GB RAM) and took the averages:


branch |                symbolic                   |                  master
-------|-------------------------------------------|------------------------------------------
   1   | 342.51s user 135.87s system 4:28.33 total | 434.89s user 147.62s system 5:14.92 total
   2   | 341.35s user 139.55s system 4:25.67 total | 339.57s user 136.74s system 4:22.00 total
   3   | 324.48s user 129.36s system 4:15.24 total | 332.86s user 132.87s system 4:25.63 total
  avg  |               262.66 s total              |               280.33 s total

So at least according to these tests, the addition of symbolic does not appear to have impacted performance of concrete execution, and may even have made things a little faster!

MrChico:
Really nice @xwvvvvwx , thanks for taking these measurements! This is great news!

src/hevm/src/EVM.hs
```diff
@@ -233,26 +254,44 @@ data ContractCode
...
+data Storage
+  = Concrete (Map Word SymWord)
+  | Symbolic (SArray (WordN 256) (WordN 256))
+  deriving (Show)
+
+-- to allow for Eq Contract (which useful for debugging vmtests)
+-- we mock an instance of Eq for symbolic storage.
+-- It should not (cannot) be used though.
+instance Eq Storage where
+  (==) (Concrete a) (Concrete b) = a == b
```
MrChico, line 268:
```suggestion
  (==) (Concrete a) (Concrete b) = second forceLit a == second forceLit b
```

src/hevm/src/EVM.hs
```diff
@@ -14,22 +15,22 @@ module EVM where
...
-
 import qualified EVM.Precompiled
 
 import Data.Binary.Get (runGetOrFail)
 import Data.Text (Text)
 import Data.Word (Word8, Word32)
-import Data.Bits (bit, testBit, complement, xor, shiftR, (.&.), (.|.), FiniteBits(..))
-
-import Control.Lens hiding (op, (:<), (|>))
+import Control.Lens hiding (op, (:<), (|>), (.>))
```
MrChico, line 33:
```suggestion
import Control.Lens hiding (op, (:<), (|>), (.>))
import Data.Bifunctor (second)
```

src/hevm/src/EVM.hs
```diff
@@ -14,22 +15,23 @@ module EVM where
...
+import Data.Bifunctor (second)
```
MrChico, line 34:
```suggestion
```

src/hevm/src/EVM.hs
```diff
@@ -233,26 +255,44 @@ data ContractCode
...
+data Storage
+  = Concrete (Map Word SymWord)
+  | Symbolic (SArray (WordN 256) (WordN 256))
+  deriving (Show)
+
+-- to allow for Eq Contract (which useful for debugging vmtests)
+-- we mock an instance of Eq for symbolic storage.
+-- It should not (cannot) be used though.
+instance Eq Storage where
+  (==) (Concrete a) (Concrete b) = second forceLit a == second forceLit b
```
MrChico, line 269:
```suggestion
  (==) (Concrete a) (Concrete b) = fmap forceLit a == fmap forceLit b
```

src/hevm/hevm-cli/hevm-cli.hs
```diff
@@ -319,6 +361,93 @@ regexMatches regexSource =
...
+    Just _ -> error "Unknown solver. Currently supported solvers; z3, cvc4"
+
+-- Although it is tempting to fully abstract calldata and give any hints about the nature of the signature
+-- doing so results in significant time spent in consulting z3 about rather trivial matters. But with cvc4 it is quite pleasant!
+
+-- If function signatures are known, they should always be given for best results.
+assert :: Command Options.Unwrapped -> IO ()
+assert cmd = do
+
+  let root = fromMaybe "." (dappRoot cmd)
```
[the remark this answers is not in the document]
MrChico: Good catch, thanks!

haskell.nix
```diff
@@ -13,17 +13,17 @@ in self-hs: super-hs:
...
+        rev = "80883626faaf3f9cf541ba9b98775d0afd00c07a";
     }) {inherit (pkgs) z3;};
 
   in {
     restless-git = dontCheck (import ./src/restless-git);
     wreq = pkgs.haskell.lib.doJailbreak super-hs.wreq;
 
     # we use a pretty bleeding edge sbv version
-    sbv_8_4 = sbv_8_4_prepatch.overrideAttrs (attrs: {
+    sbv_8_6 = sbv_8_6_prepatch.overrideAttrs (attrs: {
```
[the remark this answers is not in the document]
MrChico: for the `nixpkgs` we are using, this is more up to date

src/hevm/src/EVM/Fetch.hs
```diff
@@ -124,6 +128,7 @@ fetchSlotFrom n url addr slot =
   Session.withAPISession
     (\s -> fetchSlotWithSession n url s addr slot)
 
+
```
MrChico, line 131:
```suggestion
```

xwvvvvwx:
@MrChico something that came up while I was playing around with `symbolic`:

When executing `mint` in [`uniswap-v2-core`](https://github.com/Uniswap/uniswap-v2-core/), the output on the command line shows a succesful execution:

```
$ hevm symbolic --json-file combined.json --code $(cat combined.json | jq '.contracts."contracts/UniswapV2Pair.sol:UniswapV2Pair"."bin-runtime"' | tr -d '"') --abi $(seth --abi-function-json "mint(address)")
checking postcondition...
Q.E.D
Explored: 2 branches without assertion violations
```

but when I step through in the debugger it seems to just get stuck on `EXTCODESIZE` against the underlying token contract (pressing `n` does nothing, pressing `p` results in an `hevm: step halted unexpectedly` error), as I understand it, `hevm` should fail here with `unexpected symbolic argument`?

MrChico:
@xwvvvvwx this is indeed an issue. hevm symbolic does currently not distinguish between unexpected failures (such as unexpectedsymbolicarg), and "safe" ones like reverts. You should see VMError: UnexpectedSymbolicArg at the top of the debugger window when running with `--debug` though.

xwvvvvwx:
> You should see `VMError: UnexpectedSymbolicArg` at the top of the debugger window when running with --debug though.

Ah yes, I had missed that.

MrChico:
Now making sure that the user is alerted about paths failing due to unexpected symbolic arguments:
```sh
curl https://unpkg.com/@uniswap/v2-core@1.0.1/build/Combined-Json.json > uniswapv2.json
hevm symbolic --json-file uniswapv2.json --code $(cat uniswapv2.json | jq '.contracts."contracts/UniswapV2Pair.sol:UniswapV2Pair"."bin-runtime"' | tr -d '"') --sig "mint(address)"
checking postcondition...
Q.E.D
Explored: 2 branches without assertion violations
However, 1branches errored while exploring:
["Unexpected symbolic argument at opcode: OpExtcodesizefeature. Not supported (yet!)"]
```

src/hevm/src/EVM/SymExec.hs
```diff
@@ -0,0 +1,287 @@
...
+    , vmoptSchedule = FeeSchedule.istanbul
+    , vmoptCreate = False
+    }) & set (env . contracts . at (createAddress ethrunAddress 1))
+             (Just (contractWithStore x initStore))
+
+-- Interpreter which explores all paths at
+-- branching points.
+-- returns a list of possible final evm states
+interpret
+  :: (EVM.Query -> IO (EVM ()))
```
xwvvvvwx, line 133: nit: maybe `Fetch.Fetcher` here to match the `interpret` in `Stepper.hs`?

src/hevm/src/EVM/SymExec.hs
```diff
@@ -0,0 +1,287 @@
...
+                            return $ liftA2 (<>) a b
+               Nothing -> do a <- State.state (runState (continue 0)) >> interpret fetcher maxIter (k ())
+                             put vm
+                             b <- State.state (runState (continue 1)) >> interpret fetcher maxIter (k ())
+                             return $ liftA2 (<>) a b
+        Stepper.Wait q ->
+          do m <- liftIO (fetcher q)
+             State.state (runState m) >> interpret fetcher maxIter (k ())
+        Stepper.Note _ ->
+          -- simply ignore the note here
```
xwvvvvwx, line 180: Is the `Note` action actually required? It seems to be ignored in the interpreters in `UnitTest`, `SymExec` and `Fetch`?

src/hevm/src/EVM/SymExec.hs
```diff
@@ -0,0 +1,287 @@
...
+        return (sig' <> concat (fst <$> calldatas), 4 + (sum $ snd <$> calldatas))
+
+abstractVM :: Maybe (Text, [AbiType]) -> [String] -> ByteString -> StorageModel -> Query VM
+abstractVM typesignature concreteArgs x storagemodel = do
+  (cd', cdlen, cdconstraint) <-
+    case typesignature of
+      Nothing -> do cd <- sbytes256
+                    len <- freshVar_
+                    return (cd, len, len .<= 1024)
+      Just (name, typs) -> do (cd, cdlen) <- symCalldata name typs concreteArgs
```
xwvvvvwx, line 96:
```suggestion
      Just (name, types) -> do (cd, cdlen) <- symCalldata name types concreteArgs
```

src/hevm/src/EVM/SymExec.hs
xwvvvvwx: oh, I had missed `TTY.interpret`....

MrChico:
I'm planning on merging this approximately on wednesday the 22, unless there are strong objections or people would like more time to look it over? @livnev @xwvvvvwx @rainbreak @asymmetric

src/hevm/hevm-cli/hevm-cli.hs
```diff
@@ -318,6 +362,107 @@ regexMatches regexSource =
...
+      srcinfo = ((,) root) <$> (jsonFile cmd)
+      block'  = maybe EVM.Fetch.Latest EVM.Fetch.BlockNumber (block cmd)
+      rpcinfo = (,) block' <$> rpc cmd
+  maybesig <- case sig cmd of
+        Nothing -> return Nothing
+        Just sig' -> do method' <- functionAbi sig'
```
xwvvvvwx, line 423: `functionAbi` assumes that a solc > 0.6.7 is present in the users path (check `solidity'` in `Solidity.hs`), which is not currently the case on my system, meaning calls to `hevm symboilc` fail with:

```
Invalid option to --combined-json: storage-layout
```

maybe we can do something smart with `nix` here?

xwvvvvwx:
Pressing `p` after reacing a `REVERT` pretty consistently results in a crash with:

```
hevm: step halted unexpectedly
  error, called at src/EVM/TTY.hs:375:11 in hevm-0.40-ALk2NN6uTYt9GD084Gwlts:EVM.TTY
```

src/hevm/hevm-cli/hevm-cli.hs
xwvvvvwx: like maybe a `makeWrapper` around `hevm` that places `solc` 0.6.7 in path for `hevm`?

MrChico: oh, wow, yeah, I thought it `solc` was fixed by nix here. I wonder if adding solc [here](https://github.com/dapphub/dapptools/blob/9d48a1a8b572654adad05a0218169f2b6beef9f2/haskell.nix#L57) would be sufficient

xwvvvvwx: oh yes, I think that should work :smiley_cat:

MrChico: hmm, looks like it doesn't...

xwvvvvwx:
I get crashes when pressing `ctrl-n` while Executing `0x7665Fecb393ca45061A6338662751D6cF45A854d` using `rpc`:

```
$ hevm symbolic --rpc <NODE_ADDR>  --address 0x7665Fecb393ca45061A6338662751D6cF45A854d --solver cvc4 --sig "mint(address)" --debug --storage-model ConcreteS

hevm: src/EVM/TTY.hs:811:5-38: Non-exhaustive patterns in Just dapp
```

src/hevm/hevm-cli/hevm-cli.hs
MrChico: ok, with `--prefix` instead of `--suffix`, it seems to work

MrChico:
> `Pressing p after reacing a REVERT pretty consistently results in a crash with:`

I made sure `n` doesn't try to step past the execution. Lmk if this error persists.

> I get crashes when pressing ctrl-n while Executing 0x7665Fecb393ca45061A6338662751D6cF45A854d using rpc

happens everytime one tries to do `ctrl-n` without source. Smaller example: `hevm exec --code 6000 --debug`. Looks unrelated to this PR but there might be an easy fix here

MrChico:
The `ctrl-n` error should also be taken care of now @xwvvvvwx. Good catch

xwvvvvwx:
There seems to be a ~4x performance regression when using the `hevm` from this branch to execute the tests in [`uniswap-v2-test`](https://github.com/dapp-org/uniswap-v2-test).

running `./test` takes ~30s on my machine usinng the provided `shell.nix` and ~110s using the dapptools from this branch

xwvvvvwx:
I get the following error when exploring [`dai.sol`](https://via.hypothes.is/https://github.com/makerdao/dss/blob/master/src/dai.sol) in `dss`:

```
$ hevm symbolic --code $(cat out/dapp.sol.json | jq '.contracts."src/dai.sol:Dai"."bin-runtime"' | tr -d '"') --solver cvc4 --json-file out/dapp.sol.json
hevm: Maybe.fromJust: Nothing
```

MrChico:
> I get the following error when exploring `permit` from [`dai.sol`](https://via.hypothes.is/https://github.com/makerdao/dss/blob/master/src/dai.sol) in `dss`:

Right. This is from calling `ECREC` which is not supported for symbolic arguments, currently. I have listed this as a known shortcoming in the blogpost, but let me see if I can make it not hard crash.

> There seems to be a ~4x performance regression when using the hevm from this branch to execute the tests in uniswap-v2-test.

This is pretty unfortunate. I wonder if this is mainly the fuzz tests that are slower or what might be up...

xwvvvvwx:
> This is from calling ECREC which is not supported for symbolic arguments

Ah yes of course.

>  I wonder if this is mainly the fuzz tests that are slower or what might be up...

Not only the fuzz tests, but maybe they are suffering more than the unit tests. Running the `dss` test suite takes ~7s with `master` and ~17s with this branch for me...

MrChico:
More graceful failure for ECREC now:
```
hevm symbolic --code $(cat out/dapp.sol.json | jq '.contracts."src/dai.sol:Dai"."bin-runtime"' | tr -d '"') --solver cvc4 --json-file out/dapp.sol.json --sig "permit(address,address,uint256,uint256,bool,uint8,bytes32,bytes32)"
checking postcondition...
Q.E.D.
Explored: 3 branches without assertion violations
However, 1 branch(es) errored while exploring:
["Unexpected symbolic argument at opcode: OpStaticcall. Not supported (yet!)"]
```

xwvvvvwx:
I'm curious why we only seem to notice a slowdown when running unit tests, but not when running the blockchaintests... 🤔

MrChico:
Yeah, that's a little odd. I suggest we maybe let that be for now, and let future PRs optimize things. We could take a strategy of letting memory and calldata and stuff come in two flavors, so that we can use `ByteString` when things are concrete rather than `[Word 8]`. That should be faster

MrChico:
We can add hevm to https://github.com/ewasm/benchmarking to get some more reliable measurements

rainbreak:
How much would you prioritise the performance regression after merging? 4x is quite a lot.

xwvvvvwx:
definitely agree that optimisations shouldn't block this PR.

MrChico:
I took a stab at implementing some more performant ways of dealing with symbolic vs. concrete byte buffers, which give much better results for concrete execution:
Running some tests from https://github.com/MrChico/evmbench:
```
time ./run_geth.sh && time ./run_hevm.sh && time ./run_buffer_hevm.sh && time ./run_master_hevm.sh
0x37b7277fc606556160f9bc28b06fd55f4424d9cc000000000000000000000000
0x0f5176527280b8e3fa69a6c14ce1f759d6e9c67c000000000000000000000000
0xa94d7bf363f32a5a5b6e9f71b2edaa3f2ae31a61000000000000000000000000
./run_geth.sh  0.29s user 0.09s system 73% cpu 0.512 total
0x37b7277fc606556160f9bc28b06fd55f4424d9cc000000000000000000000000
0x0f5176527280b8e3fa69a6c14ce1f759d6e9c67c000000000000000000000000
0xa94d7bf363f32a5a5b6e9f71b2edaa3f2ae31a61000000000000000000000000
./run_hevm.sh  22.19s user 6.35s system 311% cpu 9.150 total
0x37b7277fc606556160f9bc28b06fd55f4424d9cc000000000000000000000000
0x0f5176527280b8e3fa69a6c14ce1f759d6e9c67c000000000000000000000000
0xa94d7bf363f32a5a5b6e9f71b2edaa3f2ae31a61000000000000000000000000
./run_buffer_hevm.sh  6.08s user 1.48s system 233% cpu 3.233 total
0x37b7277fc606556160f9bc28b06fd55f4424d9cc000000000000000000000000
0x0f5176527280b8e3fa69a6c14ce1f759d6e9c67c000000000000000000000000
0xa94d7bf363f32a5a5b6e9f71b2edaa3f2ae31a61000000000000000000000000
./run_master_hevm.sh  3.17s user 0.88s system 238% cpu 1.695 total
```
where `buffer_hevm` is as of `59e7475`, `run_hevm` is `ez32c838` and master is master. Haven't tried dss with this new setup, but curious to see the numbers

rainbreak:
dss `hevm dapp-test` times:

```
hevm-master@48307b16
real	0m3.955s
user	0m8.710s
sys	0m5.453s

hevm-no-buffer@e32c8389
real	0m11.927s
user	0m25.968s
sys	0m15.917s

hevm-buffer@65cff613
real	0m4.867s
user	0m10.532s
sys	0m5.836s
```

PR #363, opened 2020-04-29
Hevmrpc

MrChico:
Factors out the `--rpc` related stuff from #353, as that has grown to a more ambitious project so we can get this in faster.

Needs some testing still (I'm curious about these `step halted unexpectedly` or other TTY type of crashes), but hey, it typechecks


livnev:
Thanks for this. I will test this for a few days/weeks.

livnev:
I've tested this a fair bit, and didn't notice any issues. IMO it could be merged.

PR #364, opened 2020-04-30
fix loadContract

incertia:
so that contracts with `InitCode` may be ran normally without changing it to `RuntimeCode`

before, we initialized a contract was by

1. creating a contract with `RuntimeCode constructorCode`
2. exec'ing the constructor
3. replacing the contract with `InitCode ""`
4. call `replaceCode` on the target contract with the constructor result

this change lets us change 1 to `InitCode constructorCode` and omitting step 3.


MrChico:
Yeah, the current way things are set up looks like a mistake to me, actually. Thanks!

PR #368, opened 2020-05-04
seth calldata hack

MrChico:
fixes the notorious `seth calldata f(bool) false` issue @lucasvo

### DIFF
--- a/src/seth/libexec/seth/seth-calldata
+++ b/src/seth/libexec/seth/seth-calldata
@@ -12,7 +12,14 @@ if (process.argv.length >= 3 &&
   const sig = process.argv[2];
   try {
     const funcs = new ethers.utils.Interface(["function " + sig.replace(')(', ') returns (')]).functions;
-    const args = process.argv.slice(3).map(s => s.match(/\[.*\]/) ? JSON.parse(s) : s)
+    const args = process.argv.slice(3).map(s => {
+      try {
+        return JSON.parse(s)
+      }
+      catch (e) {
+        return s
+      }
+    })
     console.log(funcs[Object.keys(funcs)[0]].encode(args))
   } catch (e) {
     console.error(e.toString())


src/seth/libexec/seth/seth-calldata
```diff
@@ -12,7 +12,14 @@ if (process.argv.length >= 3 &&
   const sig = process.argv[2];
   try {
     const funcs = new ethers.utils.Interface(["function " + sig.replace(')(', ') returns (')]).functions;
-    const args = process.argv.slice(3).map(s => s.match(/\[.*\]/) ? JSON.parse(s) : s)
+    const args = process.argv.slice(3).map(s => {
+      try {
+        return JSON.parse(s)
+      }
+      catch (e) {
+        return s
```
asymmetric, line 20: Would `s` be an error message here?

manuelpolzhofer:
@MrChico thanks for the quick fix.

src/seth/libexec/seth/seth-calldata
MrChico: no, `s` is an argument that has been passed to this function. I'm still not 100% sure this is the right way to go because the way that javascript takes in cli args is really weird. It seems to forget any quotation marks and treat everything as a string, which means that if you pass `false`, it treats it as the string `'false'`, which if you use it in a conditional is `true`. So instead if one does a `JSON.parse` of the arguments, you turn `'false'` to `false` and quoted lists back into lists and quoted numerals back into numerals but if you had a string to begin with, you can't interpret it as a javascript object because it's just some string.

This blob tries to decode every argument using `JSON.parse` and if it doesn't work, it just keeps the argument as it.

I have no idea if this is the right way of doing things. I hope not. But it solves the immediate and very worrying concern with `bool`s, and it seems to give expected results for lists, strings and numerals as well (with the few examples I tried)

manuelpolzhofer: wow. javascript at it's best.

asymmetric: We need teeeeests ;)

MrChico: we dooooo

PR #374, opened 2020-05-23
Abiencodedecode

MrChico:
- Uses hevm for abi encoding
- hevm now a dependency of seth
- omg tests!?


MrChico:
Happy for feedback and more test cases to be added, I'm tired of this stuff breaking

src/dapp-tests/run-tests
```diff
@@ -3,3 +3,162 @@ export SOLC_FLAGS="--evm-version constantinople"
...
+}
+
+test_calldata_1() {
+    local output
+    output=$(seth --to-uint256 1 )
+    [[ $output -eq "0x0000000000000000000000000000000000000000000000000000000000000001" ]] || error
+}
+test_calldata_1
+
+test_calldata_2() {
```
asymmetric, line 19: I think it would help if the function names were more descriptive.

src/dapp-tests/run-tests
```diff
@@ -3,3 +3,162 @@ export SOLC_FLAGS="--evm-version constantinople"
 for x in *.sol; do
   dapp test-standalone "$x"
 done
+
+error() {
+    printf "fail: function '%s' at line %d.\n" "${FUNCNAME[1]}"  "${BASH_LINENO[0]}"
```
asymmetric, line 8: Maybe it would make sense if this went to stderr with `1>&2`?

src/seth/libexec/seth/seth-calldata
```diff
@@ -1,43 +1,36 @@
...
+### If <sig> has the form `<name>(<types>)', pack <sig>/<args> into calldata.
+###
+### If <file> contains a slash character, treat it as a file name to read.
+### Otherwise, ensure <data> begins with `0x' and treat it as hexdata.
+set -e
+if [[ $1 =~ ^([^\(]+)\( ]]; then
+  name=${BASH_REMATCH[1]}
+  abi=$(seth --abi-function-json "$1" | jq .[0])
+  shift
+  for arg; do
```
asymmetric, line 17: What is `arg` here?

src/seth/libexec/seth/seth-calldata
```diff
@@ -1,43 +1,36 @@
...
+    [[ $arg = @* ]] && arg=$(seth lookup "$arg")
+    args+=" --arg \"${arg//\"/\\\"}\" "
+  done
+  abi=$(echo "${abi//\"/\\\"}" | tr -d '[:space:]')
+  set -x
+  calldata=$(
+    eval "hevm abiencode --abi \"${abi}\" --name $name $args"
```
asymmetric, line 24: Do we really need `eval` here? Can't we just capture the output of the call?

MrChico: I was not able to preserve the necessary quotes around `abi` and the contents of `args` when doing a direct invocation

livnev: Also, do you really want `set -x` in there?

asymmetric: `calldata=$(hevm abiencode --abi "${abi}" --name $name $args` doesn't work?

src/seth/CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.9.0] - 2020-05-25
```
asymmetric, line 7: Please don't forget to push a tag once this is merged :)

src/seth/libexec/seth/seth-calldata
asymmetric: Anyway, I'm a bit against `eval` but if you don't wanna spend too much time finding another way, I'm ok with that.

src/seth/libexec/seth/seth-calldata
MrChico: the arguments given to `seth calldata` after the signature. Don't really know how this works, it's mainly adapted from the [previous version of `seth calldata` ](https://github.com/dapphub/dapptools/blob/78508c6a8db2d6d3e8e09437dbe122bb5e6b2e7e/src/seth/libexec/seth/seth-calldata)

src/seth/libexec/seth/seth-calldata
MrChico: > `calldata=$(hevm abiencode --abi "${abi}" --name "$name" "$args"` doesn't work?

Not directly. There is probably a better way to do this, but I'm not proficient enough in bash to know it.

removed `set -x`

src/seth/libexec/seth/seth-calldata
asymmetric: OK this is [apparently](https://www.gnu.org/software/bash/manual/html_node/Looping-Constructs.html) a horrible Bash special-caseism.

I would rather be explicit here and have `in "$@"`.

src/dapp-tests/run-tests
MrChico: yeah, that's probably better

PR #377, opened 2020-05-27
Remove submodules and simplify installation

### LINKED ISSUE #94: Document "pkg update"
```
❯ dapp pkg
dapp package manager...
current directory: /home/asymmetric/.dapp/dapptools

Options:
  init
  list
  install <package>
  uninstall <package>
  selfupdate
```

`dapp pkg update` is actually supported, but not mentioned above, which can be confusing.

### LINKED ISSUE #123: Move away from ~/.dapp
https://0x46.net/thoughts/2019/02/01/dotfile-madness/

**TL;DR**: move config to `XDG` directories. This could be a breaking change.

* Seth [already does](
https://github.com/dapphub/dapptools/tree/dfbbfddc5f199eb7c4c2df1cc60503a6cd40aaaa/src/seth/libexec/seth/seth#L84)
* dapp [doesn't](
https://github.com/dapphub/dapptools/tree/dfbbfddc5f199eb7c4c2df1cc60503a6cd40aaaa/src/dapp/libexec/dapp/dapp#L11) - this could be `XDG_DATA_HOME`

### LINKED ISSUE #376: dapptools installer should not install Nix by default
On a system that I already had Nix installed on, the dapptools installer (via the `curl` command in the README) installs Nix by default, and removed my existing Nix install.

This is problematic for several reasons:

-   The dapptools installer requires `sudo` access by default because the nix installation path it uses requires `sudo` access.
-   The user may already have Nix installed and likely wants to just use that existing install of Nix.

I see that there is a "manual" install available, but I think this should be the default behavior of the installer.

At the very least, the installer can be given a parameter `--no-nix-install` or something, so that the user can do something like `curl https://dapp.tools/install | bash -s -- --no-nix-install`.

xwvvvvwx:
The install script is kind of magical and has a non trivial support burden. This PR adopts the suggestion from @asymmetric in https://github.com/dapphub/dapptools/issues/376#issuecomment-634312047 and:

1. removes the maker specific submodules
2. adopts a pure nix based approach for the recommended install procedure

Still TODO:

- [x] deprecate `dapp-pkg`
- [x] update `dapp-nix-run` to remove its reliance on the git repo at `$DAPPTOOLS`
- [x] deal with submodules in ds-test #409
- [x] implement migration paths for users who installed dapptools using the install script
- [ ] update dapp CHANGELOG

### on [dapp-tools](https://github.com/dapphub/dapp-tools/pull/17)
- [x] change [install script](https://github.com/dapphub/dapp-tools/blob/master/install)
- [x] update website

Closes #376.
Closes #123.
Closes #94.


README.md
```diff
@@ -12,55 +12,25 @@ Contents:
...
+```
 
-    git clone --recursive https://github.com/dapphub/dapptools $HOME/.dapp/dapptools
-    nix-env -f $HOME/.dapp/dapptools -iA dapp seth solc hevm ethsign
+Then install dapptools:
 
-which will build and install these tools into your Nix profile, optionally
-fetching the binaries from the cache if you have configured it.
+```
+nix-env -A seth -A dapp -A hevm \
```
MrChico, line 29:
```suggestion
nix-env -iA seth dapp hevm \
```

MrChico: Traditionally we also include solc, but leaving it out seems okay too

MrChico: If we do install it, we should at least check if the user already has it to make sure we are not silently overwriting previous versions

asymmetric:
The workflow in dapp-pkg uses `DAPPTOOLS`, which at this point doesn't have a default location anymore, so it will break. I don't know if anyone's really using it though, and I'd consider removing it altogether.

`DAPPTOOLS` is also used [here](https://github.com/dapphub/dapptools/blob/2a08cbe1ed9f5d94efc76d0848164cd4dbdb1fa5/src/dapp/libexec/dapp/dapp---nix-run#L19), which we need for `dapp---use`.

xwvvvvwx:
Regarding `dapp --nix-run`, can we just replace the `nix run "(with import $DAPPTOOLS {}; $expr)" -c "$@"` [here](https://github.com/dapphub/dapptools/blob/master/src/dapp/libexec/dapp/dapp---nix-run#L19) with `nix run -f https://github.com/dapphub/dapptools/tarball/master  $expr -c "$@"` ?

xwvvvvwx:
regarding `dapp pkg` I think we can keep the `install` / `update` commands and have them run:

```
nix-env -A "$@" \
  -if https://github.com/dapphub/dapptools/tarball/rm-submodules \
  --substituters https://dapp.cachix.org \
  --trusted-public-keys dapp.cachix.org-1:9GJt9Ja8IQwR7YW/aF0QvCa6OmjGmsKoZIist0dG+Rs=
```

README.md
xwvvvvwx: > Traditionally we also include solc, but leaving it out seems okay too

I think that I would prefer to leave it out, exactly to avoid having to check to see if it is installed already.

asymmetric:
> Regarding `dapp --nix-run`, can we just replace the `nix run "(with import $DAPPTOOLS {}; $expr)" -c "$@"` [here](https://github.com/dapphub/dapptools/blob/master/src/dapp/libexec/dapp/dapp---nix-run#L19) with `nix run -f https://github.com/dapphub/dapptools/tarball/master $expr -c "$@"` ?
> 
> This would always use latest master instead of the version of dapptools installed locally, not sure if that's a problem or not....

I think it would be confusing if it hit the network and used a different version than the one the user (or the script) has installed. It could make reproducibility and debugging harder too, IMO. (And it would not work offline I guess? Not sure how the `fetchTarball` caching mechanism works)

> regarding `dapp pkg` I think we can keep the `install` / `update` commands and have them run:
> 
> ```
> nix-env -A "$@" \
>   -if https://github.com/dapphub/dapptools/tarball/rm-submodules \
>   --substituters https://dapp.cachix.org \
>   --trusted-public-keys dapp.cachix.org-1:9GJt9Ja8IQwR7YW/aF0QvCa6OmjGmsKoZIist0dG+Rs=
> ```

Honestly, I would try to remove as many layers of magic as possible, as long as the traditional Nix way is easy enough. In this case, considering that installing/updating is a one-liner, I would probably just have users do that. This way, they learn something about Nix in the process.

xwvvvvwx:
> I think it would be confusing if it hit the network and used a different version than the one the user (or the script) has installed. It could make reproducibility and debugging harder too, IMO. (And it would not work offline I guess? Not sure how the fetchTarball caching mechanism works)

Agree, implemented the approach you suggested in your first comment (walking up the directory tree relative to find `default.nix`).

> Honestly, I would try to remove as many layers of magic as possible, as long as the traditional Nix way is easy enough. In this case, considering that installing/updating is a one-liner, I would probably just have users do that. This way, they learn something about Nix in the process.

I think I agree in the end, although it is quite a long one liner. Removed `dapp-pkg` for now.

xwvvvvwx:
Hmmmm, seems I was a little hasty, the approach in https://github.com/dapphub/dapptools/pull/377/commits/5cfe5155b95d8ecd4e8ee1f28aa8e1e90247fafe doesn't work. `dapp` is installed into it's own store path, and so can't see the `default.nix` at the root of the `dapptools` repo.

asymmetric:
> Hmmmm, seems I was a little hasty, the approach in [5cfe515](https://github.com/dapphub/dapptools/commit/5cfe5155b95d8ecd4e8ee1f28aa8e1e90247fafe) doesn't work. `dapp` is installed into it's own store path, and so can't see the `default.nix` at the root of the `dapptools` repo.

Right, of course :man_facepalming: 

> Would it make sense to copy the dapptools source into the `dapp` derivation?

So, I guess what we want to express is the fact that the `dapp` program has a dependency on dapptool's source. Maybe there's a way to model that using Nix?

I guess we could create a derivation for the source, make it a dependency of `dapp`, and provide this path as an env var to dapp?

xwvvvvwx:
Do we need `solc` anywhere except `dapp`? Maybe we can move that into the `dapp` output? I think we can have a `solc.nix` under `src/dapp`, but I think then we would need two pins of `nixpkgs`?

asymmetric:
I don't know if I fully understand what you want to do, but IIUC one thing you're proposing is to get rid of `dapp---nix-run` and only use `dapp---use` for overriding the solc version, right? If so, it would be a bit sad to lose generality, even though we don't really use it in practice. :thinking:

asymmetric:
@xwvvvvwx [this](https://github.com/dapphub/dapptools/compare/rm-submodules...dapptools-src?w=1) is what I came up with.

xwvvvvwx:
Your version looks good to me @asymmetric. 🙏 ✨

nix/dapptools-src.nix
```diff
@@ -0,0 +1,12 @@
+{ stdenv }:
+
+stdenv.mkDerivation {
+  name = "dapptools-src";
+  src = ../.;
+
+  phases = [ "unpackPhase" "installPhase" ];
```
xwvvvvwx, line 7: why do we need `unpackPhase`?

xwvvvvwx:
I'm pretty happy with this. Wondering if there are any potential issues when existing users upgrade? Also not sure if we should be telling people to install from `master` or instead point to the latest release?

nix/dapptools-src.nix
asymmetric: To move the source to the build directory.

asymmetric:
> not sure if we should be telling people to install from master or instead point to the latest release

@xwvvvvwx if we put a release number there, our release process just got a nice new step, and I don't think anybody wants that :joy: 

Is there a way to get a tarball of the latest release? I tried `archive/latest.tar.gz` and `tarball/latest` but they didn't work unfortunately.

asymmetric:
> Wondering if there are any potential issues when existing users upgrade? 

@xwvvvvwx yeah I guess users with a `~/.dapp/dapptools` directory will need some instructions. Something like:

1. remove all installed tools with `nix-env -e`
2. remove `~/.dapp/dapptools`
3. unset `DAPPTOOLS` if you had it
4. install via `nix-env -i`

Should we provide a script (at least for the first 3)?

asymmetric:
It could also be nice to mention how to shorten the invocations by adding substituters to `~/.config/nix/nix.conf`, either manually or via cachix.

Is there no way in nix to programmatically modify that file without cachix?

README.md
asymmetric: Usually, overlays choose different attribute names (e.g. `emacsUnstable` or `firefox-nightly`), we might want to do the same.

The binary would still be available as `solc`, this only affects how you install it.

asymmetric: But I guess that's a [separate issue](#380).

xwvvvvwx:
> Is there no way in nix to programmatically modify that file without cachix?

I don't know how to do it without cachix, although some `nix run nixpkgs.cachix ...` one liner isn't too bad?

> Should we provide a script (at least for the first 3)? Maybe a gist we can easily post in the chat and/or link from the README?

Wondering if it makes sense to reinstate `dapp pkg install` and put this logic there?

> Is there a way to get a tarball of the latest release? I tried archive/latest.tar.gz and tarball/latest but they didn't work unfortunately.
>
> There is a redirect for the latest release at releases/latest but not for thelatest tarball AFAICT.

Turns out we aren't the only ones who wanna do this :grimacing:: https://gist.github.com/steinwaywhw/a4cd19cda655b8249d908261a62687f8

We can do `https://github.com/[user]/[repo]/releases/latest/download/[asset-name.tar.gz]` as long as `[asset-name.tar.gz]` is static (e.g. `release.tar.gz`). Right now the asset names differ for every release.

Otherwise we have to have make some little shell pipeline to figure out the url afaict.

I'm OK with static artifact names for the releases fwiw.

asymmetric:
### On ensuring a smooth transition

The fact that the env var is set in the wrapper ensures that the existing value (set by the users in their shell's init) is superseded - meaning that there's nothing users **need** to do to make this work. It's still good to let them know they can remove `DAPPTOOLS` from their env, but not strictly necessary.

### On which version to point users to

I think ideally it would be `latest.tar.gz` and not `master.tar.gz`, but the only way to get latest seems to be via the API, which requires lots of steps to parse its output.

Maybe we can setup a script that users can curl, and that does all the intermediate steps and returns the latest tarball? This would become the new https://dapp.tools/install.

Otherwise, I'd just go with `master.tar.gz`, and ensure that it always works.

### Binary caches

We could mention `nix.binaryCaches` or `nix.trustedBinaryCaches`?

Either way I would say we can iterate on this without blocking this PR, since it's mostly a QOL improvement.

MrChico:
I've had a smooth experience using this so far

MrChico:
Although, #374 introduced `ds-test` as another submodule. But this should be fine, right? As it's not referred to anywhere except in the ci `dapp-test`s?

asymmetric:
I'm actually not sure. We should make sure the submodule is cloned at build-time, as it won't be possible to do it afterwards.

asymmetric:
This has been unblocked by #404.

asymmetric:
I think we should use GH actions to create a release for each tag, so that we can then fetch the latest release via API in a script users run as https://dapp.tools/install.

This is because it's not possible to get the latest tag.

xwvvvvwx:
I think that is a good way forward.

@asymmetric what do you think about adding `dapp pkg` back and having it just print an error message and a link to the new install instructions?

asymmetric:
> @asymmetric what do you think about adding `dapp pkg` back and having it just print an error message and a link to the new install instructions?

Yeah that's a good idea. And maybe after a couple months we remove it altogether?

asymmetric:
Do we have to do anything on the Dockerfiles? They seemed pretty messy to me, not sure if it's worth maintaining them though.

xwvvvvwx:
I think we should rm the dockerfiles. If we want to provide container images we should produce them with nix using `buildDockerImage`.

src/dapp/libexec/dapp/dapp-pkg
```diff
@@ -1,60 +1,7 @@
...
-    echo "  init"
-    echo "  list"
-    echo "  install <package>"
-    echo "  uninstall <package>"
-    echo "  selfupdate"
-  }
-esac
+cat <<EOF
+This command has been deprecated. dapptools no longer uses git to install and manage dependencies.
+Please use nix instead <3.
```
xwvvvvwx, line 6: Maybe smth like this is more informative?

```suggestion
This command has been deprecated. Please check the readme at https://github.com/dapphub/dapptools for new install instructions.
```

asymmetric:
Oops forgot the CHANGELOG :grimacing:

Issue #383, opened 2020-06-07
NixError on MacOS: dyld: malformed mach-o image


chrismaree:
Hi! I've been having a lot of trouble getting the dapphubs took kit installed. I have tried both the easy install and the Manual install from the repo but always get this error with seth:
```
seth ls
dyld: malformed mach-o image: segment __DWARF has vmsize < filesize

```

I have googled it to the end of the earth and the only thing I really found was a connection to a Go error on a specific version. I've tried updating my go but still cant get past this. any help would be super helpful.

I'm on MacOS Catalina & I have Go installed via home brew.

Thanks!

xwvvvvwx:
Do you only see this error with `seth ls` or does it occur with all `seth` commands?

chrismaree:
> Do you only see this error with `seth ls` or does it occur with all `seth` commands?

I saw this with a bunch of commands. `seth send` for example did the same thing. I was able to run `seth call` without this error occurring.

xwvvvvwx:
interesting....

Could you try running this:

```
nix run -f $DAPPTOOLS ethsign --command ethsign
```

chrismaree:
> interesting....
> 
> Could you try running this:
> 
> ```
> nix run -f $DAPPTOOLS ethsign --command ethsign
> ```

that env variable is not set on my machine so I got:
```
error: getting status of '/Users/chris/ethsign': No such file or directory
```

I was able to simply run `ethsign --command ethsign` and got the same error:
```
ethsign --command ethsign
dyld: malformed mach-o image: segment __DWARF has vmsize < filesize
[1]    45792 abort      ethsign --command ethsign
```

MrChico:
@xwvvvvwx this is some obscure golang error which doesn't exist on the `symbolic` branch. I think bumping nixpkgs and factoring out the changes that are made to ethsign in that branch is the key here

MrChico:
fixed by #384

PR #384, opened 2020-06-08
backporting fix to ethsign from symbolic

realisation:
Wholesale copy of default.nix and src/ethsign from the symbolic branch. 

Installed with following commands from a fresh clone after replacing default.nix and src/ethsign: 
```
git submodule init
git submodule update --recursive 
nix-env -i ethsign -f . 
```

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -3,8 +3,8 @@
 let
   nixpkgs = builtins.fetchGit {
     url = "https://github.com/nixos/nixpkgs";
-    ref = "release-19.03";
-    rev = "f1707d8875276cfa110139435a7e8998b4c2a4fd";
+    ref = "release-19.09";
+    rev = "54e89941c303687a6392a3264dbe1540fe3381d8";
   };
 in
   # Now return the Nixpkgs configured to use our overlay.
--- a/src/dapp/Makefile
+++ b/src/dapp/Makefile
@@ -16,6 +16,6 @@ test:
 	for x in libexec/*/*; do [[ -f $$x ]] && files+=($$x); done; \
 	! grep '^#!/bin/sh' "$${files[@]}"; \
 	grep '^#!/usr/bin/env bash' "$${files[@]}" | \
-	cut -d: -f1 | xargs shellcheck
+	cut -d: -f1 | xargs shellcheck --exclude=2001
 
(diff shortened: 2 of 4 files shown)


src/ethsign/CHANGELOG
```diff
@@ -4,15 +4,6 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.14.0] - 2020-05-29
```
asymmetric, line 7: Why are we removing this?

If the changes have been reverted, we should mention that instead.

MrChico:
Probably better to cherry-pick the relevant commits, because there was a recent change merged to master (#378 )

MrChico:
I also found this nixpkgs bump to be less dramatic when we keep the old nixpkgs for `geth`. See the `oldnixpkgs` related sections of `overlay.nix` on the symbolic branch

asymmetric:
What's the relationship between the nixpkgs bump and the changes in ethsign? Could we do them saparately (and maybe bump to 20.03)?

MrChico:
I think the geth libraries on which ethsign depends are updated along with the nixpkgs bump, and there were some minor changes needed. Could also go with 20.03. This is mainly to address #383

realisation:
I just cherry picked the change to ethsign from the symbolic branch and updated default.nix manually since that commit was bundled up with many other changes on the symbolic branch.

I thought about bumping nixpkgs to 20.3, but that appears to involve a change to "rev" in default.nix and I'm not sure where to retrieve that. 

What is that all about?

realisation:
I just cherry picked the change to ethsign from the symbolic branch and updated default.nix manually since that commit was bundled up with many other changes on the symbolic branch.

I thought about bumping nixpkgs to 20.3, but that appears to involve a change to "rev" in default.nix and I'm not sure where to retrieve that. 

What is that all about?

MrChico:
I think we can do the bump to 20.03 later!

MrChico:
Sweet! Do you think this needs a version bump @asymmetric ?

MrChico:
Thanks for the fix @realisation !

PR #386, opened 2020-06-09
*: unify version numbers

asymmetric:
Move away from semantic versioning, and simplify release process.


src/seth/default.nix
```diff
@@ -18,6 +18,10 @@ stdenv.mkDerivation rec {
         else ""}
   '';
 
+  postFixup = ''
```
asymmetric, line 21: This should probably be `postInstall` and the current `postInstall` should be `postFixup`.

erieza09za08: Xtuocgf bpda xopn

rainbreak:
Would be nice if the `--version` output both the version number and the commit hash. This would help with 1) easy checkout and 2) disambiguation if I build locally from an unbumped development version.

asymmetric:
Sidenote: I think `--version` should actually be `version`, since it's not a flag to another command but a command in itself.

rainbreak:
I think `--version` is more expected from other cli tools, so it should at least alias to `version` (e.g. `git` allows both). Also we have `seth --to-wei` etc. as commands rather than flags.

PR #390, opened 2020-06-11
travis: install cachix from upstream

asymmetric:
So that our Haskell overrides don't interefere and cause an unnecessary
rebuild.


MrChico:
I like it and I wish it would work but it doesn't

MrChico:
Closing as this is no lonher relevant

PR #391, opened 2020-06-11
Nix solidity package: call solc just one and extract all the files af…

gbalabasquer:
…ter if needed

### DIFF
--- a/nix/solidity-package.sh
+++ b/nix/solidity-package.sh
@@ -1,20 +1,14 @@
 source $stdenv/setup
 unpackPhase
 
-jsonopts=--combined-json=abi,bin,bin-runtime,srcmap,srcmap-runtime,ast
+jsonopts=--combined-json=abi,bin,bin-runtime,srcmap,srcmap-runtime,ast,metadata
 
 export DAPP_SRC=$src
 export DAPP_OUT=out
 
-find "$DAPP_SRC" -name '*.sol' | while read -r x; do
-  dir=${x%\/*}
-  dir=${dir#$DAPP_SRC}
-  dir=${dir#/}
-  mkdir -p "$DAPP_OUT/$dir"
-  (set -x; solc --overwrite $REMAPPINGS --allow-paths $DAPP_SRC $solcFlags --abi --bin --bin-runtime -o "$DAPP_OUT/$dir" "$x")
-  json_file=$DAPP_OUT/$dir/${x##*/}.json
-  (set -x; solc $REMAPPINGS --allow-paths $DAPP_SRC $solcFlags $jsonopts "$x" >"$json_file")
-done
+mkdir -p "$DAPP_OUT"
+mapfile -t files < <(find "${DAPP_SRC}" -name '*.sol')
+(set -x; solc "$REMAPPINGS" "$jsonopts" "$solcFlags" /=/ "${files[@]}" > "${DAPP_JSON}")
 
 mkdir lib
 echo "$LIBSCRIPT" > setup.sh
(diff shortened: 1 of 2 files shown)


nix/solidity-package.sh
```diff
@@ -25,7 +19,26 @@ if [ "$doCheck" == 1 ]; then
...
+    ((count++))
+    name="${path#*:}"
+    contract=$(echo "$data" | jq '.[''"'"$path"'"'']')
+    echo "$contract" | jq '.["abi"]' -r > "$DAPP_OUT/$name.abi"
+    echo "$contract" | jq '.["bin"]' -r > "$DAPP_OUT/$name.bin"
+    echo "$contract" | jq '.["bin-runtime"]' -r > "$DAPP_OUT/$name.bin-runtime"
+    echo "$contract" | jq '.["metadata"]' -r > "$DAPP_OUT/$name.metadata"
+  done
+  echo
+  info "warning: --extract is slow and may have filename collisions. All build data can be found in $DAPP_JSON.\\n"
```
icetan, line 38:
```suggestion
```

nix/solidity-package.sh
```diff
@@ -25,7 +19,26 @@ if [ "$doCheck" == 1 ]; then
...
-cp -r $src $out/dapp/$name/src
-cp -r lib $out/dapp/$name/lib
-cp -r out $out/dapp/$name/out
+if [ "$extract" == 1 ]; then
+  mapfile -t contracts < <(<"$DAPP_JSON" jq '.contracts|keys[]' -r | sort -u -t: -k2 | sort)
+  data=$(<"$DAPP_JSON" jq '.contracts' -r)
+  count=1
+  total=${#contracts[@]}
+  for path in "${contracts[@]}"; do
+    info -ne "Extracting build data... [$count/$total]\\r"
```
icetan, line 28: Don't think `info` is a defined function here.

```suggestion
```

nix/solidity-package.sh
```diff
@@ -1,20 +1,14 @@
...
-  dir=${dir#$DAPP_SRC}
-  dir=${dir#/}
-  mkdir -p "$DAPP_OUT/$dir"
-  (set -x; solc --overwrite $REMAPPINGS --allow-paths $DAPP_SRC $solcFlags --abi --bin --bin-runtime -o "$DAPP_OUT/$dir" "$x")
-  json_file=$DAPP_OUT/$dir/${x##*/}.json
-  (set -x; solc $REMAPPINGS --allow-paths $DAPP_SRC $solcFlags $jsonopts "$x" >"$json_file")
-done
+mkdir -p "$DAPP_OUT"
+mapfile -t files < <(find "${DAPP_SRC}" -name '*.sol')
+(set -x; solc "$REMAPPINGS" "$jsonopts" "$solcFlags" /=/ "${files[@]}" > "${DAPP_JSON}")
```
icetan, line 11: Where is `DAPP_JSON` declared?

nix/solidity-package.sh
```diff
@@ -25,7 +19,26 @@ if [ "$doCheck" == 1 ]; then
...
+  mapfile -t contracts < <(<"$DAPP_JSON" jq '.contracts|keys[]' -r | sort -u -t: -k2 | sort)
+  data=$(<"$DAPP_JSON" jq '.contracts' -r)
+  count=1
+  total=${#contracts[@]}
+  for path in "${contracts[@]}"; do
+    info -ne "Extracting build data... [$count/$total]\\r"
+    ((count++))
+    name="${path#*:}"
+    contract=$(echo "$data" | jq '.[''"'"$path"'"'']')
+    echo "$contract" | jq '.["abi"]' -r > "$DAPP_OUT/$name.abi"
```
icetan, line 32: Need to add a subdirectory with the name of the dapp project.

nix/solidity-package.sh
```diff
@@ -25,7 +19,26 @@ if [ "$doCheck" == 1 ]; then
...
-cp -r out $out/dapp/$name/out
+if [ "$extract" == 1 ]; then
+  mapfile -t contracts < <(<"$DAPP_JSON" jq '.contracts|keys[]' -r | sort -u -t: -k2 | sort)
+  data=$(<"$DAPP_JSON" jq '.contracts' -r)
+  count=1
+  total=${#contracts[@]}
+  for path in "${contracts[@]}"; do
+    info -ne "Extracting build data... [$count/$total]\\r"
+    ((count++))
+    name="${path#*:}"
```
icetan, line 30: This `name` might be overwriting `name` set by the derivation name.

nix/solidity-package.sh
gbalabasquer: Right, I'm hardcoding it to `dapp.sol.json` now. If that is not an option as generates conflicts for different packages we will define something different.

nix/solidity-package.sh
gbalabasquer: Changing it just in case.

nix/solidity-package.sh
gbalabasquer: Yes you are right, I'm simplifying it to just one echo.

gbalabasquer:
Closing this PR as it was actually made by mistake. We will create a new one as soon as this is finished.

PR #397, opened 2020-06-15
add cvc4 derivation and custom sbv_8_4

MrChico:
factored out from #353, as it keeps timing out. Could also be useful on its own if someone is curious to try cvc4 out. It's pretty good on large bitvectors. Not suitable for nixpkgs upstream as it is a nightly version


MrChico:
Ah ok, will open a pr then, but merging here in the meantime

PR #399, opened 2020-06-17
Force `seth` to use local `rpc-url`

pawzar:
We need to make sure that the `rpc-url` is not taken from `.sethrc`.

### DIFF
--- a/src/dapp/libexec/dapp/dapp---testnet-launch
+++ b/src/dapp/libexec/dapp/dapp---testnet-launch
@@ -123,7 +123,7 @@ trap clean EXIT
 
 until curl -s "$ETH_RPC_URL"; do sleep 1; done
 
-ETH_FROM=$(seth rpc eth_coinbase)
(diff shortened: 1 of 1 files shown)


src/dapp/libexec/dapp/dapp---testnet-launch
```diff
@@ -123,7 +123,7 @@ trap clean EXIT
 
 until curl -s "$ETH_RPC_URL"; do sleep 1; done
 
-ETH_FROM=$(seth rpc eth_coinbase)
+ETH_FROM=$(seth --rpc-url=$ETH_RPC_URL rpc eth_coinbase)
```
MrChico, line 126:
```suggestion
ETH_FROM=$(seth --rpc-url="$ETH_RPC_URL" rpc eth_coinbase)
```

pawzar:
@MrChico :+1:

MrChico:
Thanks @pawzar !

PR #400, opened 2020-06-17
Nix solidity package: call solc just one and extract all the files af…

gbalabasquer:
…ter if needed

Co-authored-by: Christopher Fredén <me@icetan.org>

### DIFF
--- a/nix/solidity-package.nix
+++ b/nix/solidity-package.nix
@@ -4,18 +4,20 @@
       (builtins.map
         (x: {
           "${x.name}/" = "${x}/dapp/${x.name}/src/";
+          "${x.name}" = "${x}/dapp/${x.name}/src/index.sol";
          } // x.remappings)
          xs);
   libPaths = xs:
     builtins.foldl' pkgs.lib.mergeAttrs {}
       (builtins.map
         (x: {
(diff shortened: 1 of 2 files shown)


MrChico:
Can you explain usage of this technology?

icetan:
The changes intend to make the nix function `solidityPackage` behave more closely to `dapp build`.

Before `solc` was run twice to extract separate files for `.bin` etc. but is now using the same method as `dapp build --extract`.

MrChico:
Yeah, the changes themselves make sense. I'm just wondering where `solidityPackage` is used. Is this part of `dapp2nix` or something similar?

icetan:
Not really sure where it's used except in `dapp2nix` and here https://github.com/dapphub/ds-test/blob/master/default.nix#L1

I had a thought about copying `solidityPackage` to `dapp2nix` to be more flexible but I guess if it is to be compatible with `dapp` it should be housed along side it.

MrChico:
as we discussed before, I think a `--dapp2nix` flag to `dapp` which assumes that everything is being done through nix is still a good idea, and I'd be happy to accept changes into dapptools that explores this

PR #401, opened 2020-06-18
[draft] Hevm turn cheatcode

livnev:
~This is on top of `sethdebug`, assuming this can be merged later.~ This is now on top of `master`.

### DIFF
--- a/src/hevm/src/EVM.hs
+++ b/src/hevm/src/EVM.hs
@@ -1619,6 +1619,14 @@ cheatActions =
     [ action "warp(uint256)" [AbiUIntType 256] $
         \[AbiUInt 256 x] -> do
           assign (block . timestamp) (w256 (W256 x))
+          return Nothing,
+      action "store(address,bytes32,bytes32)" [AbiAddressType, AbiBytesType 32, AbiBytesType 32] $
+        \[AbiAddress a, AbiBytes 32 x, AbiBytes 32 y] -> do
+          let slot = w256 $ word x
+              new  = w256 $ word y
+          fetchAccount a $ \_ -> do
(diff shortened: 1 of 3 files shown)


src/hevm/src/EVM.hs
```diff
@@ -1625,6 +1625,13 @@ cheatActions =
     [ action "warp(uint256)" [AbiUIntType 256] $
         \[AbiUInt 256 x] -> do
           assign (block . timestamp) (w256 (W256 x))
+          return Nothing,
+      action "store(address,bytes32,bytes32)" [AbiAddressType, AbiBytesType 32, AbiBytesType 32] $
+        \[AbiAddress a, AbiBytes 32 x, AbiBytes 32 y] -> do
+          let slot = w256 $ word x
+              new = w256 $ word y
+          assign (env . contracts . ix a . storage . at slot)
```
MrChico, line 1633: Will this write to the slot if `a` doesnt exist in the `contracts` already? Im not sure. You may wanna fetch the contract first

MrChico: This would make it play nicer with rpc mode

xwvvvvwx:
What is this for?

src/hevm/src/EVM.hs
livnev: Great point, currently this ends up getting overwritten if the contract is fetched later.

livnev: Fixed. I tested it out and indeed now the `hevm.store()` cheatcode commutes with RPC fetch, as desired.

MrChico:
took the liberty to add some documentation to this.

livnev:
Merged. Github doesn't seem to know that though.

PR #404, opened 2020-06-18
Run `dapp-test` through nix-shell in ci

MrChico:
Ensures `dapp-tests` are run through travis, removes the ds-test submoule which should set the stage for #377

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ script:
     [[ "$TRAVIS_OS_NAME" == "osx" ]] && PLATFORM=darwin || PLATFORM=linux
 
     nix-build -j auto release.nix -A dapphub.$PLATFORM.stable &&
+    cd src/dapp-test && nix-shell --command 'make' --pure && cd ../..
 
     # this is here to let cachix finish pushing the last artifact before the script is killed
     sleep 10
--- a/src/dapp-tests/Makefile
+++ b/src/dapp-tests/Makefile
@@ -1,3 +1,7 @@
-all:; ./run-tests
-
+all:;
+	dapp install ds-test
+	./run-tests
+	dapp uninstall ds-test
+	git reset HEAD^^
+	rm -rf lib
 .PHONY: all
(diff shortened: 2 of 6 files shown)


.travis.yml
```diff
@@ -20,6 +20,7 @@ script:
     [[ "$TRAVIS_OS_NAME" == "osx" ]] && PLATFORM=darwin || PLATFORM=linux
 
     nix-build -j auto release.nix -A dapphub.$PLATFORM.stable &&
+    cd src/dapp-test && nix-shell --command 'make' --pure && cd ../..
```
asymmetric, line 23: You can just pass the `shell.nix` path to `nix-shell`, without `cd`ing.

src/seth/default.nix
```diff
@@ -15,8 +15,7 @@ stdenv.mkDerivation rec {
   postInstall =
     let
       path = lib.makeBinPath [
-        bc coreutils curl ethsign git gnused jshon nodejs perl
-        bc coreutils curl ethsign git gnused hevm jshon nodejs perl
+        bc coreutils curl ethsign git gnused hevm jshon nodejs perl jq
```
asymmetric, line 18: Can we sort this bad boy plz?

src/dapp-tests/shell.nix
```diff
@@ -0,0 +1,8 @@
+let dapptools = import ../../. {};
+in dapptools.mkShell {
+  buildInputs = with dapptools; [
+  dapp
```
asymmetric, line 4: Plz indent line

src/dapp-tests/shell.nix
```diff
@@ -0,0 +1,8 @@
+let dapptools = import ../../. {};
```
asymmetric, line 1: I would just call this `pkgs`, since it's actually all of nixpkgs.

src/dapp-tests/Makefile
```diff
@@ -1,3 +1,7 @@
-all:; ./run-tests
-
+all:;
+	dapp install ds-test
+	./run-tests
+	dapp uninstall ds-test
+	git reset HEAD^^
```
asymmetric, line 5: This is not super pretty. It also reinforces the coupling between dapp and git.

I'll see if making a derivation out of `ds-test` works.

MrChico: I wonder if this can be made with the nix function `solidityPaclage`

PR #406, opened 2020-06-19
Travis -> GitHub actions

asymmetric:
Nix support on Travis is currently [broken](https://github.com/travis-ci/travis-build/pull/1833) due to some issues with the macOS builders, that prevent updating the Nix version.

Moving to GitHub actions, apart from fixing that (temporary?) issue, also gives us the benefit that there's more Nix people working on actions than on Travis, so hopefully things won't break again in the future.

Also:

- better integration with GH (d'uh), specifically in relation to secrets and ACLs
- new and shiny

### Outputs are not pushed to Cachix for external PRs

:heavy_check_mark: This works as expected because secrets are [not available](https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#using-encrypted-secrets-in-a-workflow) to builders started from forks, so no push.

### Outputs from all branches are pushed to Cachix

:warning: So anybody who has write access to this repo can push to Cachix. @xwvvvvwx suggested protecting master, but I haven't fully investigated if this would actually help tighten things down.

We could also restrict pushes to Cachix to builds on the master branch, whereas tests would run for each commit on each branch. If we paired this with a smaller set of people allowed to push to master (and therefore merge PRs), then that would reduce the attack surface. Not sure who the people in that smaller set should be though.

### VM disk size limitations

:disappointed: We have to resort to the ugly hack of `rm -rf /opt` for the GNU/Linux builders to work, since their disk is [limited to 14GB](https://github.com/actions/virtual-environments/issues/709). This might be raised in the future, but it could also mean our builds stop working if the closure gets too big. We should investigate reducing it.

### Announces to RocketChat

:shushing_face: Haven't implemented them for now. Does anyone miss them?


MrChico:
I was running into the 2hr time limit on the symbolic branch, I wonder if we will run into the 14 GB limit with it as well

asymmetric:
Which branch is that? You can try rebasing on top of this and see what happens.

MrChico:
> Which branch is that? You can try rebasing on top of this and see what happens.

Tried this and it seems to work super nicely!

MrChico:
Screen Shot 2020-06-20 at 19 08 14
#sad

asymmetric:
@MrChico can you link to a build?

MrChico:
Seems to have been resolved after a restart

MrChico:
I dont see the hevm tests being run this time but I guess thats because everything was cached! It seems super fast

asymmetric:
OK let’s merge this and see how we like it.

PR #409, opened 2020-06-21
Package dapp-tests with Nix

asymmetric:
@MrChico this should be an alternative to using submodules and would close #404.

Draft because I need to clean up the history.

The main issue I had was that `solidityPackage` uses its own bulder, which means it was hard to adapt it to our needs - which are a bit different than usual, since we not only want to test by running `hevm dapp-test` on the contracts but also want to run a bash script.

So I've moved most of its logic to a `buildDappPackage` function, which uses the usual `mkDerivation` infra and doesn't require its own builder, and allows for easier overriding and adding of phases.

Ideally we would remove/alias `solidityPackage`, but I want to make sure it works for all consumers (cc @icetan).


nix/build-dapp-package.nix
```diff
@@ -0,0 +1,75 @@
...
+
+      REMAPPINGS =
+        pkgs.lib.mapAttrsToList
+          (k: v: k + "=" + v)
+          passthru.remappings;
+
+      LIBSCRIPT =
+        pkgs.lib.mapAttrsToList
+          (k: v: ''
+            ln -s ${v} lib/${k}
```
asymmetric, line 45: I think we could probably use `symlinkJoin` instead of this.

src/dapp-tests/default.nix
```diff
@@ -0,0 +1,31 @@
+{ pkgs }:
+
+let
+  ds-test-src = pkgs.fetchFromGitHub {
+    owner = "dapphub";
+    repo = "ds-test";
+    rev = "eb7148d43c1ca6f9890361e2e2378364af2430ba";
```
asymmetric, line 7: Very open to using niv here.

overlay.nix
```diff
@@ -133,6 +118,8 @@ in rec {
   seth = self.callPackage (import ./src/seth) {};
   dapp = self.callPackage (import ./src/dapp) {};
 
+  dapp-tests = self.callPackage (import ./src/dapp-tests) {};
```
MrChico, line 121: I thought you wanted to remove this from the overlay?

asymmetric: That was when we thought it would just be a shell script.

But because these integration tests are both dapp packages and other stuff (like a shell script), then I think it makes sense to use our solidity -> nix infrastructure, run tests with hevm and run anything else we want to run.

src/dapp-tests/default.nix
asymmetric: But maybe in a separate PR, where we use it for the whole repo and not just for this library.

nix/build-dapp-package.nix
```diff
@@ -0,0 +1,75 @@
...
+        export DAPP_LIB=lib
+      '';
+
+      checkPhase = args.checkPhase or ''
+        hevm dapp-test
+      '';
+      installPhase = args.installPhase or ''
+        mkdir -p $out/dapp/$name
+        cp -r $src $out/dapp/$name/src
+        cp -r lib $out/dapp/$name/lib
```
asymmetric, line 72: It could be interesting to do a multiple outputs derivation instead of this, so that they would result in different store paths.

This could mean that if e.g. a dependency doesn't change, that store path remains unchanged.

overlay.nix
```diff
@@ -101,19 +86,19 @@ in rec {
 
   solc-versions =
     let
-      fetchNixpkgs = { owner, attr }:
+      fetchSolcVersions = { owner, attr }:
```
asymmetric, line 89: These changes I've just ninja'd in, they have no relationship to this PR :)

nix/build-dapp-package.nix
```diff
@@ -0,0 +1,76 @@
...
+
+      LIBSCRIPT =
+        pkgs.lib.mapAttrsToList
+          (k: v: ''
+            ln -s ${v} lib/${k}
+          '')
+          passthru.libPaths;
+
+      buildPhase = args.buildPhase or ''
+        jsonopts=--combined-json=abi,bin,bin-runtime,srcmap,srcmap-runtime,ast
```
asymmetric, line 50: NOTE: I noticed this is not taking into account the recent changes introduced in #400, fixing this.

nix/build-dapp-package.nix
```diff
@@ -0,0 +1,88 @@
...
+        (x: {
+          "${x.name}" = "${x}/dapp/${x.name}/src";
+         } // x.libPaths)
+         xs);
+in
+  {
+    name
+  , src
+  , deps ? []
+  , extract ? true
```
MrChico, line 23: does this mean that it defaults to true? I don't think `extract` is commonly used

overlay.nix
MrChico: is it just name changes or does this change the semantics?

asymmetric: Only name

nix/build-dapp-package.nix
asymmetric: Right. I took it from [here](https://github.com/dapphub/dapptools/pull/400/files#diff-a8c475bf1172056e5043d77688ed2e1bR20), but it makes sense that it should default to `false`, probably.

nix/build-dapp-package.nix
MrChico: deduplication?

src/dapp-tests/default.nix
```diff
@@ -0,0 +1,31 @@
...
+    doCheck = false;
+  };
+
+in
+  pkgs.buildDappPackage {
+    name = "dapp-tests";
+    src = ./.;
+    deps = [ ds-test ];
+
+    solcFlags = "--evm-version constantinople";
```
MrChico, line 23: I think it's better to leave the default `evm-version` instead. The current version of solidity should default to Istanbul, methinks

src/dapp-tests/default.nix
```diff
@@ -0,0 +1,31 @@
...
+
+let
+  ds-test-src = pkgs.fetchFromGitHub {
+    owner = "dapphub";
+    repo = "ds-test";
+    rev = "eb7148d43c1ca6f9890361e2e2378364af2430ba";
+    sha256 = "1phnqjkbcqg18mh62c8jq0v8fcwxs8yc4sa6dca4y8pq2k35938k";
+  } + "/src";
+
+  ds-test = pkgs.buildDappPackage {
```
MrChico, line 11: cool!

nix/build-dapp-package.nix
```diff
@@ -0,0 +1,88 @@
...
+            echo "$contract" | jq '.["abi"]' -r > "out/$fileName.abi"
+            echo "$contract" | jq '.["bin"]' -r > "out/$fileName.bin"
+            echo "$contract" | jq '.["bin-runtime"]' -r > "out/$fileName.bin-runtime"
+            echo "$contract" | jq '.["metadata"]' -r > "out/$fileName.metadata"
+          done
+        fi
+      '';
+
+      checkPhase = args.checkPhase or ''
+        DAPP_OUT=out dapp2-test-hevm
```
MrChico, line 79: out is the default value for `DAPP_OUT` FYI

src/dapp-tests/default.nix
MrChico: I could see us using `master` here, to ensure that what's being tested is the same as people's experience using `dapp install ds-test`

src/seth/default.nix
```diff
@@ -6,10 +6,11 @@ stdenv.mkDerivation rec {
   version = "0.9.0";
   src = ./.;
 
+  buildInputs = [ nodejs ];
```
MrChico, line 9: why is this needed? `nodejs` is already added to path in `postinstall`?

src/seth/default.nix
```diff
@@ -6,10 +6,11 @@ stdenv.mkDerivation rec {
...
   nativeBuildInputs = [makeWrapper shellcheck];
-  buildPhase = "true";
+  dontBuild = true;
```
MrChico, line 11: the linking step is not needed? Its being done with `patchShebangs` or?

nix/build-dapp-package.nix
```diff
@@ -0,0 +1,88 @@
...
+          passthru.remappings;
+
+      LIBSCRIPT =
+        pkgs.lib.mapAttrsToList
+          (k: v: ''
+            ln -s ${v} lib/${k}
+          '')
+          passthru.libPaths;
+
+      buildPhase = args.buildPhase or ''
```
MrChico, line 50: Is there any way we could merge this and `dapp build` somehow? This script is essentially `dapp build`. It seems that whenever we change `dapp build` we'd also need to change it here.

nix/build-dapp-package.nix
asymmetric: Where do you see that? The default would have to be set [here](
https://github.com/dapphub/dapptools/tree/15c241a82183da46f5e170350dd378c99e1f264b/nix/dapp/dapp-test-hevm.nix#L5) for it to take effect.

src/seth/default.nix
asymmetric: seth doesn't need linking. At runtime, it calls a file named `seth-$subcommand` in libexec, see [here](
https://github.com/dapphub/dapptools/tree/15c241a82183da46f5e170350dd378c99e1f264b/src/seth/libexec/seth/seth#L174)

src/seth/default.nix
asymmetric: This is needed so that `patchShebangs` finds `noejs` in the `PATH` and therefore patches it.

I'll see if it's possible to remove the repetition, but it could be weird if only `nodejs` is treated differently compared to all the other packages in the `path` variable below.

src/dapp-tests/default.nix
asymmetric: Right?? :)

nix/build-dapp-package.nix
asymmetric: Right, maybe for `src` we would get some deduplication in some (rare?) cases. Not for `lib` though, there are only symlinks in there.

I think it's pretty cool aesthetically so we should probably do it anyway

nix/build-dapp-package.nix
asymmetric: dapp2-test-hevm is not in `libexec/`, meaning the `libexec/dapp` script is not run, so variables exported there have no effect.

src/dapp-tests/default.nix
asymmetric: WDYM with using master? I took the last commit on master on the repo, and we have to pin one commit for this to be reproducible.. What do you have in mind?

src/seth/default.nix
asymmetric: Or what linking step are you referring to?

nix/build-dapp-package.nix
asymmetric: Hmm yeah, that's a good point.

Maybe we could extract the bash stuff in the different phases into their own little script, and use those both in `build-dapp-package` and in `dapp-build`.

Is it OK if I file an issue for this or do you think it should be in this PR?

asymmetric: Also honestly, inlined Bash inside of Nix is pretty gross. It would be good to move it out for that reason too.

nix/build-dapp-package.nix
MrChico: Im wrobg

src/seth/default.nix
MrChico: I'm referring to what was hapenning in `make install`

nix/build-dapp-package.nix
MrChico: Yeah, we can def postpone this for a later pr

MrChico:
This lgtm. Commits will be squashed when we merge anyway

nix/build-dapp-package.nix
asymmetric: What??

asymmetric:
No let’s not squash them please, it would make the PR too over the place IMO.

icetan:
@asymmetric very cool, I see no reason for this not replacing `solidityPackage`.

nix/build-dapp-package.nix
MrChico: I retract my previous statement

src/dapp-tests/default.nix
MrChico: I had in mind not pinning it, making it non reproducible with the idea that whats being tested here is the same thing that people will be using

src/seth/default.nix
asymmetric: But the reason why patching the `#!/usr/bin/env node` shebang is necessary for seth used in `dapp-tests` (and isn't necessary for seth used from the shell to work) is not clear to me.

src/dapp-tests/default.nix
asymmetric: You can't make it non-repro, AFAIK.

src/seth/default.nix
asymmetric: The `buildPhase` by default [just runs `make`](https://nixos.org/nixpkgs/manual/#build-phase), which in our case is `test` - so it was running tests, which I found weird.

asymmetric: The `buildPhase` above was just running `true` btw, as in the program `true`.

PR #415, opened 2020-06-26
hevm: Remove `ExecMode`, always execute as blockchaintest

MrChico:
This makes `hevm exec` (and other commands) finalize transactions, which ensures that bytecode is replaced at creations, coinbase is paid and so on. 

`VMTest` cli endpoint is removed. I do not believe they have been run in quite some time, and what's being tested in VMTests is a proper subset of what's being done in `Blockchaintests` anyway.


MrChico:
Changed base to `sethdebug` for a cleaner diff, and because I'd like to merge `sethdebug` before this

PR #417, opened 2020-06-26
Dapptests

MrChico:
Based on `rm-submodules`.
Draft as I'm stuck

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,7 +29,7 @@ jobs:
       - uses: cachix/install-nix-action@63cf434de4e4292c6960639d56c5dd550e789d77
         with:
           # we use the nixpkgs defined in default.nix
-          skip_adding_nixpkgs_channel: true
+          skip_adding_nixpkgs_channel: false
       # v6
       - uses: cachix/cachix-action@490a246fbc7f92208d309eeb54383a4d828cedc1
         with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,5 +36,5 @@ jobs:
           name: dapp
           signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
       - name: run integration tests
-        run: nix-shell --pure src/dapp-tests/shell.nix --command 'make --directory src/dapp-tests'
+        run: nix-env -iA nixpkgs.bashInteractive && nix-shell --pure src/dapp-tests/shell.nix --command 'make --directory src/dapp-tests'
       - run: nix-build release.nix -A dapphub.${{ matrix.os_attr }}.stable
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,3 +36,5 @@ jobs:
           name: dapp
           signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
       - run: nix-build release.nix -A dapphub.${{ matrix.os_attr }}.stable
+      - name: run integration tests
+        run: nix-shell -f src/dapp-tests/shell.nix
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,10 +30,5 @@ jobs:
         with:
           # we use the nixpkgs defined in default.nix
           skip_adding_nixpkgs_channel: true
-      # v6
-      - uses: cachix/cachix-action@490a246fbc7f92208d309eeb54383a4d828cedc1
-        with:
-          name: dapp
-          signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
       - name: run integration tests
         run: nix-shell --pure src/dapp-tests/shell.nix --command 'make --directory src/dapp-tests'
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,6 +35,5 @@ jobs:
         with:
           name: dapp
           signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
-      - run: nix-build release.nix -A dapphub.${{ matrix.os_attr }}.stable
       - name: run integration tests
         run: nix-shell src/dapp-tests/shell.nix --command 'make --directory src/dapp-tests'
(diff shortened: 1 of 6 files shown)


MrChico:
@asymmetric any idea why `dapp-testnet` seems unable to be used here? I expected the `$HOME` violations to be solved by the use of `--dir`

asymmetric:
@MrChico what is the problem?

MrChico:
@asymmetric in ci: `error: creating directory '/nix/var': Permission denied` when starting `dapp testnet` from the `integration/tests.sh`. Locally, it gives me: 
```sh
error: creating directory '/homeless-shelter': Permission denied
```
while building `dapp-test`. It could be some shenanegans that geth does while starting though? Idk, I don't see anything that would seem to warrant making a new directory in HOME while using `dapp testnet` with the `--dir` flag

asymmetric:
@MrChico this is what I get:

```
❯ src/dapp-tests/integration/tests.sh
dapp-testnet:   RPC URL: http://127.0.0.1:8545
dapp-testnet:  TCP port: 38545
dapp-testnet:  Chain ID: 99
dapp-testnet:  Database: /tmp/tmp.uSCqJyqixU/8545
dapp-testnet:  Geth log: /tmp/tmp.uSCqJyqixU/8545/geth.log
dapp-testnet:   Account: 0xB5E479C9Ae0EEf45c8E63AA2Dba06E3E30A6c592 (default)
"stateful.sol" is not found.
```

asymmetric:
It seems like it's geth who's trying to write to `/nix/var` for some reason. Maybe it's dereferencing a symlink?

MrChico:
Right, the `stateful.sol` should probably be given a path. If `geth` is the culprit this would be quite annoying. 
I wonder if there is another, easier way for us to simulate some real chain interaction or how to proceed

asymmetric:
@MrChico could it be it's trying to find somewhere to log? Maybe you couold try playing around with some flags and trying to control the paths it writes to.

The problem to solve here is that geth is being run as part of the build process, so in a read-only chrooted filesystem (apart from the build directory, which is writeable).

MrChico:
hmm, should we run these tests in a nix-shell, as we discussed before instead? Then we wouldn't have this limitation, right?

asymmetric:
@MrChico I think yes, we could do that and it could even be better from a conceptual POV, since now we're pretending like we're building a package whereas we're actually running integration tests in a shell script.

So I :+1: and might try to do it if I have some time, but feel free to tackle it.

src/dapp-tests/integration/tests.sh
```diff
@@ -2,15 +2,56 @@
...
+# tests some of the behaviour of
+# `dapp testnet`
+# `seth ls`
+# `seth send`
+# `seth run-tx`
+# `hevm exec`
+dapp_testnet() {
+  TMPDIR=$(mktemp -d)
+
+  dapp testnet --dir "$TMPDIR" &
```
asymmetric, line 23: Note that this causes lots of processes to be left around after the script exits, because of the use of `&`.

asymmetric:
Also we have false positives on macOS https://github.com/dapphub/dapptools/runs/875042657?check_suite_focus=true#step:6:425

src/dapp-tests/integration/tests.sh
MrChico: yeah, it needs to be killed later. I used to have it in the script but looks like it got lost

MrChico:
macos build still failing, this seems to be the problem https://github.com/NixOS/nixpkgs/issues/71625

src/dapp-tests/integration/tests.sh
```diff
@@ -2,15 +2,56 @@
 
 set -e
 
-export SOLC_FLAGS="--evm-version constantinople"
-export DAPPSYS_PATH="lib"
+# clean up
+trap 'killall geth && rm -rf "$TMPDIR"' EXIT
```
asymmetric, line 6: It's sad that we will end up killing any instance of geth running on this host. Ideally we would be a bit more surgical about it.

MrChico: Ideally... But `dapp testnet` starts geth as another subprocess so we can't use `$!`. I'm not sure what another way to do it would be

asymmetric: Namespaces/cgroups could be a way, or otherwise process groups. I started looking into the latter but then it became a rabbit hole of itself, and the former seems a bit overkill, so I let it go.

MrChico:
This is a good start! If we make a habit out of adding testcases here when fixing features this should be quite useful

src/dapp-tests/integration/tests.sh
```diff
@@ -2,15 +2,56 @@
...
+# `seth ls`
+# `seth send`
+# `seth run-tx`
+# `hevm exec`
+dapp_testnet() {
+  TMPDIR=$(mktemp -d)
+
+  dapp testnet --dir "$TMPDIR" &
+  # give it a few secs to start up
+  sleep 20
```
asymmetric, line 26: We might not need this huge delay

.github/workflows/build.yml
```diff
@@ -29,10 +29,12 @@ jobs:
...
           # we use the nixpkgs defined in default.nix
-          skip_adding_nixpkgs_channel: true
+          skip_adding_nixpkgs_channel: false
       # v6
       - uses: cachix/cachix-action@490a246fbc7f92208d309eeb54383a4d828cedc1
         with:
           name: dapp
           signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
+      - name: run integration tests
+        run: nix-env -iA nixpkgs.bashInteractive && nix-shell --pure src/dapp-tests/shell.nix --command 'make --directory src/dapp-tests'
```
asymmetric, line 39: Am afk but why is the nix-env necessary? Is it to work around the macOS issue?

MrChico: Yeah, it resolves the "bad substitutions" problem

src/dapp-tests/integration/tests.sh
```suggestion
  sleep 10
```

src/dapp-tests/integration/tests.sh
```diff
@@ -2,15 +2,56 @@
...
+# `seth ls`
+# `seth send`
+# `seth run-tx`
+# `hevm exec`
+dapp_testnet() {
+  TMPDIR=$(mktemp -d)
+
+  dapp testnet --dir "$TMPDIR" &
+  # give it a few secs to start up
+  sleep 10
```
```suggestion
  sleep 20
```

PR #421, opened 2020-07-08
dapp: 0.28.0

asymmetric:
Updates default solc to 0.6.7.

Actually, I want this to see if stuff like [this](https://github.com/dapphub/dapp-tools/runs/849891626?check_suite_focus=true) is a fluke, due to missing caches.


asymmetric:
@MrChico you wanna change the hevm CHANGELOG here?

asymmetric:
We agreed out-of-band that the hevm doesn't need updating, since the change only affects the internals.

Issue #420, opened 2020-07-08
Failing tests cause Decoding error in Stepper.hs


MrChico:
Introduced pretty recently on master, discovered by @livnev. Not quite sure why, possibly related to the recent changes simplifying execmode?
To reproduce:
```
contract ATest is DSTest {
  function setUp() public {
  }
  function test_basic() public {
    require(false);
  }
}
```

PR #424, opened 2020-07-09
Fix UnitTest vm errors

rainbreak:
@MrChico

### DIFF
--- a/src/hevm/src/EVM/UnitTest.hs
+++ b/src/hevm/src/EVM/UnitTest.hs
@@ -454,13 +454,12 @@ runUnitTestContract
                     (Right False, vm) ->
                              pure ("\x1b[31m[FAIL]\x1b[0m "
                              <> testName <> argInfo, Left (failOutput vm dapp opts testName))
-                    (Left _, _)       ->
+                    (Left e, vm)       ->
                              pure ("\x1b[33m[OOPS]\x1b[0m "
-                             <> testName <> argInfo, Left ("VM error for " <> testName))
-
-                (Left _, _)       ->
+                             <> testName <> argInfo, Left (bailOutput e vm dapp opts testName))
+                (Left e, vm)       ->
                   pure ("\x1b[33m[OOPS]\x1b[0m "
-                        <> testName <> argInfo, Left ("VM error for " <> testName))
+                        <> testName <> argInfo, Left (bailOutput e vm dapp opts testName))
 
           -- Define the thread spawner for property based tests
           let fuzzRun (testName, types) = do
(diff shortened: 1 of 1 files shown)


src/hevm/src/EVM/UnitTest.hs
```diff
@@ -167,20 +167,23 @@ execTest UnitTestOptions { .. } method args = do
...
-     Stepper.note "Checking whether assertions failed"
-     res <- Stepper.execFullyOrFail >>= Stepper.decode AbiBoolType
-     let AbiBool failed = res
-     -- Return true if the test was successful
-     pure (shouldFail == (bailed || failed))
+  if bailed then
+    pure (False)
+  else do
+    -- Decide whether the test is supposed to fail or succeed
+    let shouldFail = "testFail" `isPrefixOf` method
```
MrChico, line 174:
```suggestion
   -- Decide whether the test is supposed to fail or succeed
  let shouldFail = "testFail" `isPrefixOf` method
  if bailed then
    pure shouldFail
  else do

```

MrChico: do you agree?

src/hevm/src/EVM/UnitTest.hs
```diff
@@ -548,6 +550,18 @@ failOutput vm dapp UnitTestOptions { .. } testName = mconcat
       _       -> indentLines 2 (showTraceTree dapp vm)
   ]
 
+bailOutput :: Stepper.Failure -> VM -> DappInfo -> UnitTestOptions -> Text -> Text
```
MrChico, line 553: If you let `failOutput` take a `Maybe Stepper.Failure` we could merge `bailOutput` and `failOutput` and get rid of this copypasta

src/hevm/src/EVM/UnitTest.hs
rainbreak: I don't think so, we shouldn't be expecting vm failures with `testFailBlah`. We could perhaps invent a new category of `testBailBlah`.

src/hevm/src/EVM/UnitTest.hs
rainbreak: We could, but I don't think we're done iterating on this area yet so imo it's a bit more flexible to keep separate until we have some bigger vision for rewriting the whole thing.

src/hevm/src/EVM/UnitTest.hs
MrChico: but reverts are vm failures, as are INVALID opcode or anything. The following test should pass:
```
function testFail() {
  require(false);
}
```
This is the current semantics (or, well, before the bug recently introduced).

rainbreak: Good point. We have

```
ContractNotFound
DecodingError
VMFailed
```

so how about only applying this treatement to `VMFailed` and `DecodingError` and `ContractNotFound` stay as always-fail?

rainbreak: `ContractNotFound` can be removed actually, it isn't used and we already have `VMError NoSuchContract`. So all we're actually talking about is `DecodingError`.

rainbreak: Ok I can see that your suggestion will do that now anyway, as `DecodingError` is only ever raised from this particular block of code.

MrChico: Yeah. tbh, keeping the decoding inside the stepper script feels a little too much and I could see `Stepper.Failure` being removed completely in favor of just VMError. Then we'd just return false here if the decoding was unsuccessful.

MrChico: but that could also be postponed to keep this focused

MrChico:
fixes #420

PR #427, opened 2020-07-13
Hevm39

MrChico:
Mainly to have something to fall back to before the release that will come with #353

### DIFF
--- a/src/hevm/CHANGELOG.md
+++ b/src/hevm/CHANGELOG.md
@@ -1,6 +1,6 @@
 # hevm changelog
 
-## 0.39 - unreleased
+## 0.39 - 2020-07-13
  - Exposes abi encoding to cli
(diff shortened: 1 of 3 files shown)


asymmetric:
So this closes #421?

asymmetric:
Also, IIUC this means using < 0.6.x in tests won't work anymore?

In general, it seems there is something we should document about this magic version number in hevm and how/when it matters for users.

MrChico:
> So this closes #421?

It's based on top of it, so it can replace it or follow it

MrChico:
> Also, IIUC this means using < 0.6.x in tests won't work anymore?

It means using `< 0.6.x` one will need to provide `dapp --use solc:0.5.2` or whatever to use with older versions. Default will be `0.6.7`

asymmetric:
So e.g. `dapp --use solc.0.5.2 test`?

MrChico:
exactly

PR #432, opened 2020-07-14
dont interpret short bytestrings as decimals in seth --show-json

MrChico:
Currently, `seth --show-json` displays hex strings shorter than 20 chars as a decimal. This is quite unexpected behaviour on its own, as it sometimes causes `seth code` or `seth tx` to display short bytecode or input data in decimal format.
Things get worse since `seth debug` relies quite a bit on `seth --show-json`, misinterpreting the decimal data as hexidecimal. Cf. https://dapphub.chat/channel/dev?msg=LQAyMhNsJySC35sho


MrChico:
Tried to make sure everywhere `seth --show-json` is used (implicitly in `seth rpc`), decimal strings are printed to keep current behaviour. Still could use another look, or better yet, wait for #417 to get some tests before merging.

src/seth/libexec/seth/seth-run-tx
```diff
@@ -21,6 +21,7 @@ fi
 
 TO=$(<<< "$tx" seth --field to)
 DATA=$(<<< "$tx" seth --field input)
+
```
MrChico, line 24:
```suggestion
```

rainbreak:
seth call needs conversions too

rainbreak:
I think as a compromise we should have appropriate fields dec encoded when doing `seth block` and e.g. `seth block number latest`.

MrChico:
another way of doing this would be to just attack the places where this is a known problem, like `seth run-tx` and wrap the relevant parameters in `seth --to-hex`, which would be a less invasive change

MrChico:
this might still be a good idea, but the place where the current state of affairs were really causing problems was fixed by #516 , so I'll close this for now

Issue #426, opened 2020-07-13
Test setup fails when external libraries are used


realisation:
Whenever I use an external library, my tests fail at the setup phase.

```
[FAIL] setUp()

Failure: setUp
 
  src/test/deposits/suiteOne.t.sol:SelectiveDepositSuiteOne
   ├╴constructor
   └╴initialize test
      ├╴create <unknown contract> <no source map>
      │  └╴← 0 bytes of code
      ├╴create <unknown contract> <no source map>
          └╴← 0 bytes of code
```

Those calls are the creation of mock ERC20 contracts, interesting that they return 0 bytes.

More worrisome though is that I can't test my code in the state it will use for production. 

Is this a problem with linking libraries? Will it pose more problems when it is time to deploy?

MrChico:
Do you have a setUp function defined?

realisation:
Yeah, the problem above arises when I uncomment lines in the contract I am testing that call external functions from a library I made. If I recomment them, or make those functions internal, the problem goes away.

Those logs are the creation of the ERC20 mocks that occur by calling a function inside the setup function of `suiteOne.t.sol`

realisation:
I was able to reproduce the issue here - https://github.com/realisation/dapp-test-external-library-example

Its a bare bones contract with bare bones external lib showing how `dapp test` fails in the same manner reported above

MrChico:
Thanks for the example @realisation, I have a fix for you to try: #437

MrChico:
Fixed by #437

PR #437, opened 2020-07-16
dapp: fix dapp test for use with libraries

MrChico:
Addresses #426, @realisation please try it out

### DIFF
--- a/src/dapp/CHANGELOG.md
+++ b/src/dapp/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+Support for libraries. To test a project which includes libraries you need to use:
+```sh
+DAPP_BUILD_EXTRACT=1 dapp test
+```
+
 ## [0.28.0] - 2020-07-13
 ### Added
 - Support for solc 0.6.7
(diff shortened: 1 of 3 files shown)


MrChico:
For this to work you need to run dapp test with the environment var `DAPP_BUILD_EXRACT` set to 1

src/dapp/CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+Support for libraries. To test a project which includes libraries you need to use:
```
asymmetric, line 10:
```suggestion
- Support for libraries. To test a project which includes libraries you need to use:
```

Or something like that :)

src/dapp/libexec/dapp/dapp---make-library-state
```diff
@@ -18,20 +18,11 @@ if [ "$DAPP_VERBOSE" ]; then set -x; fi
 # history tracking.)
 
 sender=0x00a329c0648769a73afac7f9381e08fb43dbea72
-nonce=1048576
+nonce=48576
```
rainbreak, line 21: Do you know what caused this to change? The previous value was valid at some point.

MrChico: afaik, the value is not relevant, it just needs to match whats in dapp build. `dapp address` couldn't handle such a high nonce value though.

PR #438, opened 2020-07-16
remove dappsys

MrChico:
yagni anyone?


asymmetric:
What does dappsys even do

MrChico:
age

PR #439, opened 2020-07-17
seth usability tweaks

rainbreak:
Read stdin from conversion functions and add `--{max,min}-(u)int` helpers.

### DIFF
--- a/src/seth/CHANGELOG.md
+++ b/src/seth/CHANGELOG.md
@@ -16,6 +16,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `seth block` can now be passed the `--full` option, which returns the full block.
 
+### Changed
+- `seth --to-{hex,wei,word,address,dec,int256,ascii,fix,uint256}` and
+  `seth --from-{ascii,wei}` can now read values from stdin.
+
 
 ## [0.9.0] - 2020-05-25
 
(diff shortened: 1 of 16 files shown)


src/seth/CHANGELOG.md
```diff
@@ -13,9 +13,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `seth run-tx <tx-hash> [--debug, --source <file>, --state <repository>]`,
   set hevm options according to the options given and `hevm exec`
   - `seth debug <tx-hash> [--no-src]` downloads etherscan source, executes previous txs in block and enters into an hevm interactive session for the given tx. `--no-src` skips the first step.
+  - `seth --{max-uint,max-int,min-int}` print the largest numbers (in hex) of a given bitsize
 
 - `seth block` can now be passed the `--full` option, which returns the full block.
 
+### Changed
+- `seth --to-{hex,wei,word,address,dec,int256,ascii,fix,uint256}` and
+  `seth --from-{ascii,wei}` can now convert values from stdin.
```
asymmetric, line 22:
```suggestion
  `seth --from-{ascii,wei}` can now read values from stdin.
```

src/seth/libexec/seth/seth---max-int
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+### seth---max-int -- print the maximum int of the given bits
+### Usage: seth --max-int <bits>
```
asymmetric, line 3:
```suggestion
### Usage: seth --max-int [<bits>]
```

Since it's optional?

asymmetric:
What do you think of this approach instead:

```
seth --to-bytes32 <<< 16
```

This way, we don't have to explicitly handle stdin, since the shell provides us with facilities for that already, and we [avoid a call to cat](https://en.wikipedia.org/wiki/Cat_%28Unix%29#Useless_use_of_cat).

Nitpicking here, obviously.

rainbreak:
I'm not sure I understand what you mean. The intended uses here would be something like

    seth --to-wei 22 eth | seth --to-uint256
    
    bc <<< "some complicated calculation" | seth --to-wei

The aim is to reduce nestings of `$(...)` that you typically get in long seth pipelines.

asymmetric:
I mean `seth --to-bytes32 <<< $(seth --to-wei 22 eth)`

PR #441, opened 2020-07-18
actually, do debug things

MrChico:


MrChico:
I don't get why push fails and pull request succeeds

src/dapp-tests/integration/tests.sh
```diff
@@ -1,10 +1,10 @@
 #!/usr/bin/env bash
 
-set -e
+set -ex
```
asymmetric, line 3: Is this nice to have on CI outputs? Or locally?

MrChico: It's nice on ci outputs!

src/dapp-tests/integration/tests.sh
```diff
@@ -23,7 +23,7 @@ dapp_testnet() {
 
   dapp testnet --dir "$TMPDIR" &
   # give it a few secs to start up
-  sleep 20
+  sleep 30
```
```suggestion
  sleep 60
```

xwvvvvwx:
looks like this is just failing due to `solc-0.4.18` on macos, which should be fixed as of c351dde206dc8d6f3237118f251cffb18553f292

MrChico:
Still failing for some mysterious reason

PR #445, opened 2020-07-20
seth: set DAPPTOOLS in default.nix

xwvvvvwx:
Turns out `seth` also depends on `DAPPTOOLS`....


MrChico:
should we also remove [this line](https://github.com/dapphub/dapptools/blob/master/src/seth/libexec/seth/seth---nix-run#L18) then? At least `dapp --nix-run` doesn't have it

xwvvvvwx:
nice catch :pray:

MrChico:
ci fail unrelated

PR #446, opened 2020-07-20
seth call --debug

rainbreak:
Adds two new flags:

- `seth call --hevm`: execute a call with hevm
- `seth call --debug`: execute a call interactively with hevm.

No squash merge please.

### DIFF
--- a/src/seth/libexec/seth/seth
+++ b/src/seth/libexec/seth/seth
@@ -217,6 +217,6 @@ if [[ ${#keystores[@]} -gt 0 ]]; then
 fi
 
 [[ -n $ETH_KEYSTORE ]] && export ETH_KEYSTORE
-export ETH_RPC_URL=${ETH_RPC_URL:-$eth_rpc_url}
+export ETH_RPC_URL=${ETH_RPC_URL:-${eth_rpc_url:-${ETH_RPC_HOST:-http://127.0.0.1}:${ETH_RPC_PORT:-8545}}}
 
 "${0##*/}-${1-help}" "${@:2}"
--- a/src/seth/libexec/seth/seth
+++ b/src/seth/libexec/seth/seth
@@ -26,6 +26,9 @@
 ### Options for seth-call(1):
 ###
 ###    -B, --block=<number>       block number (default: `latest')
+###    --hevm                     execute with hevm
+###    --debug                    execute with hevm interactive mode
+###    --code                     overwrite the code of the called contract (hevm only)
 ###
 ### Options for seth-send(1):
 ###
@@ -43,6 +46,7 @@
 ###    --debug                    run in hevm interactive mode
 ###    --source=<file>            path to combined.json source
 ###    --state=<dir>              directory to store hevm state in
+###
 ###
 ### Other options:
 ###
@@ -74,6 +78,9 @@ full                 get block with full transaction information
 
   Options for seth-call(1):
 B,block=number       block number (default: \`latest')
+hevm                 execute with hevm
+debug                execute with hevm interactive mode
+code                 bytecode to replace the called contract with
 
   Options for seth-send(1):
 create               use the calldata to create a new object
@@ -84,7 +91,7 @@ resend               force current nonce to overwrite tx
   Options for seth-run-tx(1):
 state=directory      git repo where state is kept
 timestamp=number     block timestamp for the tx
-debug                run in hevms interactive debugger
+debug                run in hevm's interactive debugger
 source=file          combined-json source information file
 
   Options for seth-debug(1):
@@ -145,7 +152,9 @@ while [[ $1 ]]; do
        --state)      shift; export HEVM_STATE=$1;;
        --timestamp)  shift; export HEVM_TIMESTAMP=$1;;
        --debug)             export HEVM_DEBUG=yes;;
+       --hevm)              export HEVM_EXEC=yes;;
        --source)     shift; export DAPP_JSON=$1;;
+       --code)       shift; export HEVM_CODE=$1;;
 
     -D|--dir)        shift; export SETH_DIR=$1;;
 
--- a/src/seth/libexec/seth/seth
+++ b/src/seth/libexec/seth/seth
@@ -104,6 +104,7 @@ D,dir=directory      directory to store output
 async                do not wait for transaction receipt
 follow               go into a loop watching for new logs
 j,json               print output as JSON (if applicable)
+v,verbose            print additional debugging information
 "
 
 set -e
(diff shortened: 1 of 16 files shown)


src/seth/libexec/seth/seth
```diff
@@ -145,7 +150,9 @@ while [[ $1 ]]; do
        --state)      shift; export HEVM_STATE=$1;;
        --timestamp)  shift; export HEVM_TIMESTAMP=$1;;
        --debug)             export HEVM_DEBUG=yes;;
+       --hevm)              export HEVM_EXEC=yes;;
        --source)     shift; export DAPP_JSON=$1;;
+       --code)       shift; export HEVM_CODE=$1;;
```
MrChico, line 155: this is also a flag to `seth call`? Will you add it to the options above as well? This would load the $TO contract and then overwrite its code, iiuc. This is pretty esoteric, but maybe cool to allow

rainbreak: Yes, will add docs.

rainbreak: One use would be for becoming the root user of some system, say you overwrote the governance contract. Another would be testing out some different mechanism somewhere that is heavily integrated with internal/external contracts that you don't want to replicate.

src/seth/libexec/seth/seth
```diff
@@ -74,6 +78,9 @@ full                 get block with full transaction information
 
   Options for seth-call(1):
 B,block=number       block number (default: \`latest')
+hevm                 execute with hevm
+debug                execute with hevm interactive mode
+code                 bytecode to replace the called contract with
```
MrChico, line 83: state as well, right?

src/seth/libexec/seth/seth-call
```diff
@@ -1,29 +1,100 @@
...
+  [[ $ETH_BLOCK = [0-9]* ]] && ETH_BLOCK=$(seth --to-hex "$ETH_BLOCK")
+  jshon+=(-s "${ETH_BLOCK-latest}" -i append)
+  result=$(seth rpc eth_call -- "${jshon[@]}")
+  seth --abi-decode "$2" "$result"
+else
+  opts=()
+  [[ "$HEVM_DEBUG" ]] && opts+=(--debug)
+  [[ "$HEVM_STATE" ]] && opts+=(--state "$HEVM_STATE")
+
+  opts+=(--rpc "$ETH_RPC_URL")
```
MrChico, line 58: in other places this defaults to a local node when not defined, see https://github.com/dapphub/dapptools/pull/446/files#diff-d3a1496555063b19b8bdb94c3476de0aR37

src/seth/libexec/seth/seth
rainbreak: Yes. It will already be captured because it is defined in the run-tx opts, I'm just hesitant to document it fully as an option currently.

src/seth/libexec/seth/seth-call
rainbreak: Will fix. We should really set all these defaults at the top level imo.

src/seth/libexec/seth/seth
MrChico: ok, I guess it can be our secret for now

rainbreak: It's partly because I think --state should only actually be writeable by `seth send`, where `seth call` would be read-only, but I think it needs a little more thought. We could have a persistent blockchain stored in the filesystem under git, which would be great for learning / experimentation compared to geth/parity db.

MrChico: hmmm, that's an interesting thought. So then --state would cause tx to not get sent at all?

rainbreak: I think so yes. --state is only meaningful when using hevm. `seth send --hevm --state dir` would write to the state dir, or discard the tx if the state dir is undefined.

Something like:

- `seth call`: read-only chain, no access to --state
- `seth send`: write chain, no access to --state
- `seth call --hevm`, read-only chain, read-only --state
- `seth send --hevm`, read-only chain, write --state

MrChico:
I did some testing of this and discovered that hevm does not support setting of `chainId`.
This causes discrepancy in running for `seth call --create Chai.sol --hevm` vs. `seth call --create Chai.sol` against geth rpc. Although it could be argued that this is outside of the scope of this PR

MrChico:
I guess we can make a separate issue of that and not further block this PR. This LGTM

rainbreak:
What discrepancy do you notice? hevm doesn't offer a way to set the chainId afaik.

MrChico:
I tried deploying the chai contract. This gives an assertion violation in hevm

rainbreak:
@MrChico does this work for you now?

MrChico:
If we want to make the chainid work properly we should probably add `seth chainid` and ensure that the relevant chainid is passed to `seth call --debug` and `seth debug`

MrChico:
or instead of `seth chainid` we can just use `seth rpc eth_chainId` directly

rainbreak:
It could also fallback to fetching from the rpc, which is implemented but not used currently. The priority would be `--chainid -> rpc chainId --> 1`.

src/seth/libexec/seth/seth-call
```diff
@@ -70,6 +70,7 @@ else
   opts+=(--block     "${ETH_BLOCK:-$(seth --field number <<< "$block")}")
   opts+=(--number    "${ETH_NUMBER:-$(seth --field number <<< "$block")}")
   opts+=(--timestamp "${HEVM_TIMESTAMP:-$(seth --field timestamp <<< "$block")}")
+  opts+=(--chainid   "${ETH_CHAINID:-$(seth rpc eth_chainId | seth --to-dec)}")
```
rainbreak, line 73: @MrChico are you satisfied?

MrChico: no, can you add it to seth-run-tx too plx

src/seth/libexec/seth/seth-run-tx
```diff
@@ -36,6 +36,7 @@ opts+=(--number "$(<<< "$tx" seth --field blockNumber)")
 opts+=(--block  "$(("$(<<< "$tx" seth --field blockNumber)" - 1))")
 opts+=(--rpc    "${ETH_RPC_URL}")
 opts+=(--timestamp "${HEVM_TIMESTAMP:-$(seth block "$(<<< "$tx" seth --field blockNumber)" timestamp)}")
+opts+=(--chainid   "${ETH_CHAINID:-$(seth rpc eth_chainId | seth --to-dec)}")
```
rainbreak, line 39: @MrChico

MrChico: ok I'm satisfied

rainbreak: Did you test with chai deploy?

Issue #447, opened 2020-07-21
dapp create --verify fails with seth --to-hexdata


realisation:
```
Verifying contract at 0x8dc2a50faea58cb3732eae0803dd08364d8b44d6
+ [[ -n F68P23IXYDN4KSPBSXSP97ZCAHVWC1K9KZ ]]
++ seth chain
+ chain=kovan
+ case "$chain" in
+ export ETHERSCAN_API_URL=https://api-kovan.etherscan.io/api
+ ETHERSCAN_API_URL=https://api-kovan.etherscan.io/api
+ export ETHERSCAN_URL=https://kovan.etherscan.io/address
+ ETHERSCAN_URL=https://kovan.etherscan.io/address
+ path=src/assimilators/kovan/susdReserves/kovanSUsdToSUsdAssimilator.sol:KovanSUsdToSUsdAssimilator
+ name=KovanSUsdToSUsdAssimilator
+ address=0x8dc2a50faea58cb3732eae0803dd08364d8b44d6
++ seth calldata ''
seth---to-hexdata: error: invalid hexdata: `'
+ constructorArguments=
0x8dC2a50faeA58cB3732EAE0803Dd08364d8B44d6
```

realisation:
#448

MrChico:
Closed by #448

PR #448, opened 2020-07-21
verify-contract working when constrcutor args are absent

realisation:
I am a noob at bash, so maybe there is a superior way to do this

the problem appeared to be that the argument being passed to `seth calldata` was in fact an empty string, or something along those lines, so checking for the existence of the variable `$4` and opting out of that branch if it is absent allows the script to run as planned

### DIFF
--- a/src/dapp/libexec/dapp/dapp-verify-contract
+++ b/src/dapp/libexec/dapp/dapp-verify-contract
@@ -31,9 +31,12 @@ esac
 path=${1?contractname}
 name=${path#*:}
 address=${2?contractaddress}
-constructorArguments=$(seth calldata "$4" "${@:5}")
-constructorArguments=${constructorArguments#0x}
-constructorArguments=${constructorArguments:8}
+
+if [[ $4 ]]; then
+    constructorArguments=$(seth calldata "$4" "${@:5}")
+    constructorArguments=${constructorArguments#0x}
+    constructorArguments=${constructorArguments:8}
+fi
 
 meta=$(<"$DAPP_JSON" jshon -e contracts -e "$path" -e metadata -u)
 version=$(jshon <<<"$meta" -e compiler -e version -u)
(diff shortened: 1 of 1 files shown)


realisation:
#447

realisation:
Also removed `set -ex` `sex +x` and `set -x` to avoid excessive printing to the console while verification runs

realisation:
maybe `set -e` should be in there, though

src/dapp/libexec/dapp/dapp-verify-contract
```diff
@@ -93,7 +95,8 @@ while [ $count -lt 5 ]; do
   response=$(curl -fsS "$ETHERSCAN_API_URL" -d "$query" \
   --data-urlencode "compilerversion=$version" \
   --data-urlencode "sourceCode@"<(echo "$source") \
-  --data-urlencode "constructorArguements=$constructorArguments" -X POST)
+  --data-urlencode "$constructorArguments && constructorArguements=$constructorArguments" \
```
MrChico, line 98: this is a noop, `&&` will always make sure the second argument executes. If you remove this change the script should work the same

MrChico:
I find this hard to review because the only way to test it is to constantly deploy new contracts and verify their code. But the change is ACK and with the changes I requested this LGTM if you had success using it @realisation

src/dapp/libexec/dapp/dapp-verify-contract
MrChico, line 98:
```suggestion
  --data-urlencode constructorArguements=$constructorArguments" \
```

MrChico:
> maybe `set -e` should be in there, though

Yes. This makes sure failures propagate properly. Please readd

src/dapp/libexec/dapp/dapp-verify-contract
asymmetric: Btw there is a typo in the original code - `Arguements`.

asymmetric: Ah I see you noticed this too, great.

MrChico: yes, it's etherscans api that has the typo lol

asymmetric:
```suggestion
  # NOTE: the Arguements typo is in etherscan's API
  --data-urlencode "constructorArguements=$constructorArguments" \
```

realisation: I included that because I figured that if $constructorArguments wasn't defined, then the left hand side of `&&` would be false. 

ironically it looks like this isn't needed, since after all this PR is about having no constructor args and this is working fine

MrChico:
As an aside, looking at this script I don't understand what $3 is supposed to be. Some usage tips in the beginning of the file would be nice.

MrChico:
Thanks for your contribution @realisation !

PR #451, opened 2020-07-21
overlay.nix: spring-cleaning

### LINKED ISSUE #407: Review attributes in overlay.nix
Springtime cleaning!

Are there any packages we don't need anymore? If so, removing them would reduce closure size and build times, and improve maintainability.

Some possible candidates:

- [ ] dappsys
- [ ] dappsys-merged
- [ ] dapp-tests (cfr. #404)
- [ ] known-contracts
- [ ] dapp-which
- [ ] dapp2
- [ ] evmdis
- [x] setzer/terra/chief (cfr. #377)
- [ ] myetherwallet
- [ ] dafny
- [ ] qrtx/qrtx-term

xwvvvvwx:
This cleans up a bunch of old attributes from `overlay.nix` that are (as far as I'm aware) currently unused, this removes everything from https://github.com/dapphub/dapptools/issues/407 and a few more.

The removed attributes are:

- `profilingHaskellPackages`
- `callSolidityPackage`
- `dapps`
- `evmdis`
- `known-contracts`
- `dapp-which`
- `celf`
- `myEtherWallet`
- `dafny`

If you use something on this list, I'm super happy to keep it, I just prefered to be aggresive and then readd anything that was actually useful if needed.

Closes #407.


overlay.nix
```diff
@@ -130,62 +105,7 @@ in rec {
     '';
   });
 
-  qrtx = self.bashScript {
```
asymmetric, line 133: Is this used by anybody? Could be cool in theory

MrChico: what is it?

xwvvvvwx: I think it prints a tx as a qr code? I guess it's useful if you're doing offline signing.

asymmetric:
[This](

https://github.com/dapphub/dapptools/tree/c351dde206dc8d6f3237118f251cffb18553f292/nix/build-dapp.nix#L43) should've probably been removed in #438 - can you remove it here instead?

overlay.nix
rainbreak: Yes, it's for using air-gapped signing devices. I vote leave these in.

nix/evmdis.nix
```diff
@@ -1,21 +0,0 @@
-{ stdenv, lib, buildGoPackage, fetchFromGitHub }:
```
rainbreak, line 1: I haven't used evmdis for a while, but we do advertise it on dapp.tools

overlay.nix
```diff
@@ -108,8 +85,6 @@ in rec {
 
   evmdis = self.callPackage ./nix/evmdis.nix {};
 
-  token = self.callPackage (import ./src/token) {};
```
rainbreak, line 111: I think this may have some users still, not sure.

rainbreak: Could just delete and see if there are complaints.

MrChico:
`dafny`, `celf` and `evmdis` all have equal version or newer in nixpkgs. So I'm fine with removing them. Does Mr. @livnev use `token` maybe? There's nothing I would really miss here.

overlay.nix
livnev: We should keep this.

overlay.nix
livnev: This is cool, +1 for keep

nix/evmdis.nix
xwvvvvwx: The version w

xwvvvvwx: The version we have here is more than 3 years old, I had assumed that it would be broken on most new contracts?

xwvvvvwx:
I put `qrtx`, `qrtx-term`, and `token` back.

MrChico:
> [This](https://github.com/dapphub/dapptools/tree/c351dde206dc8d6f3237118f251cffb18553f292/nix/build-dapp.nix#L43) should've probably been removed in #438 - can you remove it here instead? Or too OT?

plzz

nix/evmdis.nix
xwvvvvwx: There is a maintained and up to date version in `nixpkgs` fwiw: https://github.com/NixOS/nixpkgs/blob/84cf00f98031e93f389f1eb93c4a7374a33cc0a9/pkgs/development/tools/analysis/evmdis/default.nix

xwvvvvwx:
I'm not sure what tag to give the release after this has merged?

asymmetric:
Do we have to make a release for this? It's mostly internal (and deprecation of stuff nobody used, hopefully), so maybe it should just be bundled with the next user-facing release.

xwvvvvwx:
I'm fine with that.

Issue #442, opened 2020-07-19
Dapp has trouble accepting --libraries flag for solc


realisation:
If I run 

```
SOLC_FLAGS='--libraries "Controller:0x388be118C18Df3180F4FE0f0A19A4a4180ff45bC SelectiveLiquidity:0x970CC9F79B28662c198167f589af7f854b3BD3F8 ProportionalLiquidity:0x01F6252E54a272145CB2d035bF6a13DBD269426f Shells:0x725C05594edc80E48E8bE2220C9ef02c4d28F617 Swaps:0x9665552d314AC6d8004d3b8f53f6612Af4eF5760"' dapp build --extract
```
Then it tells me `"SelectiveLiquidity:0x970CC9F79B28662c198167f589af7f854b3BD3F8" is not found.`

This particular libraries flag works when I perform compiling by running solc instead of dapp.

MrChico:
I believe you need to use the full qualified path of the contract, see https://solidity.readthedocs.io/en/v0.6.11/contracts.html?highlight=library#libraries

realisation:
Hmm, the docs from the using the compiler section don't mention a full qualified path, but they do mention something like `Controller.sol:Controller:address`. 

That didn't work for me, what did wind up working for me was that same libraries flag mentioned above, run from one directory above where the source files including the libraries were located.

I'll try the full qualified path and see if that works with dapp

realisation:
I tried it with both a relative path and the absolute path and the same error returned

I did notice it is always the second library in the string, though

MrChico:
I think this is a shell string quoting issue. Try this:
```
SOLC_FLAGS='--libraries "Controller:0x388be118C18Df3180F4FE0f0A19A4a4180ff45bC SelectiveLiquidity:0x970CC9F79B28662c198167f589af7f854b3BD3F8 ProportionalLiquidity:0x01F6252E54a272145CB2d035bF6a13DBD269426f Shells:0x725C05594edc80E48E8bE2220C9ef02c4d28F617 Swaps:0x9665552d314AC6d8004d3b8f53f6612Af4eF5760"'
dapp build --extract
```

realisation:
with that, there isn't an error, but nor is any linking performed

realisation:
there is something to what you said - 

```
[...]

PR #454, opened 2020-07-23
SOLC_FLAGS --library flag successfully interpreted into solc

realisation:
I couldn't figure out how to get `SOLC_FLAGS` into the correct format without quotes when it had the `--libraries "lib:address lib:address"` flag , which was breaking the `solc` command, so instead I turned the `solc` command into a string and eval'd it.

would love to know if there is a better way to do this, but this works.

caveat: it still does not work when doing `solc --use 0.5.15`, just when using the default solc in the dapp distribution (which I have also had trouble changing in `overlay.nix`)

#442

### DIFF
--- a/src/dapp/libexec/dapp/dapp-build
+++ b/src/dapp/libexec/dapp/dapp-build
@@ -36,7 +36,6 @@ info() {
   echo >&2 "$FLAGS" "${0##*/}: $ARGS"
 }
 
-IFS=" " read -r -a opts <<<"$SOLC_FLAGS"
 json_opts=--combined-json=abi,bin,bin-runtime,srcmap,srcmap-runtime,ast,metadata
 solc --help | grep -q -- --storage-layout && json_opts+=,storage-layout
 solc --help | grep -q -- --overwrite && opts+=(--overwrite)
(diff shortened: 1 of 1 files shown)


realisation:
What's holding this one up? Should I test various other options to see if I didn't break them? I've been using this ever since I made it to compile external libraries and deploy them.

xwvvvvwx:
@realisation is this still broken with `dapp --use solc:x.y.z`?

realisation:
Ah, I wasn’t aware there was a specific problem with that particularly when
using other solc versions.

If I recall correctly, at the time I was experiencing a general brokenness
in the new version of dapp when using —use solc:x.y.z but I never thought
that was relevant to this particular change to parsing solc flag params in
dapp build.

Maybe I was mistaken?

xwvvvvwx:
Ah sorry, then I misunderstood your comment above.

xwvvvvwx:
@realisation if you rebase this onto master then I think we should be able to merge this

rainbreak:
Would strongly prefer this was done without `eval`.

rainbreak:
Also, `--use` has been changed to not use nix, so perhaps this won't be an issue for you now? Try it out and see - it links to a binary that's in your path rather than some isolated nix environment.

MrChico:
libraries are now dealt with using the `DAPP_LIBRARIES` environment variable

PR #459, opened 2020-07-25
hevm: default smttimeout to 20s

xwvvvvwx:
unlimited by default is probabily a little user unfriendly

### DIFF
--- a/src/hevm/hevm-cli/hevm-cli.hs
+++ b/src/hevm/hevm-cli/hevm-cli.hs
@@ -111,7 +111,7 @@ data Command w
       , arg           :: w ::: [String]           <?> "Values to encode"
       , debug         :: w ::: Bool               <?> "Run interactively"
       , getModels     :: w ::: Bool               <?> "Print example testcase for each execution path"
-      , smttimeout    :: w ::: Maybe Integer      <?> "Timeout given to SMT solver in milliseconds"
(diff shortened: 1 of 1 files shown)


src/hevm/hevm-cli/hevm-cli.hs
```diff
@@ -373,7 +373,7 @@ equivalence cmd =
        Just sig' -> do method' <- functionAbi sig'
                        return $ Just (view methodSignature method', snd <$> view methodInputs method')
 
-     void . runSMTWithTimeOut (solver cmd) (smttimeout cmd) . query $
+     void . runSMTWithTimeOut (solver cmd) (fromMaybe 5000 (smttimeout cmd)) . query $
```
MrChico, line 376: if you set the default in `runSMTWithTimeOut` we don't have to set it in two places

xwvvvvwx: done.

xwvvvvwx:
wondering if we want to have some syntax for setting an unlimited timeout? maybe `--smttimeout=0`?

MrChico:
yeah, I was thinking about that too, but maybe one can just set a ridiculously high timeout in that case instead

MrChico:
I suppose you have to increase the timeout for the `factor` test for it to work

xwvvvvwx:
hmmm, maybe 5000 as a default is too short in that case?

MrChico:
Maybe 20s?

PR #460, opened 2020-07-26
Hevmnixinfra

MrChico:
just some house keeping,
- fixing a get-model printing case overlap
- touching up the haskell nix infra in similar vein to https://github.com/ethereum/act/pull/25
- removing arcane haskell js infrastructure which I don't think is being used


MrChico:
Now also bumps sbv to catch some low hanging fruit optimizations discussed [here](https://github.com/LeventErkok/sbv/issues/547), and compiles both sbv and hevm with `O2` optimizations enabled. These changes alone seem to increase performance with about 5% for concrete executions:
```
records:6
sum:36.51
avg:6.085
std:0.0965660395790871
med:6.095max:6.22
min:5.92
```
vs.
```
records:6
sum:34.23
avg:5.705
std:0.104363148029616
med:5.74max:5.79
min:5.48
```

Issue #457, opened 2020-07-25
hevm: unexpected runtime error during symbolic execution


xwvvvvwx:
`hevm` fails with `hevm: Maybe.fromJust: Nothing` when stepping over the `revert(0,x)` in the following yul program:

```
{
        calldatacopy(0,0,1024)

        for {} calldatasize() { mstore(1, 2) } {
                let x := 7
                mstore(4, 5)
                revert(0, x)
        }

}
```

Note that this error is not reproducible when executing concretely:

```
$hevm exec --gas 99999999999 --code 6104006000600037361560195760076005600452806000fd505b --debug --calldata $(seth --to-bytes32 $(seth --to-hex 9)
```

Or when symbolicaly executing into a simple `revert(0,7`:

```
hevm symbolic --code $(tmp=$(mktemp) && echo '{ revert(0,7) }' > $tmp && solc  --yul --yul-dialect evm $tmp |  awk 'f{print;f=0} /Binary representation:/{f=1}') --debug
```

*Steps To Reproduce*

1. `hevm symbolic --code 6104006000600037361560195760076005600452806000fd505b`

MrChico:
I think we are currently not accepting symbolic bytes to be returned in reverts. 
I took the easy route of forcing concreteness since I expected the revertdata to be concrete error strings anyway. It's possible that it could be easy to generalize though

xwvvvvwx:
hmmmmm, it's a little curious that we get a `Maybe.fromJust: Nothing` with symbolic args to `revert` since I see `hevm: Unexpected symbolic argument at opcode: OpRevert. Not supported (yet!)` when I run:

```
hevm equivalence --code-a 61040060006000376055565b6000600083601f8401121515601e578182fd5b8235905067ffffffffffffffff8111156035578182fd5b602083019150836020828501011115604d5760006000fd5b5b9250929050565b6061600151600051600b565b8082525050 --code-b 61040060006000376055565b6000600083601f8401121515601e578182fd5b8235905067ffffffffffffffff8111156035578182fd5b602083019150836020828501011115604d5760006000fd5b5b9250929050565b6061600151600051600b565b8082525050
```

MrChico:
[...]

PR #465, opened 2020-07-29
Max iterations in `--debug`

MrChico:
Makes `hevm symbolic --debug` respect `--max-iterations` and fixes two issues discussed in #457

### DIFF
--- a/src/hevm/hevm-cli/hevm-cli.hs
+++ b/src/hevm/hevm-cli/hevm-cli.hs
@@ -695,9 +695,9 @@ symvmFromCommand cmd = do
   (calldata', cdlen, pathCond) <- case (calldata cmd, sig cmd) of
     -- fully abstract calldata (up to 1024 bytes)
     (Nothing, Nothing) -> do
-      cd <- sbytes1024
+      cd <- sbytes256
       len <- freshVar_
-      return (SymbolicBuffer cd, len, len .<= 1024)
+      return (SymbolicBuffer cd, len, len .<= 256)
     -- fully concrete calldata
     (Just c, Nothing) ->
       let cd = ConcreteBuffer $ decipher c
--- a/src/hevm/hevm-cli/hevm-cli.hs
+++ b/src/hevm/hevm-cli/hevm-cli.hs
@@ -695,7 +695,7 @@ symvmFromCommand cmd = do
   (calldata', cdlen, pathCond) <- case (calldata cmd, sig cmd) of
     -- fully abstract calldata (up to 1024 bytes)
     (Nothing, Nothing) -> do
-      cd <- sbytes256
+      cd <- sbytes1024
       len <- freshVar_
       return (SymbolicBuffer cd, len, len .<= 1024)
     -- fully concrete calldata
--- a/src/hevm/hevm-cli/hevm-cli.hs
+++ b/src/hevm/hevm-cli/hevm-cli.hs
@@ -440,7 +440,7 @@ assert cmd = do
     runSMTWithTimeOut (solver cmd) (smttimeout cmd) $ query $ do
       preState <- symvmFromCommand cmd
       smtState <- queryState
-      io $ void $ EVM.TTY.runFromVM srcinfo (EVM.Fetch.oracle smtState rpcinfo) preState
+      io $ void $ EVM.TTY.runFromVM (maxIterations cmd) srcinfo (EVM.Fetch.oracle smtState rpcinfo) preState
 
   else
     runSMTWithTimeOut (solver cmd) (smttimeout cmd) $ query $ do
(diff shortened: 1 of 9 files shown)


MrChico:
Also, 09b4f6c seems to increase performance substantially:
running: 
```time hevm symbolic --code 608060405234801561001057600080fd5b506004361061002b5760003560e01c80637bc5bbbf14610030575b600080fd5b6100e66004803603602081101561004657600080fd5b810190808035906020019064010000000081111561006357600080fd5b82018360208201111561007557600080fd5b8035906020019184602083028401116401000000008311171561009757600080fd5b919080806020026020016040519081016040528093929190818152602001838360200280828437600081840152601f19601f8201169050808301925050505050505091929192905050506100e8565b005b6002816000815181106100f757fe5b60200260200101818152505060018160008151811061011257fe5b60200260200101511461012157fe5b5056fea26469706673582212204a2766360edefd8eb1702916081173517092c6ba1780f966803c12606c548eb964736f6c63430006070033 --solver cvc4 --smttimeout 1
```
```
./6a3f29ff/bin/hevm symbolic --code  --solver cvc4 --smttimeout 1  9.19s user 1.20s system 126% cpu 8.184 total
./6a3f29ff/bin/hevm symbolic --code  --solver cvc4 --smttimeout 1  9.31s user 1.32s system 125% cpu 8.502 total
./6a3f29ff/bin/hevm symbolic --code  --solver cvc4 --smttimeout 1  10.07s user 1.45s system 124% cpu 9.289 total
./maxIterDebug/bin/hevm symbolic --code  --solver cvc4 --smttimeout 1  4.97s user 0.75s system 121% cpu 4.686 total
./maxIterDebug/bin/hevm symbolic --code  --solver cvc4 --smttimeout 1  4.82s user 0.69s system 122% cpu 4.502 total
./maxIterDebug/bin/hevm symbolic --code  --solver cvc4 --smttimeout 1  4.98s user 0.75s system 121% cpu 4.729 total
```

MrChico:
or take this:
master
```
hevm symbolic --code $(seth code 0xFa9FfdF4F8b2D74526021Aa088486E5fA6F81132)   493.28s user 13.22s system 99% cpu 8:27.07 total
```
this branch:
```
hevm symbolic --code $(seth code 0xFa9FfdF4F8b2D74526021Aa088486E5fA6F81132)   270.40s user 3.04s system 98% cpu 4:38.22 total
```

src/hevm/hevm-cli/hevm-cli.hs
```diff
@@ -693,7 +695,7 @@ symvmFromCommand cmd = do
   (calldata', cdlen, pathCond) <- case (calldata cmd, sig cmd) of
     -- fully abstract calldata (up to 1024 bytes)
     (Nothing, Nothing) -> do
-      cd <- sbytes256
```
xwvvvvwx, line 696: was this a correctness issue?

xwvvvvwx:
If I try stepping through [this contract](https://github.com/ethereum/solidity/blob/v0.6.7/test/libsolidity/smtCheckerTests/functions/functions_recursive.sol) with `--max-iterations 1` and `--debug` I never seem to break out of the `if` block, if I explore the contract without `--debug` and `--max-iterations 1`, it finishes almost instantly.

asciinema: https://asciinema.org/a/fyjlS1T9FHJvrE7LAGsOApM6a

src/hevm/hevm-cli/hevm-cli.hs
MrChico: Yes, I think so. Calldata would not be initialized more than 256 bytes, instead of 1024 as advertized

xwvvvvwx: can we add a test that would cover this?

MrChico:
Ya, you're right. This doesnt work as expected

xwvvvvwx:
Very nice performance boost, but I'm curious why adding more calldata makes things go fatster?

MrChico:
Actually, this calculation is off.SWord  1024 does not occupy 1024 bytes. This is actually less cd; 128 bytes

MrChico:
Okay, the performance boost was definitely due to less calldata being given, and not some other form of magic. Increasing it to 1024 does seem to slow down things considerably (I'm reading  `42.06s` for what currently takes 9), so I suggest we keep things as they are at 256 bytes, at least for the scope of this PR. We can generalize it later by using an actual smt array for calldata. 
It's not surprising that things don't scale very well, because the current way to fake dynamic calldata is [kind of a hack](https://github.com/dapphub/dapptools/blob/master/src/hevm/src/EVM/Symbolic.hs#L143).

Anyway, `--max-iterations` should now work as advertised in `--debug` mode. After spending a lot of time debugging the `interpret` and `takeStep` function of `TTY` both in this PR and in #301 I do believe that TTY and UnitTest are the most spaghetti parts of this code base and I want to take a stab at simplifying them. But I won't do that here.

MrChico:
closing in favor of #469

Issue #466, opened 2020-07-30
hevm: unreported assertion violation


xwvvvvwx:
When exploring the contract at https://github.com/ethereum/solidity/blob/v0.6.7/test/libsolidity/smtCheckerTests/types/mapping_aliasing_2.sol in debug mode I can reach an invalid opcode by entering the else clause in `g` and exploring to the assertion on line 15.

If I explore the contract without debug mode, then no assertion violations are reported.

A recording of this behaviour can be found here: https://asciinema.org/a/5FJJli984j0qiVhjCVXOWEUbs

I am exploring using the json file at http://ix.io/2sFz with the following bytecode:

```
608060405234801561001057600080fd5b506004361061002b5760003560e01c80630336578714610030575b600080fd5b6100726004803603606081101561004657600080fd5b810190808035151590602001909291908035906020019092919080359060200190929190505050610074565b005b821561009c576100976000600160008481526020019081526020016000206100d0565b6100cb565b6100ca60016000848152602001908152602001600020600160008481526020019081526020016000206100d0565b5b505050565b600282600080815260200190815260200160002081905550602a60008080815260200190815260200160002081905550602a60016000808152602001908152602001600020600080815260200190815260200160002081905550602a60026000808152602001908152602001600020600080815260200190815260200160002060006101000a81548160ff021916908360ff1602179055506001816000808152602001908152602001600020819055506002826000808152602001908152602001600020541461019c57fe5b602a60008080815260200190815260200160002054146101b857fe5b602a60016000808152602001908152602001600020600080815260200190815260200160002054146101e657fe5b602a60026000808152602001908152602001600020600080815260200190815260200160002060009054906101000a900460ff1660ff161461022457fe5b6001816000808152602001908152602001600020541461024057fe5b505056fea26469706673582212202d02b6ea4a12532af4c33b3386abd08c920086e6f760cb9e33ef4c5b5079fef264736f6c63430006070033
```

MrChico:
Oh, wow, soundness breakage. 
With `--get-models`, that path is reported as inconsistent...
```
-- Branch (4/6) --
Inconsistent path conditions: dead path
Just (VMFailure (UnrecognizedOpcode 254))
-- Branch (5/6) --
Inconsistent path conditions: dead path
Just (VMFailure (UnrecognizedOpcode 254))
```
but it should definitely be possible to reach an assert here. Will need to investigate closer. I suspect it has to do with the KECCAK uninterpreted function

[...]

Issue #456, opened 2020-07-25
non-derministic branching when backstepping during symbolic execution


xwvvvvwx:
I came across some bytecode that seems to have non deterministic branching behaviour. If I step forwards through the program I see only one branch point, but if I step forwards and then backwards and then forwards again across a `JUMPI` I am asked to choose a branch when stepping forwards across the `JUMPI` for the second time only.

I can only reproduce this behaviour after backstepping. It seems to be reproducible with `z3` and `cvc4`, although in the test program below I observed the non deterministic branching with `z3` at both PC13 and PC18, and with `cvc4` at PC13 only.

The test program was built from the following yul code:

```
{
    calldatacopy(0,0,1024)
    let y := mload(0x20)
    for {} and(y, 8) { pop(y) } {
        if y { continue }
        y := 0
    }
}
```

*Steps To Reproduce*

1. `hevm symbolic --code 61040060006000376020515b60088116156027578015601c576021565b600090505b8050600b565b50 --debug`
2. Step forwards until the first `JUMPI` at pc 13
3. Press `0` (do not jump)
4. Step forwards and backwards through the loop a few times

*Results*

After a while the user is non-deterministicaly asked to select a branch when stepping forwards over the `JUMPI` on PC13 or PC18.

MrChico:
Hmm, is this really non-deterministic though? I'm occassionally asked to choose a branch at 13 or 18 but it seems to happen after a couple of iterations, so the query would be different. As can be seen in [askSMT](https://github.com/dapphub/dapptools/blob/master/src/hevm/src/EVM.hs#L1474), we identify queries based on their codelocation and the number of times we have seen that codelocation before. I think this should be sufficient to distinguish them properly

xwvvvvwx:
hmmm I thought it was non deterministic because I could only repro the issue after backstepping? Could you get it to happen without stepping backwards?

MrChico:
Ah, no, I think you are right... Something weird is going on here

MrChico:
Precise instructions to reproduce:
```
hevm symbolic --code 61040060006000376020515b60088116156027578015601c576021565b600090505b8050600b565b50 --debug
```
1. `e`: Step to first branching point
2. `0`: Choose no jump.
3. `n*` until arriving at pc 11.
4. `p`: once, stepping back to pc 10.
5. `n`: until arriving at branch again at pc 13.

Notice that without backstepping, skipping step 4, the jumpi query at pc 13 will be resolved without a user input.

PR #467, opened 2020-07-31
Integrate solc test suite for hevm-symbolic

xwvvvvwx:
## Overview

This PR integrates two parts of the solc test suite and uses them to verify `hevm symbolic`:

1. the yul optimzer tests: a corpus of ~400 pairs of yul programs that have been written to be equivalant. Used to test `hevm equivalence`
2. the SMTCheker tests: a corpus of ~450 solidity contracts containing assertions, and annotations on whether or not those assertions should be violated. Used to test `hevm symbolic`

Not every test was useful in our context, but accounting for skipped tests, this gives us 474 new test cases.

The yul equivlance tests are run with both `cvc4` and `z3`, but the smt checker tests can only be run in  a reasonable time using `cvc4`. Interstingly `z3` is slightly faster on the equivalance tests (~30s for z3 and ~40s for cvc4 on my machine).

## Issues Found

During the course of integrating these tests I found 3 `hevm` bugs (including one soundness issue)

- https://github.com/dapphub/dapptools/issues/456
- https://github.com/dapphub/dapptools/issues/457
- https://github.com/dapphub/dapptools/issues/466

I also found four instances of incorrect tests in the yul optimizer test suite (https://github.com/ethereum/solidity/issues/9500).

The smtChecker tests are currently failing due to the soundness issue (https://github.com/dapphub/dapptools/issues/466).

## Results

*SMTChecker*:

```
ran: 451
passed: 288
skipped: 162
failed (smt reports assertion, hevm reports safe): 1
failed (hevm reports assertion, smt reports safe): 0
hevm timeout: 0 
hevm failure: 0
```

*yulEquivlence*

```
ran: 409
passed: 185
same bytecode: 89
ignored: 135
could not compile first program: 0
could not compile second program: 0
hevm timeout: 0
hevm execution failed: 0
```

cc @leonardoalt

### DIFF
--- a/nix/hevm-tests/smt-checker.nix
+++ b/nix/hevm-tests/smt-checker.nix
@@ -398,9 +398,9 @@ pkgs.runCommand "smtCheckerTests-${solver}" {} ''
      [ $smt_reports != 0 ]  || \
      [ $hevm_reports != 0 ]
   then
-    exit 0
-  else
     exit 1
+  else
+    exit 0
   fi
 ''
 
--- a/nix/hevm-tests/smt-checker.nix
+++ b/nix/hevm-tests/smt-checker.nix
@@ -5,6 +5,7 @@ let
   # --- binaries ---
 
   awk = "${pkgs.gawk}/bin/awk";
+  bc = "${pkgs.bc}/bin/bc";
   basename = "${pkgs.coreutils}/bin/basename";
   cat = "${pkgs.coreutils}/bin/cat";
   echo = "${pkgs.coreutils}/bin/echo";
@@ -84,6 +85,7 @@ let
     hevmAssertionViolation = "Assertion violation found";
     hevmErrorInBranch = "branch(es) errored while exploring";
     hevmCouldNotExplore = "FAIL: hevm was unable to explore the contract";
+    smtCheckerFailed = "SKIP: smtChecker could not be invoked on this file";
   };
 
   smtCheckerTests = "${solidity}/test/libsolidity/smtCheckerTests";
@@ -94,7 +96,7 @@ let
 
   divider = pkgs.writeShellScript "divider" ''
     ${echo}
-    ${echo} --------------------------------------------------------------------------------------------
+    ${echo} ============================================================================================
     ${echo}
   '';
 
@@ -114,12 +116,12 @@ let
 
     explore() {
       set -x
-      hevm_output=$(${timeout} 10s ${hevm} symbolic --code "$1" --solver "$2" --json-file "$3" $4 2>&1)
+      hevm_output=$(${timeout} 60s ${hevm} symbolic --code "$1" --solver "$2" --json-file "$3" $4 2>&1)
       status=$?
       set +x
 
       # handle timeouts
-      if [[ $status == 124 ]]; then ${echo} "${strings.timeout}"; fi
+      if [[ $status == 124 ]]; then ${echo} && ${echo} "${strings.timeout}"; fi
 
       # handle non assertion related hevm failures
       if [[ $status == 1 ]]; then
@@ -159,7 +161,12 @@ let
   # $1 == input file
   # $2 == hevm smt backend
   runSingleTest = pkgs.writeShellScript "runSingleTest" ''
-    #!/usr/bin/sh
+    print_test() {
+      ${echo}
+      ${echo} "*** Input File: $(${testName} $1) ***"
+      ${echo}
+      ${cat} "$1"
+    }
 
     testName=$(${testName} $1)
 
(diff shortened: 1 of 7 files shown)


nix/hevm-tests/smt-checker.nix
```diff
@@ -0,0 +1,383 @@
...
+    fi
+
+    ${grep} -q 'Error trying to invoke SMT solver.' $1
+    if [ $? == 0 ]; then ${echo} ${strings.smtCheckerFailed} && exit; fi
+    ${grep} -q 'Assertion checker does not yet implement' $1
+    if [ $? == 0 ]; then ${echo} ${strings.smtCheckerFailed} && exit; fi
+    ${grep} -q 'Assertion checker does not yet support' $1
+    if [ $? == 0 ]; then ${echo} ${strings.smtCheckerFailed} && exit; fi
+    ${grep} -q 'Internal error: Expression undefined for SMT solver.' $1
+    if [ $? == 0 ]; then ${echo} ${strings.smtCheckerFailed} && exit; fi
```
leonardoalt, line 298: Do you need to exit if that's the case (for the block) above? It could still report some assertion violation

overlay.nix
```diff
@@ -21,9 +21,12 @@ in rec {
   buildDappPackage = import ./nix/build-dapp-package.nix { inherit (self) pkgs; };
 
   # Here we can make e.g. integration tests for Dappsys,
```
leonardoalt, line 23:
```suggestion
  # Here we can make e.g. integration tests for Dappsys.
```

nix/hevm-tests/smt-checker.nix
```diff
@@ -0,0 +1,383 @@
...
+    "functions/constructor_hierarchy_mixed_chain_with_params.sol"
+    "functions/constructor_hierarchy_mixed_chain_with_params_2.sol"
+    "functions/constructor_simple.sol"
+    "functions/constructor_state_value_inherited.sol"
+    "functions/constructor_state_value.sol"
+    "inheritance/constructor_hierarchy_mixed_chain_with_params.sol"
+    "inheritance/constructor_state_variable_init_chain_run_all.sol"
+    "inheritance/constructor_state_variable_init_chain_run_all_2.sol"
+
+    # --- unbounded looping ---
```
MrChico, line 56: We can add `--max-iterations n` and try some of these. It looks like `functions/functions_recursive.sol` and `functions/functions_recursive_indirect` should work, even for `n = 1` for example.

MrChico:
This is super nice!
I have some reflections;
- First of all, I would expect ci to fail since the build is failing, but this doesn't seem to be the case. Do you know why?
- The summary seems particularly valuable, and it would be nice to have it easily accessible after a ci run. Maybe it would make sense to have these tests in their own ci workflow. 
Then we also wouldn't push them to the cache, because when we do have a cache hit, the test results are buried somewhere in a previous run. Or perhaps this is a good thing, because we don't have to wait for things that we know don't change. Regardless, it should be easy to find the summary.
- It could be nice to add some basic performance for the summary as well. Maybe we can just prefix all hevm invocations with `time` and add it up in some nice way in the end.

nix/hevm-tests/smt-checker.nix
```diff
@@ -0,0 +1,383 @@
...
+    "types/tuple_assignment_array.sol"
+    "types/array_dynamic_3.sol"
+    "types/array_static_1.sol"
+
+    # bounds checking on enum args during abi decoding
+    "typecast/enum_to_uint_max_value.sol"
+    "types/enum_explicit_values.sol"
+    "types/enum_transitivity.sol"
+
+    # --- smt checker false positives ---
```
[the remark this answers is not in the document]
xwvvvvwx: What makes you say that?

nix/hevm-tests/smt-checker.nix
xwvvvvwx: oh, I didn't realise that was the case. Removing this didn't change the numbers in the summary as the only new test executed was an SMTChecker false positive.

xwvvvvwx:
- fixed ci status (tests were not actuall being run in ci..)
- added `--max-iterations`
- added performance numbers to the summary

@MrChico regarding summary output / CI stages I'm not sure what the best approach is. Do you think it would be enough to run the hevm tests in the `run integration tests` section of the pipeline? Personally I think that not rerunning the tests if nothing has changed is quite nice, but maybe we can add some script that always prints the summary in CI even if the tests have not been re-executed?

MrChico:
I think we should ignore the soundness bug before merging this, so as to not block this PR.
This pipeline might warrant its own ci step

MrChico:
```
/nix/store/v5abzbplf5wflc4c47i5mg0s4hdznfcx-forceSymbolicMemory: line 20: /nix/store/wjpqpikc6szzv6lvgs984ilnwwv9vq4d-utillinux-1003.1-2008/bin/rev: No such file or directory
```
also these guys are in the logs

xwvvvvwx:
- ignored test failing due to soundness issue
- moved ci tests to their own stage
- used `rev` from `busybox` instead of `util-linux` (should be cross platform :crossed_fingers:)

nix/hevm-tests/smt-checker.nix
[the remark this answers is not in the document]
leonardoalt: the smtcheckertests has a bunch of other "assertion violation happens here" because of aliasing etc

leonardoalt:
Note for potentially another PR: all the tests in `test/libsolidity/semanticTests/*` that have the strings `// compileViaYul: also` or `// compileViaYul: true` could be compiled to Yul with (`solc --ir-optimized a.sol`) and without (`solc --ir`) the optimizer and equivalence checked.

nix/hevm-tests/smt-checker.nix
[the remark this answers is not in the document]
xwvvvvwx: Are those false positives? If I understand correctly, unless the function is marked as `external` aliasing issues could manifest from within the same contract, or from usage in an inherited contract, so I think I view those as valid warnings?

xwvvvvwx:
@leonardoalt thx for the nice suggestion, I think that's out of scope for this PR atm. I've added it to the list at https://github.com/dapphub/dapptools/issues/468 to keep track of it.

xwvvvvwx:
CI is passing on both platforms now, I think this is good to go :slightly_smiling_face:

leonardoalt:
Amazing!

PR #469, opened 2020-08-04
TTY refactor

MrChico:
Fairly radical refactor of the TTY module, based on top of #465 and #301.

Working on those PRs I was beginning to get quite tired with the way the code in these parts look, so I have simplified things in the following way:

Removed:
-  `Stepper.Fail`; things become a lot easier if monadic interpreters cannot fail. They should never fail because of user error anyway, only if there's a bug in hevm.
-  `Stepper.Note`. This seems to be completely unused at the moment.

Changed:
- Incorporated the resolve of blocking queries into `TTY.interpreter` directly instead of in `takeStep`, as alluded to in [this comment](https://github.com/dapphub/dapptools/blob/master/src/hevm/src/EVM/TTY.hs#L212).
- Unified the various forms of `stepMode` to either `Step Int` or `StepUntil Pred`.
- Unified `maxIterations` checking of `TTY` and `SymExec`.

I'm pretty sure I figured out the cause of the `"can't replace code of contract"` errors as well: it occurs when one tries to perform `exec1` (step one opcode) past the end of a vm execution. If the current action is deploying a contract, this tries to deposit its bytecode again.
The fix for this is to remember to check whether we have reached the end of the current Stepper.action (by checking if there is a result), and if so, carry on with the next step instead. The current code checks this only for the [`StepOne` case](https://github.com/dapphub/dapptools/blob/master/src/hevm/src/EVM/TTY.hs#L171), but in this PR this check happens whenever we are in the `Stepper.Exec` case.


src/hevm/src/EVM/SymExec.hs
```diff
@@ -249,36 +243,31 @@ equivalenceCheck bytecodeA bytecodeB maxiter signature' = do
...
-      checkSat >>= \case
-         Unk -> error "solver said unknown!"
-         Sat -> return $ Right preStateA
-         Unsat -> return $ Left (aVMs, bVMs)
+  let differingEndStates = uncurry distinct <$> [(a,b) | a <- aVMs, b <- bVMs]
+      distinct a b = let (aPath, bPath) = both' (view pathConditions) (a, b)
+                         (aSelf, bSelf) = both' (view (state . contract)) (a, b)
+                         (aEnv, bEnv) = both' (view (env . contracts)) (a, b)
+                         (aResult, bResult) = both' (view result) (a, b)
+                         (Symbolic aStorage, Symbolic bStorage) = (view storage (aEnv ^?! ix aSelf), view storage (bEnv ^?! ix bSelf))
```
rainbreak, line 253: pls line width :sweat_smile:

MrChico: at least this PR is an improvement :)

MrChico: but I'll fix

xwvvvvwx:
I can get stuck inside the `if` block of [this contract](https://github.com/ethereum/solidity/blob/v0.6.7/test/libsolidity/smtCheckerTests/functions/functions_recursive.sol) forever if I execute it with `--debug` and `--max-iterations 1` ([asciinema](https://asciinema.org/a/Uz4EaDKDA7WxQfkTFnWbbGMPw)). The same contract is explored almost instantly if I drop the `--debug` and keep the `--max-iterations 1`.

xwvvvvwx:
backstepping in long tests is _so fast_ :heart_eyes_cat: :sparkles: :+1:

src/hevm/src/EVM/UnitTest.hs
```diff
@@ -108,38 +109,31 @@ type ABIMethod = Text
...
   -- Constructor is loaded; run until it returns code
-  bytes <- Stepper.execFullyOrFail
-  case bytes of
-    SymbolicBuffer bs -> error "unexpected symbolic contract bytecode"
-    ConcreteBuffer bs -> do
-      addr <- Stepper.evm (use (state . contract))
+  void Stepper.execFully
 
-      -- Mutate the current contract to use the new code
-      Stepper.evm $ replaceCodeOfSelf (RuntimeCode bs)
```
xwvvvvwx, line 125: how come we can omit this now?

xwvvvvwx:
Progress indicators seem to be missing when running fuzz tests now.

asciinema: https://asciinema.org/a/1nyxXEVCCnOuCuLehWfhN6LNv

src/hevm/src/EVM/UnitTest.hs
```diff
@@ -177,26 +162,25 @@ checkFailures UnitTestOptions { .. } method args bailed = do
...
 fuzzTest :: UnitTestOptions -> Text -> [AbiType] -> VM -> Property
 fuzzTest opts sig types vm = forAllShow (genAbiValue (AbiTupleType $ Vector.fromList types)) (show . ByteStringS . encodeAbiValue)
   $ \args -> ioProperty $
-    (runStateT (interpret (oracle opts) (runUnitTest opts sig args)) vm) >>= \case
-      (Left _, _) -> return False
-      (Right b, _) -> return b
+    fst <$> (runStateT (interpret (oracle opts) (runUnitTest opts sig args)) vm)
+    -- >>= \case
+    --   (Left _, _) -> return False
+    --   (Right b, _) -> return b
```
xwvvvvwx, line 183: Did you intend to leave this here?

MrChico:
> Progress indicators seem to be missing when running fuzz tests now.
> 
> asciinema: https://asciinema.org/a/1nyxXEVCCnOuCuLehWfhN6LNv

Pass  `-v` for progress indicator

src/hevm/hevm-cli/hevm-cli.hs
```diff
@@ -222,8 +222,14 @@ instance Options.ParseRecord (Command Options.Wrapped) where
 optsMode :: Command Options.Unwrapped -> Mode
 optsMode x = if debug x then Debug else Run
 
-unitTestOptions :: Command Options.Unwrapped -> IO UnitTestOptions
-unitTestOptions cmd = do
+unitTestOptions :: Command Options.Unwrapped -> String -> IO UnitTestOptions
+unitTestOptions cmd testFile = do
+  let root = fromMaybe "." (dappRoot cmd)
+  srcInfo <- readSolc testFile >>= \case
+    Nothing -> error "Could not find read .sol.json file"
```
MrChico, line 229:
```suggestion
    Nothing -> error "Could not read .sol.json file"
```

src/hevm/src/EVM/TTY.hs
```diff
@@ -110,28 +110,31 @@ makeLenses ''UiTestPickerState
...
+     = Stopped a              -- ^ Program finished
+     | Continue (Stepper a)   -- ^ Took one step; more steps to go
+
 
 -- | This turns a @Stepper@ into a state action usable
 -- from within the TTY loop, yielding a @StepOutcome@ depending on the @StepMode@.
 interpret
-  :: (?fetcher :: Fetcher)
+  :: (?fetcher :: Fetcher
+  ,   ?maxIter :: Maybe Integer)
```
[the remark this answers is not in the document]
MrChico: [Implicit parameter](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#implicit-parameters)

src/hevm/src/EVM/TTY.hs
```diff
@@ -141,109 +144,67 @@ interpret mode =
...
       :: Operational.ProgramView Stepper.Action a
-      -> State UiVmState (StepOutcome a)
+      -> StateT UiVmState IO (Continuation a)
 
     eval (Operational.Return x) =
-      pure (Returned x)
+      pure (Stopped x)
 
     eval (action Operational.:>>= k) =
       case action of
```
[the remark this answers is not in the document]
MrChico: I am aware that I am omitting it. We are never calling this interpreter with a stepper that has `run` as an action. `run` and `exec` are pretty much the same only `run` gives you the whole vm, and `exec` just the result

src/hevm/src/EVM/UnitTest.hs
MrChico: The `initialUnitTestVm` is now tx for which `create` is true, so finalization of that tx deposits the bytecode

src/hevm/src/EVM/UnitTest.hs
MrChico: oops no

src/hevm/src/EVM/SymExec.hs
```diff
@@ -128,64 +136,54 @@ loadSymVM x initStore addr callvalue' calldata' =
...
-                             put vm
-                             b <- State.state (runState (continue 1)) >> interpret fetcher maxIter (k ())
-                             return $ liftA2 (<>) a b
+        Stepper.Ask (EVM.PleaseChoosePath continue) -> do
+          vm <- get
+          case maxIterationsReached vm maxIter of
+            Nothing -> do a <- interpret fetcher maxIter (Stepper.evm (continue True) >>= k)
+                          put vm
+                          b <- interpret fetcher maxIter (Stepper.evm (continue False) >>= k)
+                          return $ a <> b
```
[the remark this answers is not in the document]
MrChico: There might be a way, but I'm not sure if it's gonna be easy. I think it would have been a lot simpler if we wouldn't have an open socket with the solver

MrChico:
> I can get stuck inside the `if` block of [this contract](https://github.com/ethereum/solidity/blob/v0.6.7/test/libsolidity/smtCheckerTests/functions/functions_recursive.sol) forever if I execute it with `--debug` and `--max-iterations 1` ([asciinema](https://asciinema.org/a/Uz4EaDKDA7WxQfkTFnWbbGMPw)). The same contract is explored almost instantly if I drop the `--debug` and keep the `--max-iterations 1`.

oops, I forgot to negate the previous choice

MrChico:
> oops, I forgot to negate the previous choice

Fixed

src/hevm/hevm-cli/hevm-cli.hs
MrChico, line 229:
```suggestion
    Nothing -> error "Could not read .sol.json file"
```

PR #470, opened 2020-08-04
Support solc 0.6.12

desaperados:


desaperados:
I think this is failing because the test suite now requires evmone. Should we add this dependency at over at nixpgs or here on dapptools?

desaperados:
Actually - the nixpkgs solc version builds without issue. The build only fails here. And semantic tests are skipped if evmone is not present so that's not an issue.

xwvvvvwx:
smtChecker tests are failing, maybe due to a different z3 version, they always require the latest afaik

MrChico:
Still unclear why macos builds are failing. Could be a clang error (waiting for @leonardoalt to confirm), but I think this is fine to merge as is and if we figure out what causes the macos error we can update things

src/dapp/CHANGELOG.md
```diff
@@ -14,12 +14,13 @@ DAPP_BUILD_EXTRACT=1 dapp test
 
 ## [0.28.0] - 2020-07-13
 ### Added
+- Support for solc 0.6.12
```
asymmetric, line 17: Seems like this should go under `Unreleased` :)

PR #472, opened 2020-08-05
hevm: add knowledge to path conditions when query is Known

### LINKED ISSUE #456: non-derministic branching when backstepping during symbolic execution

MrChico:
Fixes #456. I previously failed to add the knowledge gained from queries returning `Known`, thinking that if this only happens when there is only one alternative, in which case we are not really adding any new information. But `Known` can also be the result of a cached choice, so the newly gained knowledge should be stored.

### DIFF
--- a/src/hevm/CHANGELOG.md
+++ b/src/hevm/CHANGELOG.md
@@ -8,6 +8,8 @@
 - z3 updated to 4.8.7
 - Generate more interesting values in property based testing, 
  and implement proper shrinking for all abi values.
+- Fixed soundness bug when using KECCAK or SHA256 opcode/precompile
+- Fixed an issue in debug mode where backstepping could cause path information to be forgotten
 
 ### Added
 
--- a/src/hevm/CHANGELOG.md
+++ b/src/hevm/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Fixed soundness bug when using KECCAK or SHA256 opcode/precompile
 - Fixed an issue in debug mode where backstepping could cause path information to be forgotten
 - Ensure that pathconditions are consistent when branching, and end the execution with VMFailure: DeadPath if this is not the case
+- Fixed a soundness bug where nonzero jumpconditions  were assumed to equal one.
 
 ### Added
 
--- a/src/hevm/CHANGELOG.md
+++ b/src/hevm/CHANGELOG.md
@@ -10,6 +10,7 @@
  and implement proper shrinking for all abi values.
 - Fixed soundness bug when using KECCAK or SHA256 opcode/precompile
 - Fixed an issue in debug mode where backstepping could cause path information to be forgotten
+- Ensure that pathconditions are consistent when branching, and end the execution with VMFailure: DeadPath if this is not the case
 
 ### Added
 
--- a/src/hevm/src/EVM/Fetch.hs
+++ b/src/hevm/src/EVM/Fetch.hs
@@ -167,16 +167,15 @@ oracle state info q = do
          case noJump of
             -- Unsat means condition
             -- cannot be nonzero
-            -- We assume it to be zero in this case.
-            -- Theoretically, it is possible that our pathconditions
-            -- are inconsistent, either due to:
-            -- a) a false assumption
-            -- b) a previous timeout
-            -- We allow ourselves to ignore these cases with the motivation that
-            -- a) should be checked elsewhere and,
-            -- if b) occurs, we are in a unreachable path, and its better
-            -- to not keep on branching anyway.
-            Unsat -> return $ continue (EVM.Known 0)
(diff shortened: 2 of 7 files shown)


MrChico:
hmm, but this actually screws things up

MrChico:
Okay, I found the cause of the soundness issue, it's a one character change (65138d1). This is included in this PR and I also removed the test case that found this from the list of ignores

MrChico:
Still not quite right...

MrChico:
Okay, all the failures are only timeouts now

MrChico:
ah, but now it's possible to end up in an inifite loop

MrChico:
Okay, I'm satisfied with this now. Please review if you can @xwvvvvwx. I had to ignore some new tests because things can become a little slower, because I'm taking some extra precautions now when branching:
At each JUMPI, after checking if `jumpcondition /= 0` is satisfiable, I now always also check its negation `jumpcondition == 0`, and end the path with failure if neither condition is satisfiable. This happens if we end up in a path with inconsistent path conditions, which can happen for example because of a previous timeout.

In summary, this PR changes the following things:
- When the jumpcondition is nonzero, assume that it is nonzero instead of assuming that it is one.
- Always add the knowledge gained from taking a path. Fixes #456
- End paths with failure if their pathconditions are inconsistent.
- Fix the embedding of bytestrings into integers, resolving the injectivity problem with keccak that causes #466

src/hevm/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 ### Changed
 
 - z3 updated to 4.8.7
+- Fixed soundness bug when using KECCAK or SHA256 opcode/precompile
+- Fixed an issue in debug mode where backstepping could cause path information to be forgotten
```
MrChico, line 10:
```suggestion
- Fixed an issue in debug mode where backstepping could cause path information to be forgotten
- Ensure that pathconditions are consistent when branching, and end the execution with VMFailure: DeadPath if this is not the case
```

src/hevm/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
...
+- Ensure that pathconditions are consistent when branching, and end the execution with VMFailure: DeadPath if this is not the case
```
MrChico, line 11:
```suggestion
- Ensure that pathconditions are consistent when branching, and end the execution with VMFailure: DeadPath if this is not the case
- Fixed a soundness bug where nonzero jumpconditions  were assumed to equal one.
```

src/hevm/src/EVM/Fetch.hs
```diff
@@ -167,23 +167,22 @@ oracle state info q = do
...
+                          -- Assume 0 is still possible.
+                          Unk -> return $ continue (EVM.Iszero True)
             -- Sat means its possible for condition
             -- to be nonzero.
             Sat -> do jump <- checksat $ pathconds .&& jumpcondition .== 0
                       -- can it also be zero?
                       case jump of
                         -- No. It must be nonzero
-                        Unsat -> return $ continue (EVM.Known 1)
+                        Unsat -> return $ continue (EVM.Iszero False)
```
xwvvvvwx, line 185: If I understood you correctly, this was another soundness bug @MrChico?

Could we add a test case that covers this behaviour?

MrChico: I discovered it through [this test](https://github.com/ethereum/solidity/blob/develop/test/libsolidity/smtCheckerTests/operators/compound_assignment_division_1.sol)

MrChico: Which was previously not a problem because we weren't storing the path condition information upon branching when the query was Known, but now we are adding this information anyway so that test should be sufficient 👍

Issue #473, opened 2020-08-06
choose more interesting values when fuzzing


MrChico:
The current generators for uint values of various sizes selects a value at random, and only very rarely generates extreme case values like 0, 1 or uint(-1). We should probably use some like integers instead, or hard code some interesting corner case values.

xwvvvvwx:
Would be cool to try running `hevm symbolic --get-models` and use that to generate starting values for fuzzing.

rainbreak:
For numeric inputs it would be interesting to look at discontinuities in behaviour. A simple case would be input ranges that revert / succeed, which could probably be done statically.

MrChico:
#474 is a simple way to make things more interesting, but there are still possible alternatives to consider which utilizes symbolic execution. But we can maybe make a more detailed issue about that once we have some better ideas

PR #474, opened 2020-08-09
better shrinking and generator choices for abivalues

MrChico:
addresses #473


MrChico:
Actually, this generates awfully low values for uints. I think the generator for uint should probably look the same as the one for ints

PR #475, opened 2020-08-10
Fuzz

MrChico:
After trying out the recently merged #473 , I noticed that it's preference for small values is a little too high.
This generator is instead a mix of small, midrange and high values by combining close to zero values, close to max values and uniform selection


MrChico:
wow these work great

PR #481, opened 2020-08-12
hevm: abi: fix large values in integral generator

xwvvvvwx:
- fix underflow for values close to the bound
- squash compiler warning about inferred type for `mn`


xwvvvvwx:
Hmmmm actually this seems to not be working, doesn't find counterexamples for `testSwapReferal` and `testDepositVirtualBalanceScaling` in the mooniswap test suite. More investigation required...

MrChico:
I'm gonna merge this until we can think of something better

MrChico:
I'm gonna merge this until we think of something better

PR #482, opened 2020-08-15
Globalstorage

MrChico:
Implements the ability to have symbolic storage for multiple contracts by creating fresharrays in Fetch.hs and using the new `sArrayList` technology from sbv

### DIFF
--- a/src/hevm/hevm-cli/hevm-cli.hs
+++ b/src/hevm/hevm-cli/hevm-cli.hs
@@ -732,8 +732,6 @@ symvmFromCommand cmd = do
 
     _ -> error "incompatible options: calldata and abi"
 
-  let model = fromMaybe SymbolicS (storageModel cmd)
-
   store <- case storageModel cmd of
     -- InitialS and SymbolicS can read and write to symbolic locations
     -- ConcreteS cannot (instead values can be fetched from rpc!)
@@ -800,7 +798,7 @@ symvmFromCommand cmd = do
       , EVM.vmoptSchedule      = FeeSchedule.istanbul
       , EVM.vmoptChainId       = word chainid 1
       , EVM.vmoptCreate        = create cmd
-      , EVM.vmoptStorageModel  = model
+      , EVM.vmoptStorageModel  = fromMaybe SymbolicS (storageModel cmd)
       }
     word f def = fromMaybe def (f cmd)
     addr f def = fromMaybe def (f cmd)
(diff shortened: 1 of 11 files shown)


src/hevm/hevm-cli/hevm-cli.hs
```diff
@@ -725,6 +732,8 @@ symvmFromCommand cmd = do
 
     _ -> error "incompatible options: calldata and abi"
 
+  let model = fromMaybe SymbolicS (storageModel cmd)
```
xwvvvvwx, line 735: Is this used anywhere?

src/hevm/hevm-cli/hevm-cli.hs
```diff
@@ -791,6 +800,7 @@ symvmFromCommand cmd = do
       , EVM.vmoptSchedule      = FeeSchedule.istanbul
       , EVM.vmoptChainId       = word chainid 1
       , EVM.vmoptCreate        = create cmd
+      , EVM.vmoptStorageModel  = ConcreteS
```
xwvvvvwx, line 803: Shoudl this be `model` or `fromMaybe SymbolicS (storageModel cmd)`?

src/hevm/src/EVM.hs
```diff
@@ -1911,8 +1938,12 @@ create self this xGas xValue xs newAddr initCode = do
         touchAccount self
         touchAccount newAddr
         let
+          store = case view (env . storageModel) vm0 of
+            ConcreteS -> Concrete mempty
+            SymbolicS -> Symbolic $ sListArray (Just 0) []
```
xwvvvvwx, line 1943:
```suggestion
            SymbolicS -> Symbolic $ sListArray Nothing []
```

src/hevm/hevm-cli/hevm-cli.hs
MrChico: yea, this should be `model`. Nice catch

MrChico, line 803:
```suggestion
      , EVM.vmoptStorageModel  = model
```

src/hevm/src/EVM.hs
MrChico: This is creating a new contract, so it's storage is assumed to be empty. We keep storage fully symbolic only if the account has already been initialized previous to this transaction; i.e. when it's loaded from rpc, is the initial contract, or loaded from state (currently not yet supported)

xwvvvvwx: ah had missed that this was in `create` :+1:

PR #483, opened 2020-08-17
Hackage release candidate

MrChico:
I'm now added as a maintainer for `hevm` on hackage, and would like to upload a more modern version. As advised by the hackage folks, I have added upper bounds to the projects dependencies. Hackage packages are expected to comply to with https://pvp.haskell.org/faq/#semver, which we haven't really been following so far. It could be a good idea to bump hevm to `0.1.0` (or similar) before we upload it to hackage. The last version uploaded was `0.24`, and if we don't bump the major version, the current version is assumed to be compatible with this one. What do folks think about this?


MrChico:
We could also start at `0.4.0` instead, maybe this would appear to be more similar to the `0.40` we have right now. I don't really know how one is supposed to start when adapting existing code into a new versioning scheme

rainbreak:
tbh I think semver is a massive pita unless you're big enough to have a Release Team and doesn't even make sense for <v1 projects. Perhaps hevm should be made v1.0.0 :shrug:

MrChico:
PVP is not semver though, and it doesn't have the same loophole for `<v1`. I don't really care for the numbers much, I just want to at least look like we are complying with the hackage way

rainbreak:
What's wrong with 0.41.0?

xwvvvvwx:
> The last version uploaded was 0.24, and if we don't bump the major version, the current version is assumed to be compatible with this one.

MrChico:
0.41.0 also works

rainbreak:
Ok so imo go with that, or 0.40.1, as it preserves the current version ordering while also satisfying hackage.

PR #489, opened 2020-08-24
upgrade stack.yaml for hevm-0.41

incertia:
i have gotten hevm-0.41 to build with the latest LTS stackage, at the expense of deleting all the requirements in `hevm.cabal`.  i'm unsure if we want to keep using the tight package bounds but i am definitely down to bump the minimum package versions to the latest versions that work via stack.


MrChico:
The package bounds in cabal, including upper bounds were recommended for the hackage upload, but if we can admit versions that work with stackage as well that would obviously be great

sshine:
I have also gotten Hevm to work with lts-16 by bumping individual package versions.

See the [`sshine/dapptools:lts-16.11`](https://github.com/sshine/dapptools/commit/71e568b7d4bfe7d64beaa15138bea190142c2a1a) branch.

Note that:
 - [`monoidal-containers-0.6.0.1`](https://hackage.haskell.org/package/monoidal-containers-0.6.0.1) still depends on `these-1.0.1` even though `lts-16` contains `these-1.1.1.1`.
   (I submitted a PR, but apparently two others had come before me, so that one is stalled.)
 - [`witherable-0.3.5`](https://www.stackage.org/lts-16.11/package/witherable-0.3.5) does not appear to be in Stackage yet.
 - [`sbv-8.7.5`](https://hackage.haskell.org/package/sbv) does not appear to be in Hackage yet.

sshine:
> if we can admit versions that work with stackage as well that would obviously be great

Once #492 is merged, the next step in that direction would be:

https://github.com/commercialhaskell/stackage/blob/master/MAINTAINERS.md#adding-a-package

There's a bit of extra work because of the C dependencies on libsecp256k1 and libff.

Even if not uploading on Stackage, having a stack.yaml around is greatly appreciated for nix retards like myself. :-)

PR #491, opened 2020-08-27
Dynamic buffers

MrChico:
Introduces support for dynamic data types in memory and calldata (and all over the place, really) through [SMT lists](https://rise4fun.com/z3/tutorialcontent/sequences). These are currently only supported by z3. 
Some clean up needs to happen before this is ready to merge, but I am at the point where I want to trigger some more tests.

There will also be a need for some more execution modes, such as `--no-gas-checks` or `--no-memory-checks` in order to avoid constantly queries z3 about stuff that isn't particularly interesting from a security perspective.

### DIFF
--- a/src/hevm/hevm-cli/hevm-cli.hs
+++ b/src/hevm/hevm-cli/hevm-cli.hs
@@ -518,7 +518,7 @@ assert cmd = do
                         "Stopped"
                       else io $ putStrLn $
                         "Returned: " <> show (ByteStringS msg)
-                    Just (EVM.VMSuccess (SymbolicBuffer msg)) -> do
+                    Just (EVM.VMSuccess (StaticSymBuffer msg)) -> do
                       out <- mapM (getValue.fromSized) msg
                       io . putStrLn $
                         "Returned: " <> show (ByteStringS (ByteString.pack out))
@@ -581,7 +581,7 @@ launchExec cmd = do
           exitWith (ExitFailure 2)
         Just (EVM.VMSuccess buf) -> do
           let msg = case buf of
-                SymbolicBuffer msg' -> forceLitBytes msg'
+                StaticSymBuffer msg' -> forceLitBytes msg'
                 ConcreteBuffer msg' -> msg'
           print $ ByteStringS msg
           case state cmd of
@@ -686,7 +686,7 @@ vmFromCommand cmd = do
 
         vm1 c = EVM.makeVm $ EVM.VMOpts
           { EVM.vmoptContract      = c
-          , EVM.vmoptCalldata      = (calldata', literal . num $ len calldata')
+          , EVM.vmoptCalldata      = calldata'
           , EVM.vmoptValue         = w256lit value'
           , EVM.vmoptAddress       = address'
           , EVM.vmoptCaller        = litAddr caller'
@@ -713,22 +713,18 @@ symvmFromCommand :: Command Options.Unwrapped -> Query EVM.VM
 symvmFromCommand cmd = do
   caller' <- maybe (SAddr <$> freshVar_) (return . litAddr) (caller cmd)
   callvalue' <- maybe (sw256 <$> freshVar_) (return . w256lit) (value cmd)
-  (calldata', cdlen, pathCond) <- case (calldata cmd, sig cmd) of
-    -- fully abstract calldata (up to 1024 bytes)
+  calldata' <- case (calldata cmd, sig cmd) of
+    -- static calldata (up to 256 bytes)
     (Nothing, Nothing) -> do
-      cd <- sbytes256
-      len <- freshVar_
-      return (SymbolicBuffer cd, len, len .<= 256)
+      StaticSymBuffer <$> sbytes256
     -- fully concrete calldata
     (Just c, Nothing) ->
-      let cd = ConcreteBuffer $ decipher c
-      in return (cd, num (len cd), sTrue)
+      return $ ConcreteBuffer $ decipher c
     -- calldata according to given abi with possible specializations from the `arg` list
     (Nothing, Just sig') -> do
       method' <- io $ functionAbi sig'
       let typs = snd <$> view methodInputs method'
-      (cd, cdlen) <- symCalldata (view methodSignature method') typs (arg cmd)
-      return (SymbolicBuffer cd, cdlen, sTrue)
+      StaticSymBuffer <$> staticCalldata (view methodSignature method') typs (arg cmd)
 
     _ -> error "incompatible options: calldata and abi"
 
@@ -749,7 +745,7 @@ symvmFromCommand cmd = do
           error $ "contract not found."
         Just contract' ->
           return $
-            vm1 cdlen calldata' callvalue' caller' (contract'' & set EVM.storage store)
+            vm1 calldata' callvalue' caller' (contract'' & set EVM.storage store)
           where
             contract'' = case code cmd of
               Nothing -> contract'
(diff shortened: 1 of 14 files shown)


src/hevm/hevm-cli/hevm-cli.hs
```diff
@@ -749,7 +745,10 @@ symvmFromCommand cmd = do
           error $ "contract not found."
         Just contract' ->
           return $
-            vm1 cdlen calldata' callvalue' caller' (contract'' & set EVM.storage store)
+            vm1 calldata' callvalue' caller'
+                (contract''
+                   & set EVM.storage store
+                   & set EVM.origStorage store)
```
xwvvvvwx, line 751: this was a bug in the previous implementation right?

src/hevm/src/EVM/ABI.hs
```diff
@@ -437,7 +437,7 @@ genAbiValue = \case
    AbiTupleType ts ->
      AbiTuple <$> mapM genAbiValue ts
   where
-    genUInt n = AbiUInt n <$> arbitraryIntegralWithMax n
+    genUInt n = AbiUInt n <$> arbitraryIntegralWithMax (2^n-1)
```
xwvvvvwx, line 440: this is the fix for the property based tests?

src/hevm/test/test.hs
```diff
@@ -155,17 +224,17 @@ main = defaultMain $ testGroup "hevm"
             }
           }
           |]
-        let asWord :: [SWord 8] -> SWord 256
+        let asWord :: HasCallStack => [SWord 8] -> SWord 256
             asWord = fromBytes
-            pre preVM = let SymbolicBuffer bs = ditch 4 (fst $ view (state . calldata) preVM)
-                            (x, y) = splitAt 32 bs
+            pre preVM = let StaticSymBuffer bs = ditch 4 $ view (state . calldata) preVM
+                            (x, y) = trace ("calldata length: " <> show (length bs)) $ splitAt 32 bs
```
xwvvvvwx, line 230: did you mean to leave this `trace` in?

src/hevm/src/EVM.hs
```diff
@@ -2532,16 +2542,19 @@ costOfPrecompile (FeeSchedule {..}) precompileAddr input =
     -- ECRECOVER
     0x1 -> 3000
     -- SHA2-256
-    0x2 -> num $ (((len input + 31) `div` 32) * 12) + 60
+    0x2 -> num $ (((l input + 31) `div` 32) * 12) + 60
+      where l i = fromMaybe (error "unsupported: dynamic data to SHA256") (unliteral $ len input)
     -- RIPEMD-160
-    0x3 -> num $ (((len input + 31) `div` 32) * 120) + 600
+    0x3 -> num $ (((l input + 31) `div` 32) * 120) + 600
+      where l i = fromMaybe (error "unsupported: dynamic data to SHA256") (unliteral $ len input)
```
xwvvvvwx, line 2549: error message seems wrong here?

src/hevm/src/EVM.hs
```diff
@@ -2532,16 +2542,19 @@ costOfPrecompile (FeeSchedule {..}) precompileAddr input =
...
     -- IDENTITY
-    0x4 -> num $ (((len input + 31) `div` 32) * 3) + 15
+    0x4 -> num $ (((l input + 31) `div` 32) * 3) + 15
+      where l i = fromMaybe (error "unsupported: dynamic data to SHA256") (unliteral $ len input)
```
xwvvvvwx, line 2552: error message seems wrong here?

src/hevm/src/EVM.hs
```diff
@@ -2561,11 +2574,13 @@ costOfPrecompile (FeeSchedule {..}) precompileAddr input =
     -- ECMUL
     0x7 -> g_ecmul
     -- ECPAIRING
-    0x8 -> num $ ((len input) `div` 192) * (num g_pairing_point) + (num g_pairing_base)
+    0x8 -> num $ ((l input) `div` 192) * (num g_pairing_point) + (num g_pairing_base)
+      where l i = fromMaybe (error "unsupported: dynamic data to SHA256") (unliteral $ len input)
```
xwvvvvwx, line 2578: error message seems wrong here?

src/hevm/src/EVM/SymExec.hs
```diff
@@ -45,33 +45,31 @@ sbytes1024 = liftA2 (++) sbytes512 sbytes512
...
-symAbiArg (AbiBytesType n) | n <= 32 = do x <- sbytes32
-                                          return (x, 32)
-                           | otherwise = error "bad type"
+symAbiArg AbiAddressType = sbytes32
+
+symAbiArg (AbiBytesType n)
+  | n <= 32 = sbytes32
+  | otherwise = error "bad type"
 
 -- TODO: is this encoding correct?
```
xwvvvvwx, line 65: Is this `TODO` still valid?

src/hevm/hevm-cli/hevm-cli.hs
```diff
@@ -713,22 +713,18 @@ symvmFromCommand :: Command Options.Unwrapped -> Query EVM.VM
 symvmFromCommand cmd = do
   caller' <- maybe (SAddr <$> freshVar_) (return . litAddr) (caller cmd)
   callvalue' <- maybe (sw256 <$> freshVar_) (return . w256lit) (value cmd)
-  (calldata', cdlen, pathCond) <- case (calldata cmd, sig cmd) of
-    -- fully abstract calldata (up to 1024 bytes)
+  calldata' <- case (calldata cmd, sig cmd) of
```
xwvvvvwx, line 716: If I'm reading this right, there is currently no command line flag that allows the user to execute with calldata set to a `DynamicSymBuffer`?

xwvvvvwx:
There are some test cases in the smt checker test suite that use dynamic data (under `OpCalldatacopy` in the `unsupported opcodes` section`). It would be nice to enable these.

Unfortunately the smt checker tests are executed with `cvc4` only, as `z3` has performance issues when symbolicly executing through the function dispatch code with static symbolic data, so there will probably need to be some rework of the test machinery to allow execution of at least some tests with `z3`.

src/hevm/hevm-cli/hevm-cli.hs
MrChico: Right now that's true. You can with `loadSymVM` though, which I've been experimenting with. Will include it in the cli once things mature

src/hevm/hevm-cli/hevm-cli.hs
MrChico: yeah, I think so

src/hevm/src/EVM/ABI.hs
MrChico: yeah, I think I need to rebase this

src/hevm/src/EVM/SymExec.hs
MrChico: It hasn't really seen any testing yet so I'm not sure

src/hevm/test/test.hs
MrChico: nope, it's removed later

src/hevm/src/EVM.hs
```diff
@@ -170,7 +170,7 @@ data Cache = Cache
 -- | A way to specify an initial VM state
 data VMOpts = VMOpts
   { vmoptContract :: Contract
-  , vmoptCalldata :: (Buffer, (SWord 32)) -- maximum size of uint32 as per eip 1985
+  , vmoptCalldata :: Buffer -- maximum size of uint32 as per eip 1985
```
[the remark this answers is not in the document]
MrChico: We assume the length of calldata is limited to a uint32. But this isn't really relied upon here since we don't carry the length around anymore, so I'll remove this comment

src/hevm/hevm-cli/hevm-cli.hs
xwvvvvwx: What tests would catch this?

MrChico: Maybe comparing gas usage of `seth --run-tx` and `seth send`

PR #492, opened 2020-08-27
Update resolver from lts-14.14 to lts-16.11

sshine:
Increase version bounds.

Remove extra-deps that were added to lts-16.

Add extra-deps that were not yet added to lts-16.

This fixes #489.


src/hevm/stack.yaml
```diff
@@ -1,29 +1,23 @@
...
 - text-format-0.3.2
-- text-zipper-0.10.1
 - Unixutils-1.54.1
-- vty-5.25.1
-- word-wrap-0.4.1
+- witherable-0.3.5
 - github: dmjio/semver-range
   commit: d8d9db892ddb6ae267c9bcbc4f6602668433f12a
+- github: leventerkok/sbv # 8.7.5
+  commit: a33239a8bc494282b10426ccf717a8ca635add60
```
MrChico, line 20: Lets use the same revision we are pinning to in our [nix infra](https://github.com/dapphub/dapptools/blob/master/haskell.nix#L18)
```suggestion
  commit: 91637c043d206530bc64d7eac88d2f80e8db0b85
```

MrChico:
Thank you! This looks good! Just a small nit

PR #496, opened 2020-08-30
Added instructions on how to run on infura kovan testnet

rootux:
# Implements
* Added instructions on how to run on `Infura kovan testnet`
* This make sure the user is not confused and also set export `SETH_CHAIN=kovan` which does not work otherwise


MrChico:
Thanks for the contribution, if you think this is clearer, then this looks good to me.

PR #498, opened 2020-09-02
Added rpc-addr parameter for geth

ramilexe:
This new parameter allows binding the geth to not only the 127.0.0.1 interface.
For example `dapp testnet --rpc-addr 0.0.0.0`


xwvvvvwx (commented):
Thanks! please add a note in the changelog, otherwise lgtm :pray:

MrChico (approved):

PR #502, opened 2020-09-08
add forceBuffer

incertia:
quick and easy


MrChico:
Seems reasonable. Thanks

PR #507, opened 2020-09-14
optimize SHR with respect to frombytes

MrChico:
Introduces a `FromBytes` whiff, enabling us to specialize `shr(calldataload(x), n)` to be known concretely if the first `256 - n / 8` bytes of calldata are known. While this may seem like a pretty specific optimization, this shows up in almost every contract as part of the function signature check.
Avoiding the query which checks the function signature is a valuable time saver on its own, but it will become even more important with the new symbolic execution strategy which is being developed at https://github.com/dapphub/dapptools/tree/altsymexec


src/hevm/src/EVM/Symbolic.hs
```diff
@@ -95,6 +92,13 @@ sgt :: SymWord -> SymWord -> SymWord
 sgt (S _ x) (S _ y) =
   sw256 $ ite (sFromIntegral x .> (sFromIntegral y :: (SInt 256))) 1 0
 
+shiftRight' :: SymWord -> SymWord -> SymWord
+shiftRight' a@(S _ a') b@(S _ b') = case (num <$> unliteral a', b) of
```
xwvvvvwx, line 96: `a` is unused here

src/hevm/src/EVM/Symbolic.hs
```diff
@@ -95,6 +92,13 @@ sgt :: SymWord -> SymWord -> SymWord
...
+  (Just n, (S (FromBytes (SymbolicBuffer a)) i)) | n `mod` 8 == 0 && n <= 256 ->
```
xwvvvvwx, line 97: `i` is unused here

src/hevm/src/EVM/Symbolic.hs
```diff
@@ -147,9 +151,14 @@ select' xs err ind = walk xs ind err
...
-readSWordWithBound ind (SymbolicBuffer xs) bound =
-  let boundedList = [ite (i .<= bound) x 0 | (x, i) <- zip xs [1..]]
-  in sw256 . fromBytes $ [select' boundedList 0 (ind + j) | j <- [0..31]]
+readSWordWithBound ind (SymbolicBuffer xs) bound = case (num <$> fromSized <$> unliteral ind, num <$> fromSized <$> unliteral bound) of
+  (Just i, Just b) ->
+    let bs = truncpad 32 $ drop i (take b xs)
+    in S (FromBytes (SymbolicBuffer bs)) (fromBytes bs)
+  _ -> 
+    let boundedList = [ite (i .<= bound) x 0 | (x, i) <- zip xs [1..]]
+    in sw256 . fromBytes $ [select' boundedList 0 (ind + j) | j <- [0..31]]
```
xwvvvvwx, line 160: can't really follow this, what is `select'` doing?

src/hevm/src/EVM/Symbolic.hs
MrChico, line 97:
```suggestion
shiftRight' :: SymWord -> SymWord -> SymWord
shiftRight' (S _ a') b@(S _ b') = case (num <$> unliteral a', b) of
  (Just n, (S (FromBytes (SymbolicBuffer a)) _)) | n `mod` 8 == 0 && n <= 256 ->
```

src/hevm/src/EVM/Symbolic.hs
MrChico: specialized to `select' :: [SWord 8] -> SWord 8 -> Int -> SWord 8` here, `select'` creates a massive folded `ite` over all of the elements of the list, which picks the element of the list at index `(ind + j)` if in range, otherwise 0.

MrChico: n.b. that this is written using list comprehension syntax https://wiki.haskell.org/List_comprehension

xwvvvvwx: ok got it now, it builds a new 32 element list, where each element is some giant hairy `ite` over the original list selecting the element at `ind + j`.

Why do we need the `select'` construct? Why not use `readArray boundedList (ind + j)`?

MrChico: Because static calladata are modeled as a (haskell) list, not an array

xwvvvvwx: ah yes, of course....

xwvvvvwx (approved):

Issue #511, opened 2020-09-18
Balance is increased when executing empty functions with positive gas price


ggrieco-tob:
In the latest revision (0afd97e739ec0e38e5a47c1efc24abcf2217660f), executing an empty function allows `msg.sender` to increase its balance when gas price is positive (e.g. 1). To reproduce, create a new dapp project and use `src/Test.t.sol` like this one:

```solidity
pragma solidity ^0.5.7;

import "ds-test/test.sol";

contract TestTest is DSTest {
    uint balance;
    address sender;

    constructor () payable public {
        sender = msg.sender;
        balance = msg.sender.balance;
    }

    function setUp() public { /* this function is expected to be empty */}

    function test_basic_balance() public {
        assertTrue(sender == msg.sender && balance == msg.sender.balance);
    }
}

```

Then you can run `dapp test` like this:

```
$ DAPP_TEST_GAS_PRICE=0 dapp test
...
Running 1 tests for src/Test.t.sol:TestTest
[PASS] test_basic_balance() (gas: 2601)

$ DAPP_TEST_GAS_PRICE=1 dapp test
...
Running 1 tests for src/Test.t.sol:TestTest
[FAIL] test_basic_balance()

Failure: test_basic_balance
  
  "Assertion failed"
```

MrChico:
Thanks for reporting @ggrieco-tob , I think the problem is that currently `msg.sender` is not charged for the initial gas cost, but still rewarded the remaining gas after successful calls. We should make sure to charge for the initial gas cost which should ensure that the msg.sender balance will not increase. But I don't think that the test case you have given will hold as currently stated even with this change. But the weaker `assertTrue(msg.sender.balance <= balance);` would.

ggrieco-tob:
> But I don't think that the test case you have given will hold as currently stated even with this change. But the weaker assertTrue(msg.sender.balance <= balance); would.

So, this means that the `msg.sender` should have enough balance to make a call, or the call will revert? (I don't think older hevm versions had this feature).

MrChico:
That's right. Older versions of hevm did not perform transaction finalization when doing unit tests, so they did not charge or refund gas. Now we do finalization, but missed charging the initial gas cost.
I think it's reasonable to charge the caller for gas, but we should set default values so it can always afford it

PR #512, opened 2020-09-22
Simplify VMTest for a unified transaction initialization procedure

MrChico:
Mainly a clean up of `VMTest.hs`, removing things related to the ethereum tests/[VMTest](https://github.com/ethereum/tests/tree/develop/VMTests) which we haven't run in quite a while, as well as simplifying things generally.

One of my concerns here is to try to use less special case logic when running the ethereum tests, so that we can increase confidence in the correctness of other endpoints.

I'm also adding the setupTx functionality to `UnitTest` so that gas is paid for appropriately, fixing #511. If the sending address cannot afford the initial gas deposit this yields a hard error.

This `setupTx` (and `initTx`) process which increments the origin nonce, pays for gas, and sets the relevant tx reversion state is not done for the `hevm exec` or `hevm symbolic` endpoints right now, but I would like to open up the discussion for whether it should. I think some default values would need to be adjusted if that would be the case, and we may need to start fetching additional information (origin balance), and decide what happens if the tx.origin cannot afford the gas deposit.

### DIFF
--- a/src/hevm/hevm-cli/hevm-cli.hs
+++ b/src/hevm/hevm-cli/hevm-cli.hs
@@ -726,9 +726,8 @@ vmFromCommand cmd = do
     (_, _, Nothing) ->
       error $ "must provide at least (rpc + address) or code"
 
-  return $ withCache (vm0 miner ts blockNum diff contract) &
-     EVM.env . EVM.contracts . ix address' . EVM.balance +~ (w256 value')
-      where
+  return $ VMTest.initTx $ withCache (vm0 miner ts blockNum diff contract)
+    where
         decipher = hexByteString "bytes" . strip0x
         block'   = maybe EVM.Fetch.Latest EVM.Fetch.BlockNumber (block cmd)
         value'   = word value 0
@@ -823,19 +822,19 @@ symvmFromCommand cmd = do
               -- if both code and url is given,
               -- fetch the contract and overwrite the code
               Just c -> EVM.initialContract (codeType $ decipher c)
-                        & set EVM.storage     store
                         & set EVM.origStorage (view EVM.origStorage contract')
                         & set EVM.balance     (view EVM.balance contract')
                         & set EVM.nonce       (view EVM.nonce contract')
                         & set EVM.external    (view EVM.external contract')
 
     (_, _, Just c)  ->
-      return $ (EVM.initialContract . codeType $ decipher c) & set EVM.storage store
+      return $ (EVM.initialContract . codeType $ decipher c)
     (_, _, Nothing) ->
       error $ "must provide at least (rpc + address) or code"
 
-  return $ withCache $ vm0 miner ts blockNum diff cdlen calldata' callvalue' caller' contract'
+  return $ (VMTest.initTx $ withCache $ vm0 miner ts blockNum diff cdlen calldata' callvalue' caller' contract')
     & over EVM.pathConditions (<> [pathCond])
+    & set (EVM.env . EVM.contracts . (ix address') . EVM.storage) store
 
   where
     decipher = hexByteString "bytes" . strip0x
--- a/src/hevm/hevm-cli/hevm-cli.hs
+++ b/src/hevm/hevm-cli/hevm-cli.hs
@@ -9,12 +9,13 @@
 {-# Language NumDecimals #-}
 {-# Language OverloadedStrings #-}
 {-# Language TypeOperators #-}
+{-# Language RecordWildCards #-}
 
 module Main where
 
 import EVM (StorageModel(..))
 import qualified EVM
-import EVM.Concrete (createAddress, w256)
+import EVM.Concrete (createAddress, w256, wordValue)
 import EVM.Symbolic (forceLitBytes, litAddr, w256lit, sw256, SymWord(..), len)
 import qualified EVM.FeeSchedule as FeeSchedule
 import qualified EVM.Fetch
@@ -129,7 +130,7 @@ data Command w
       , smttimeout    :: w ::: Maybe Integer <?> "Timeout given to SMT solver in milliseconds (default: 20000)"
       , maxIterations :: w ::: Maybe Integer <?> "Number of times we may revisit a particular branching point"
       , solver        :: w ::: Maybe Text    <?> "Used SMT solver: z3 (default) or cvc4"
-      , smtoutput     :: w ::: Bool               <?> "Print verbose smt output"
+      , smtoutput     :: w ::: Bool          <?> "Print verbose smt output"
       }
   | Exec -- Execute a given program with specified env & calldata
       { code        :: w ::: Maybe ByteString <?> "Program bytecode"
@@ -259,7 +260,7 @@ unitTestOptions cmd testFile = do
 
   vmModifier <- applyCache (state cmd, cache cmd)
 
-  params <- getParametersFromEnvironmentVariables
+  params <- getParametersFromEnvironmentVariables (rpc cmd)
 
   let
     testn = testNumber params
@@ -686,7 +687,18 @@ tohexOrText s = case "0x" `Char8.isPrefixOf` encodeUtf8 s of
 vmFromCommand :: Command Options.Unwrapped -> IO EVM.VM
 vmFromCommand cmd = do
   withCache <- applyCache (state cmd, cache cmd)
-  vm <- case (rpc cmd, address cmd, code cmd) of
+
+  (miner,ts,blockNum,diff) <- case rpc cmd of
+    Nothing -> return (0,0,0,0)
+    Just url -> EVM.Fetch.fetchBlockFrom block' url >>= \case
+      Nothing -> error $ "Could not fetch block"
+      Just EVM.Block{..} -> return (_coinbase
+                                   , wordValue _timestamp
+                                   , wordValue _number
+                                   , wordValue _difficulty
+                                   )
+
+  contract <- case (rpc cmd, address cmd, code cmd) of
     (Just url, Just addr', Just c) -> do
       EVM.Fetch.fetchContractFrom block' url addr' >>= \case
         Nothing ->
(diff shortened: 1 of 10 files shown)


src/hevm/src/EVM.hs
```diff
@@ -1246,6 +1245,12 @@ exec1 = do
         xxx ->
           vmError (UnrecognizedOpcode xxx)
 
+transfer :: Addr -> Addr -> Word -> EVM ()
+transfer xFrom xTo xValue =
+  zoom (env . contracts) $ do
```
xwvvvvwx, line 1250: what does `zoom` do?

src/hevm/src/EVM.hs
```diff
@@ -1963,15 +1966,13 @@ create self this xGas xValue xs newAddr initCode = do
 
         zoom (env . contracts) $ do
           oldAcc <- use (at newAddr)
```
xwvvvvwx, line 1968: what is this doing exactly?

src/hevm/src/EVM.hs
```diff
@@ -1963,15 +1966,13 @@ create self this xGas xValue xs newAddr initCode = do
...
-          let oldBal = case oldAcc of
-                Nothing -> 0
-                Just c  -> view balance c
-          assign (at newAddr) (Just newContract)
-          assign (ix newAddr . balance) (oldBal + xValue)
-          assign (ix newAddr . nonce) 1
-          modifying (ix self . balance) (flip (-) xValue)
+          let oldBal = maybe 0 (view balance) oldAcc
+
+          assign (at newAddr) (Just (newContract & balance .~ oldBal))
```
xwvvvvwx, line 1971: whose `balance` is being written to here?

src/hevm/src/EVM/UnitTest.hs
```diff
@@ -548,14 +549,19 @@ word32Bytes :: Word32 -> ByteString
...
-  assign (state . calldata) $ (ConcreteBuffer $ abiMethod sig args, literal . num . BS.length $ abiMethod sig args)
+  assign (state . calldata) (ConcreteBuffer $ abiMethod sig args, literal . num . BS.length $ abiMethod sig args)
   assign (state . caller) (litAddr testCaller)
   assign (state . gas) (w256 testGasCall)
+  origin' <- fromMaybe (initialContract (RuntimeCode mempty)) <$> use (env . contracts . at testOrigin)
+  let originBal = view balance origin'
+  when (originBal <= (w256 testGasprice) * (w256 testGasCall)) $ error "insufficient balance for gas cost"
+  modifying (env . contracts) (setupTx testOrigin testCoinbase (w256 testGasprice) (w256 testGasCall))
+  vm <- get
+  put $ initTx vm $ view (env . contracts) vm
```
xwvvvvwx, line 564: Why do we need `setupTx` and `initTx`, they seem like they are doing very similar things?

xwvvvvwx:
>> This setupTx (and initTx) process which increments the origin nonce, pays for gas, and sets the relevant tx reversion state is not done for the hevm exec or hevm symbolic endpoints right now, but I would like to open up the discussion for whether it should.

Why would we not do this?

src/hevm/src/EVM.hs
```diff
@@ -455,14 +455,15 @@ initialContract theContractCode = Contract
       keccak (stripBytecodeMetadata theCode)
   , _storage  = Concrete mempty
   , _balance  = 0
-  , _nonce    = 0
+  , _nonce    = if creation then 1 else 0
```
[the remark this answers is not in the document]
MrChico: KEVM also sets the nonce to 1 for newly created accounts as part of `#mkCreate` [here](https://github.com/kframework/evm-semantics/blob/master/evm.md#account-creationdeletion).
We are not always setting it to 1, however, because EOAs start with the nonce at 0.

src/hevm/src/EVM.hs
MrChico: it's a lens operator that allows you to zoom in on a substructure and modify it, without having to refer to its place in the larger structure all the time

MrChico: in other words it makes the `. env . contracts` implicits in all further lens operations

src/hevm/src/EVM.hs
MrChico: it's fetching the account that's currently deployed at the address and overwrites with a new account but keeps the balance.

src/hevm/src/EVM.hs
MrChico: `newContract`

src/hevm/src/EVM/UnitTest.hs
MrChico: hm, yes can maybe be unified

src/hevm/src/EVM.hs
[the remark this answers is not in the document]
xwvvvvwx: makes sense, thanks!

rainbreak:
This is great!

src/hevm/src/EVM.hs
```diff
@@ -1818,49 +1818,46 @@ cheat (inOffset, inSize) (outOffset, outSize) = do
...
-                             (LS.fromStrict input') of
-                        Right ("", _, args) ->
-                          action (toList args) >>= \case
-                            Nothing -> do
-                              next
-                              push 1
-                            Just (encodeAbiValue -> bs) -> do
-                              next
-                              modifying (state . memory)
-                                (writeMemory (ConcreteBuffer bs) outSize 0 outOffset)
```
rainbreak, line 1836: Do you think this was intended for return data or something else?

src/hevm/hevm-cli/hevm-cli.hs
```diff
@@ -686,36 +687,47 @@ tohexOrText s = case "0x" `Char8.isPrefixOf` encodeUtf8 s of
...
+
+  (miner,ts,blockNum,diff) <- case rpc cmd of
+    Nothing -> return (0,0,0,0)
+    Just url -> EVM.Fetch.fetchBlockFrom block' url >>= \case
+      Nothing -> error $ "Could not fetch block"
+      Just EVM.Block{..} -> return (_coinbase
+                                   , wordValue _timestamp
+                                   , wordValue _number
+                                   , wordValue _difficulty
+                                   )
```
rainbreak, line 699: cli arguments should override these, if they exist.

src/hevm/hevm-cli/hevm-cli.hs
```diff
@@ -686,36 +687,47 @@ tohexOrText s = case "0x" `Char8.isPrefixOf` encodeUtf8 s of
 vmFromCommand :: Command Options.Unwrapped -> IO EVM.VM
 vmFromCommand cmd = do
   withCache <- applyCache (state cmd, cache cmd)
-  vm <- case (rpc cmd, address cmd, code cmd) of
+
+  (miner,ts,blockNum,diff) <- case rpc cmd of
+    Nothing -> return (0,0,0,0)
```
rainbreak, line 692:
```suggestion
    Nothing -> return ( fromMaybe 0 (coinbase cmd)
                      , fromMaybe 0 (timestamp cmd)
                      , fromMaybe 0 (number cmd)
                      , fromMaybe 0 (difficulty cmd)
                      )
```

src/hevm/hevm-cli/hevm-cli.hs
rainbreak:
```suggestion
      Just EVM.Block{..} -> return ( fromMaybe _coinbase (coinbase cmd)
                                   , fromMaybe (wordValue _timestamp) (timestamp cmd)
                                   , fromMaybe (wordValue _number) (number cmd)
                                   , fromMaybe (wordValue _difficulty) (difficulty cmd)
                                   )
```

src/hevm/hevm-cli/hevm-cli.hs
```diff
@@ -755,6 +767,17 @@ vmFromCommand cmd = do
...
+    Just url -> io $ EVM.Fetch.fetchBlockFrom block' url >>= \case
+      Nothing -> error $ "Could not fetch block"
+      Just EVM.Block{..} -> return (_coinbase
+                                   , wordValue _timestamp
+                                   , wordValue _number
+                                   , wordValue _difficulty
+                                   )
```
rainbreak, line 779: similarly, cli args should override

rainbreak:
> > This setupTx (and initTx) process which increments the origin nonce, pays for gas, and sets the relevant tx reversion state is not done for the hevm exec or hevm symbolic endpoints right now, but I would like to open up the discussion for whether it should.
> 
> Why would we not do this?

Because `hevm exec` could be simulating a subcall rather than a tx call, in which case we don't want to bother with the changes. Perhaps a `--tx Bool` arg to exec would be useful.

src/hevm/hevm-cli/hevm-cli.hs
MrChico: These vars are only the default values for all these params. See `word`  function in vm0

src/hevm/src/EVM.hs
MrChico: Yes, I think so. n.b. that this now is now assumed to happen in the cheatAction

livnev:
> > > This setupTx (and initTx) process which increments the origin nonce, pays for gas, and sets the relevant tx reversion state is not done for the hevm exec or hevm symbolic endpoints right now, but I would like to open up the discussion for whether it should.
> > 
> > 
> > Why would we not do this?
> 
> @xwvvvvwx because `hevm exec` could be simulating a subcall rather than a tx call, in which case we don't want to bother with the changes. Perhaps a `--tx` arg to exec would be useful.

Hmm, don't we want `hevm exec` to be included in a wrapper that acts as a double to `seth call`? Or are you envisioning that this should be done with the `--tx` flag?

MrChico:
I think the default should be to perform tx initialization, so maybe a `--no-tx-init` flag or smth would be more applicable

src/hevm/hevm-cli/hevm-cli.hs
MrChico: they already do

MrChico:
> > > This setupTx (and initTx) process which increments the origin nonce, pays for gas, and sets the relevant tx reversion state is not done for the hevm exec or hevm symbolic endpoints right now, but I would like to open up the discussion for whether it should.
> > 
> > 
> > Why would we not do this?
> 
> @xwvvvvwx because `hevm exec` could be simulating a subcall rather than a tx call, in which case we don't want to bother with the changes. Perhaps a `--tx` arg to exec would be useful.

hmm, I'm not sure this makes sense though since we always perform tx finalization...

For those of you following along at home I think I will just add tx initialization here and then once we find a need for skipping it we can add it

MrChico:
Will merge this tomorrow unless anything else is found. For those of you following along at home, this PR implements the following potpourri:
- Refactoring of VMTest
- Perform tx initialization in `hevm dapp-test`, `hevm exec` and `hevm symbolic`
- Fetch `miner`, `timestamp`, `blocknumber` and `difficulty` when `--rpc` is given, and set default values accordingly.
- Share cache between dapp-test functions for faster `--rpc` interaction.
- `load(address,bytes32)` cheat code

PR #514, opened 2020-09-24
clearer unknown contract call formatting

rainbreak:
@MrChico

### DIFF
--- a/src/hevm/src/EVM/Format.hs
+++ b/src/hevm/src/EVM/Format.hs
@@ -213,10 +213,22 @@ showTrace dapp trace =
       t
     FrameTrace (CreationContext hash _ _ ) ->
       "create " <> maybeContractName (preview (dappSolcByHash . ix hash . _2) dapp) <> pos
-    FrameTrace (CallContext _ _ _ _ hash abi calldata _ _) ->
+    FrameTrace (CallContext target context _ _ hash abi calldata _ _) ->
       case preview (dappSolcByHash . ix hash . _2) dapp of
         Nothing ->
-          "call [unknown]" <> pos
+          if (target == context) then
+            "call "
+              <> pack (show target)
+              <> "::"
+              <> (formatSBinary calldata)
+              <> pos
+          else
+            "delegatecall "
+              <> pack (show target)
+              <> "::"
+              <> (formatSBinary calldata)
+              <> pos
+
         Just solc ->
           "call "
             <> "\x1b[1m"
--- a/src/hevm/src/EVM/Format.hs
+++ b/src/hevm/src/EVM/Format.hs
@@ -198,7 +198,7 @@ showTrace dapp trace =
         _ ->
           "\x1b[91merror\x1b[0m " <> pack (show e) <> pos
 
-    ReturnTrace out (CallContext _ _ hash (Just abi) _ _ _) ->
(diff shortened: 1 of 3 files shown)


src/hevm/src/EVM/Format.hs
```diff
@@ -138,7 +149,7 @@ formatSBinary (ConcreteBuffer bs) = formatBinary bs
 showTraceTree :: DappInfo -> VM -> Text
 showTraceTree dapp =
   traceForest
-    >>> fmap (fmap (unpack . showTrace dapp))
+    >>> fmap (fmap (unpack . (showTrace dapp)))
```
MrChico, line 152:
```suggestion
    >>> fmap (fmap (unpack . showTrace dapp))
```

src/hevm/src/EVM/Format.hs
```diff
@@ -213,20 +235,32 @@ showTrace dapp trace =
       t
     FrameTrace (CreationContext hash _ _ ) ->
       "create " <> maybeContractName (preview (dappSolcByHash . ix hash . _2) dapp) <> pos
-    FrameTrace (CallContext _ _ hash abi calldata _ _) ->
-      case preview (dappSolcByHash . ix hash . _2) dapp of
+    FrameTrace (CallContext target context _ _ hash abi calldata _ _) ->
+      let calltype = if (target == context)
+                     then "call "
+                     else "delegatecall "
```
MrChico, line 241: this could also be `CALLCODE` I believe

src/hevm/src/EVM/Format.hs
```diff
@@ -213,20 +235,32 @@ showTrace dapp trace =
       t
     FrameTrace (CreationContext hash _ _ ) ->
       "create " <> maybeContractName (preview (dappSolcByHash . ix hash . _2) dapp) <> pos
-    FrameTrace (CallContext _ _ hash abi calldata _ _) ->
-      case preview (dappSolcByHash . ix hash . _2) dapp of
+    FrameTrace (CallContext target context _ _ hash abi calldata _ _) ->
+      let calltype = if (target == context)
```
MrChico, line 239:
```suggestion
      let calltype = if target == context
```

src/hevm/src/EVM/Format.hs
rainbreak: Not as simple to determine unfortunately, so it will just show up as delegatecall.

MrChico: should we display it as `delegatecall/callcode`?

rainbreak: Imo no because it's clutter the vast majority of the time. Does anyone use callcode anymore?

rainbreak: If we really wanted to show it we could add a new `calltype` field to `CallContext`.

rainbreak: If you still think it's worth it, then how about also distinguishing `create` and `create2`?

MrChico: ok, I'm fine with `delegatecall` for now

MrChico (approved):

src/hevm/src/EVM/Format.hs
rainbreak: We also don't distinguish `staticcall`.

PR #519, opened 2020-09-26
hotfix for 3f31ca3

rainbreak:
Also don't think the `cd $DAPP_ROOT` are necessary, but will address those separately. @lucasvo @MrChico


MrChico:
Oops. Nice fix

PR #520, opened 2020-09-27
EVM.hs: add load cheatcode for reading arbitrary storage

livnev:


src/hevm/src/EVM.hs
```diff
@@ -1857,7 +1857,15 @@ cheatActions =
               new  = w256lit $ word y
           fetchAccount a $ \_ -> do
             modifying (env . contracts . ix a . storage) (writeStorage slot new)
-          return Nothing
+          return Nothing,
+      action "load(address,bytes32)" [AbiAddressType, AbiBytesType 32] $
+        \[AbiAddress a, AbiBytes 32 x] -> do
+          let slot = w256lit $ word x
+          return $ use (env . contracts . at a) >>= \case
+            Just c ->
```
xwvvvvwx, line 1865: Don't we need to handle the `Nothing` side of these cases?

MrChico:
A version that works with symbolic values and integrates with rpc can be found here https://github.com/dapphub/dapptools/commit/19ae51d500259d4712995c3763fc8890441b1d44. Closing this

PR #522, opened 2020-10-01
Added util to `token` for EIP2612 and DAI permits

lucasvo:


lucasvo:
Tested and confirmed to work: https://kovan.etherscan.io/tx/0xbdd5a5c0785535f68a092245cb2411c1c49f6241774321b4a0d6e7e6e434f648 (EIP2612) and https://kovan.etherscan.io/tx/0x70499a3e54d4e1d22cab63bbb186ede820f7838f02d9916712cf8dcec91d8d33 (Kovan DAI)

src/token/libexec/token/token-permit
```diff
@@ -0,0 +1,77 @@
...
+if [ $# -lt 2 ];  then
+    display_usage
+    exit 1
+fi
+
+TOKEN=$1
+SPENDER=$2
+
+EIP2612_TYPEHASH=$(seth keccak $(seth --from-ascii "Permit(address owner,address spender,uint256 value,uint256 nonce,uint256 deadline)"))
+PERMIT_TYPEHASH=$(seth call $TOKEN "PERMIT_TYPEHASH()(bytes32)")
```
xwvvvvwx, line 30: what will `PERMIT_TYPEHASH` evaluate too if this call fails (i.e. there is no `PERMIT_TYPEHASH()` method on the contract)?

src/token/libexec/token/token-permit
```diff
@@ -0,0 +1,77 @@
...
+
+if [ "$EIP2612_TYPEHASH" != "$PERMIT_TYPEHASH" ]; then
+    DAI=yes
```
xwvvvvwx, line 33: Are there other tokens that use `dai` style permits? It might be more robust to hardcode the `dai` address here?

xwvvvvwx:
Some general comments:

- wondering if it might be nice to have some initial check to ensure that the token actually supports permit?
- Can we add some tests for this new feature?

src/token/libexec/token/token-permit
lucasvo: Chai uses it.

src/token/libexec/token/token-permit
lucasvo: I don't know but the update to EIP2616 is making it part of the standard that it should be a public method so we should be able to rely on it evaluating properly. I'll check

lucasvo:
Testing: Is there test coverage for other parts of bash scripts? I will have to look into it.

xwvvvvwx:
re: tests we have some testing scripts here: https://github.com/dapphub/dapptools/blob/master/src/dapp-tests/integration/tests.sh

src/token/libexec/token/token-permit
MrChico: Actually, the PERMIT_TYPEHASH is not added to 2612, but DOMAIN_SEPARATOR is.
But this feels okay to me anyway for the purposes of this script

MrChico:
Gonna merge this in 24hrs unless there are major disagreements or it seems likely that tests will actually appear

PR #526, opened 2020-10-07
Symtimestamp

MrChico:
Generalizes timestamp to admit symbolic values


rainbreak:
Do you want to do the same for block number as well?

MrChico:
Eventually, yes

PR #527, opened 2020-10-14
Bigbackstep

MrChico:
Implements `P` and `C-p` commands which takes larger steps back, analogously to how `N` and `C-n` step forward


lucasvo:
@MrChico you're my hero!

PR #530, opened 2020-10-22
Fall back to old method for detecting bytecode metadata

MrChico:
as the cborg based detection could not account for constructor args,
resulting in no srcmaps for init methods with arguments.

Revert "fix stripBytecodeMetadata"

This reverts commit 78508c6a8db2d6d3e8e09437dbe122bb5e6b2e7e.

### DIFF
--- a/src/hevm/src/EVM/Solidity.hs
+++ b/src/hevm/src/EVM/Solidity.hs
@@ -6,6 +6,7 @@
 module EVM.Solidity
   ( solidity
   , solcRuntime
+  , solidity'
   , JumpType (..)
   , SolcContract (..)
   , StorageItem (..)
@@ -66,7 +67,6 @@ import Data.Sequence        (Seq)
 import Data.Text            (Text, pack, intercalate)
 import Data.Text.Encoding   (encodeUtf8)
 import Data.Text.IO         (readFile, writeFile)
-import Data.Tuple           (swap)
 import Data.Vector          (Vector)
 import Data.Word
 import GHC.Generics         (Generic)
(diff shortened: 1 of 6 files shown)


xwvvvvwx:
wasn't the old method somehow broken?

MrChico:
It was missing some special cases, one that I added

xwvvvvwx:
Can we add some tests plz

MrChico:
of what exactly?

xwvvvvwx:
That this metadata stripping code actually does what we think it does. We've gone back and forth on implementations here a few times already...

src/hevm/src/EVM/Solidity.hs
```diff
@@ -440,14 +436,24 @@ solidity' src = withSystemTempFile "hevm.sol" $ \path handle -> do
...
+knownBzzrPrefixes :: [ByteString]
+knownBzzrPrefixes = [
+  -- a1 65 "bzzr0" 0x58 0x20 (solc <= 0.5.8)
+  BS.pack [0xa1, 0x65, 98, 122, 122, 114, 48, 0x58, 0x20],
+  -- a2 65 "bzzr0" 0x58 0x20 (solc >= 0.5.9)
+  BS.pack [0xa2, 0x65, 98, 122, 122, 114, 48, 0x58, 0x20],
+  -- a2 65 "bzzr1" 0x58 0x20 (solc >= 0.5.11)
+  BS.pack [0xa2, 0x65, 98, 122, 122, 114, 49, 0x58, 0x20],
+  -- a2 64 "ipfs" 0x58 0x22 (solc >= 0.6.0)
+  BS.pack [0xa2, 0x64, 0x69, 0x70, 0x66, 0x73, 0x58, 0x22]
```
rainbreak, line 455: Can solc > 0.6.0 still use the bzzr prefix as well?

rainbreak:
Looks reasonable, agree some basic tests would be nice.

src/hevm/src/EVM/Solidity.hs
xwvvvvwx: I think it's a command line flag: `--metadata-hash`

xwvvvvwx (approved):
lgtm

PR #531, opened 2020-10-22
Use `fetchTarball` instead of `fetchGit` to get `nixpkgs`

icetan:
Because of excruciating download times for `nixpkgs` when using `fetchGit` on CIs where no cache/store is already populated.


xwvvvvwx:
❤️

PR #537, opened 2020-11-07
ds-prove

xwvvvvwx:
Adds support for symbolic execution to `ds-test`.

Tests prefixed with `prove` will be symbolically executed, where the arguments to the test method are given as symbolic values. Test setup is still executed concretely, allowing users to instantiate a contract system in a concrete configuration and then execute symbolicaly against that state.

Symbolic tests use the `InitialS` storage model, meaning all state is defaulted to zero, and the tests are effectivly run against an empty blockchain. In the future it may be interesting to allow symbolic test execution against a fully symbolic VM (`SymbolicS` storage).

**Notable changes:**

- A new EVM query `PleaseMakeUnique` has been added, that asks the SMT solver to provide a concrete model for a symbolic variable iff the variable can be inhabited by a single value.
- All `call` like opcodes now use this new query to try and resolve the target of the call to a concrete value
- The output of `keccak` is probabalisticaly assumed to always be greater than 100, allowing us to ignore the potential for storage collisions when executing contracts that contain mappings
- `UnitTestOptions` now carries an `SBV.State`, allowing smt queries to be made without needing to be in the `Query` monad (see e.g. `initialUiVmStateForTest` in `TTY.hs`)
- The contents of the `Keccak` module have been moved to `Type` to workaround cyclic module dependency issues
- Addresses are now displayed in checksum format
- Removes the `hevm interactive` alias (as it was causing some unescessary duplication)

**Stil todo:**

- [x] support rpc
- [x] show traces in output for failing tests
- [x] support smt flags in `hevm interactive`
- [ ] tests
- [x] handle smt timeouts gracefully
- [ ] handle coverage (not really sure exactly how to do this...)
- [ ] add a `mayRevert` modifier (can maybe leave for a subsequent pr)


MrChico:
Added traces and ensured that failing a `testFail` test is actually a success

src/hevm/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 - A `--show-tree` option to `hevm symbolic` which prints the execution tree explored.
 - Some symbolic terms are displayed with richer semantic information, instead of the black box `<symbolic>`.
+- `hevm dapp-test` now supports symbolic execution of test methods that are prefixed with `prove` or `proveFail`
+- The `hevm interactive` alias has been removed
```
MrChico, line 10:
```suggestion
- The `hevm interactive` alias has been removed, as it is equivalent to `hevm dapp-test --debug`
```

src/hevm/hevm-cli/hevm-cli.hs
```diff
@@ -377,30 +369,28 @@ findJsonFile Nothing = do
...
-            in
-              Git.saveFacts (Git.RepoAt path) (Facts.cacheFacts cache')
-
-        unless (all id passing) exitFailure
-      Nothing ->
-        error ("Failed to read Solidity JSON for `" ++ solcFile ++ "'")
+dappTest :: UnitTestOptions -> String -> Maybe String -> Query ()
+dappTest opts solcFile cache = do
+  out <- liftIO $ readSolc solcFile
+  case out of
```
MrChico, line 375:
```suggestion
    liftIO $ readSolc solcFile >>= \case
```

xwvvvvwx: afk right now, but I think this doesn't work because we need to be in the `Query` monad for the call to `runUnitTestContract` below...

MrChico: I was hoping that `liftIO` would only bind as `liftIO (readSolc solcFile)`, and that the `>>= \case` would be over `Query` still

xwvvvvwx: Gives a type error on the `runUnitTestContract` line:

```
[typecheck] [E] • Couldn't match expected type ‘Maybe
                                  (Map Text SolcContract, SourceCache)
                                -> IO ()’
              with actual type ‘QueryT IO ()’
• In a stmt of a 'do' block:
    results <- concatMapM
                 (runUnitTestContract opts contractMap) unitTests
  In the expression:
    do let matcher = regexMatches (EVM.UnitTest.match opts)
           unitTests
             = fmap (filterTests matcher) $ findUnitTests $ elems contractMap
       results <- concatMapM
                    (runUnitTestContract opts contractMap) unitTests
       let (passing, vms) = unzip results
       case cache of
         Nothing -> pure ()
         Just path
           -> let cache' = mconcat [view cache vm | vm <- vms]
              in liftIO $ saveFacts (RepoAt path) (cacheFacts cache')
       liftIO $ unless (and passing) exitFailure
  In a case alternative:
      Just (contractMap, _)
        -> do let matcher = regexMatches (EVM.UnitTest.match opts)
                  unitTests
                    = fmap (filterTests matcher) $ findUnitTests $ elems contractMap
              results <- concatMapM
                           (runUnitTestContract opts contractMap) unitTests
              let (passing, vms) = unzip results
              case cache of
                Nothing -> pure ()
                Just path
                  -> let cache' = mconcat [view cache vm | vm <- vms]
                     in liftIO $ saveFacts (RepoAt path) (cacheFacts cache')
              liftIO $ unless (and passing) exitFailure
```

MrChico: ok, nvm, just looked like it could work

MrChico (approved):

Issue #541, opened 2020-11-11
0.41 -> 0.42 Breaking change for RPC: CallStack (from HasCallStack)


brockelmore:
One of these two things causes a breaking change for cheat-code heavy tests. I will try to get a minimally reproduced example. Here is full trace:
```
CallStack (from HasCallStack):
    error, called at src/EVM/Symbolic.hs:54:14 in hevm-0.42.0-1B7A8J2pBT6B3cFKLMhsva:EVM.Symbolic
```
Here are the changes that would affect it:
- Default to on chain values for coinbase, timestamp, difficulty, blocknumber when rpc is provided
- Perform tx initialization (gas payment, value transfer) in hevm exec, hevm symbolic and hevm dapp-test.

I imagine its the second. Here is the full command:
```
export ETH_RPC_URL="https://localhost:8545"

export DAPP_SRC=./contracts
export DAPP_SOLC_VERSION=0.5.15
export SOLC_FLAGS="--optimize --optimize-runs 50000"

dapp build

block=$(seth block latest)

export DAPP_TEST_TIMESTAMP=$(seth --field timestamp <<< "$block")
export DAPP_TEST_NUMBER=$(seth --field number <<< "$block")
export DAPP_TEST_ORIGIN="0x683A78bA1f6b25E29fbBC9Cd1BFA29A51520De84"
export DAPP_TEST_CHAINED=99
export DAPP_TEST_ADDRESS="0x683A78bA1f6b25E29fbBC9Cd1BFA29A51520De84"
printf 'Running test for address %s\n' "$DAPP_TEST_ADDRESS"
LANG=C.UTF-8 dapp test --rpc-url $ETH_RPC_URL -v --match test_kill
```

MrChico:
Interesting. The error is from [forcing a symbolic value to be concrete](https://github.com/dapphub/dapptools/blob/master/src/hevm/src/EVM/Symbolic.hs#L54), but your test should only be dealing with concrete values anyway, so I'm confused why this would happen. A minimal example would really help

brockelmore:
Found it. Here minimally reproducible:

```
// SPDX-License-Identifier: GPL-3
// This program is free software: you can redistribute it and/or modify
// it under the terms of the GNU General Public License as published by
// the Free Software Foundation, either version 3 of the License, or
// (at your option) any later version.

// This program is distributed in the hope that it will be useful,
// but WITHOUT ANY WARRANTY; without even the implied warranty of
[...]

PR #544, opened 2020-11-11
get storagemodel from PleaseFetchAccount, not in oracle

MrChico:
Passing the storagemodel in oracle can lead to inconsistencies. This solves #541.


xwvvvvwx (approved):
Looks great :sparkles:

If we can introduce a test that would cover this issue given the constraints of our current CI infrastructure I would be very happy, but feel free to merge if it's very annoying to do so....

MrChico:
The issue here has to do with rpc, so we would either have to spin up some more intricate testing environment in the style of dapp testnet or deal with live main chain state and some CI secrets

Issue #288, opened 2019-10-28
hevm: support json trace


MrChico:
`geth`, `parity` and `aleth` and others support a traced output in json format. The details differ... Here is geth:
```
{"pc":0,"op":96,"gas":"0x2540be400","gasCost":"0x0","memory":"0x","memSize":0,"stack":[],"depth":1,"refund":0,"opName":"PUSH1","error":""}
{"pc":2,"op":96,"gas":"0x2540be3fd","gasCost":"0x0","memory":"0x","memSize":0,"stack":["0x80"],"depth":1,"refund":0,"opName":"PUSH1","error":""}
{"pc":4,"op":82,"gas":"0x2540be3fa","gasCost":"0xc","memory":"0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000","memSize":96,"stack":["0x80","0x40"],"depth":1,"refund":0,"opName":"MSTORE","error":""}
{"pc":5,"op":52,"gas":"0x2540be3ee","gasCost":"0xc","memory":"0x000000000000000000000000000000000000000000000000000000000
```
`aleth`:
```
{"pc": 0, "gas": "0x7fffffffffffffff", "gasCost": "0x3", "stack": [], "depth": 0, "opName": "PUSH1"}
{"pc": 2, "gas": "0x7ffffffffffffffc", "gasCost": "0x3", "stack": ['0x80'], "depth": 0, "opName": "PUSH1"}
{"pc": 4, "gas": "0x7ffffffffffffff9", "gasCost": "0x3", "stack": ['0x80', '0x40'], "depth": 0, "opName": "MSTORE"}
{"pc": 5, "gas": "0x7fffffffffffffed", "gasCost": "0x2", "stack": [], "depth": 0, "opName": "CALLVALUE"}
{"pc": 6, "gas": "0x7fffffffffffffeb", "gasCost": "0x3", "stack": ['0x00'], "depth": 0, "opName": "DUP1"}
```

See more discussion here:
https://github.com/ethereum/tests/issues/249

This would be helpful for more testing of hevm, including random testing against other implementations

gumb0:
We plan to change the format to closer match geth https://github.com/ethereum/aleth/issues/5804

livnev:
Would also be cool to support some [parity tracing features](https://wiki.parity.io/JSONRPC-trace-module).

MrChico:
[...]

PR #547, opened 2020-11-16
Gethcompliance

MrChico:
superseedes #546.
This PR introduces the json tracing functionality discussed in #288, and sets up some basic infrastructure for differentially testing hevm against geth. In the process I found a couple of issues:
- If a PUSH opcode is given too short data, geth pads it to the appropriate length, while hevm just takes the data it is given, effectively leftpadding the bytestring. I don't believe this can ever actually make a meaningful difference, since the only setting in which this can occur is if the PUSH instruction is the final instruction of a contract, so the value will be ignored on the stack anyway. Regardless, I have adjusted hevm to also rightpad the data.
- hevm can panic in TTY mode if it tries to navigate beyond the final opcode (try for example `hevm exec --code 60006000601260136014601560166018f4 --debug --gas 0xffffff` and press `e`). This is fixed by making `_traceOpIx` into a `Maybe`.

I have added a simple python script that uses `hypothesis` for comparing the traces generated by hevm and geth in `dapp-test/integration/diff-fuzz.py`, but I am currently not asking it to run anything besides the examples provided. 

The reason is that hevm and geth have a slight difference in opinion when it comes to calling empty contracts, where hevm enters the new contracts and immediately calls STOP and exits, while geth skips the call entirely and just moves on to the next opcode. I don't want to address this discrepancy now and don't want to run tests that I know can fail on a particular input. Do you guys still think it's worth including the `diff-fuzz` script in its current form?

EDIT 18 Nov:
Also addresses a performance regression introduced in a001d0e6ca08efbd5f19a5b8ede7b85e6c8c2451 and optimizes `checkJump` further by replacing a `Vector.find` which is `O(n)` with a lookup which is `O(1)`. This is a performance improvement of roughly 35-40% for concrete testing.

### DIFF
--- a/src/hevm/hevm-cli/hevm-cli.hs
+++ b/src/hevm/hevm-cli/hevm-cli.hs
@@ -22,7 +22,7 @@ import qualified EVM.Flatten
 import qualified EVM.Stepper
 import qualified EVM.TTY
 import qualified EVM.Emacs
-import EVM.Dev (concatMapM)
+import EVM.Dev (concatMapM, interpretWithTrace)
 
 #if MIN_VERSION_aeson(1, 0, 0)
 import qualified EVM.VMTest as VMTest
@@ -154,6 +154,7 @@ data Command w
       , difficulty  :: w ::: Maybe W256       <?> "Block: difficulty"
       , chainid     :: w ::: Maybe W256       <?> "Env: chainId"
       , debug       :: w ::: Bool             <?> "Run interactively"
+      , jsontrace   :: w ::: Bool             <?> "Print json trace output at every step"
       , trace       :: w ::: Bool             <?> "Dump trace"
       , state       :: w ::: Maybe String     <?> "Path to state repository"
       , cache       :: w ::: Maybe String     <?> "Path to rpc cache repository"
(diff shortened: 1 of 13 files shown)


src/hevm/hevm-cli/hevm-cli.hs
```diff
@@ -179,11 +181,12 @@ data Command w
...
   | BcTest -- Run an Ethereum Blockhain/GeneralState test
-      { file    :: w ::: String    <?> "Path to .json test file"
-      , test    :: w ::: [String]  <?> "Test case filter - only run specified test method(s)"
-      , debug   :: w ::: Bool      <?> "Run interactively"
-      , diff    :: w ::: Bool      <?> "Print expected vs. actual state on failure"
-      , timeout :: w ::: Maybe Int <?> "Execution timeout (default: 10 sec.)"
+      { file      :: w ::: String    <?> "Path to .json test file"
+      , test      :: w ::: [String]  <?> "Test case filter - only run specified test method(s)"
+      , debug     :: w ::: Bool      <?> "Run interactively"
+      , jsontrace :: w ::: Bool             <?> "Print json trace output at every step"
```
xwvvvvwx, line 187: nit: alignment....

xwvvvvwx (approved):

src/hevm/hevm-cli/hevm-cli.hs
MrChico, line 187:
```suggestion
      , jsontrace :: w ::: Bool     <?> "Print json trace output at every step"
```

PR #552, opened 2020-11-22
hevm debug: step through the callstack using up and down arrows

MrChico:
Adds the ability to step through the callstack (actually the trace tree to be precise) by using the up and down arrow keys in debug mode.
As the trace pane grows, we now follow it so that new entries are always visible. This can look a little wonky when backstepping (sometimes we only see the last entry, hiding previous ones even though there's space), but I think this is good enough for now.
Also fixes an issue where it was possible to step past the end of execution (try pressing `e` and then `N` on master right now).


rainbreak:
Would be nice to be able to scroll the trace pane, like the memory pane can be scrolled now.

PR #554, opened 2020-11-24
Formatting

MrChico:


xwvvvvwx (approved):
🥰

rainbreak:
Were you using `logs` before this? I changed it to `formatSString` so that you can actually log strings with it.

rainbreak:
How about we have this?

- `log(bytes)` -> `formatSBytes` 
- `logs(bytes|string)` -> `formatSString`

MrChico:
Don't we need to add a `log` event to ds-test before this becomes useful? `logs` was pretty useful even when it wasn't being picked up by the custom handlers of UnitTest. It falls back to the formatting of logs given in `Format.hs`, which still looks pretty nice

MrChico:
We also need a smother failure mode for `formatSString`; it currently throws a hard error if the string is not utf8 decodable

rainbreak:
> Don't we need to add a `log` event to ds-test before this becomes useful? `logs` was pretty useful even when it wasn't being picked up by the custom handlers of UnitTest. It falls back to the formatting of logs given in `Format.hs`, which still looks pretty nice

Yes, but until we add that and the other ds-test improvements, you just need to add `event log(bytes)` to your test. The unittest handler doesn't care about the types, just the name of the event.

Right now I need something that will print human readable strings to the non-trace output so I can make sense of long tests with `logs("here we go...")` etc.

MrChico:
I see. It would be nice to merge the functionality of `formatTestLog` with the part of `Format.showTrace` that deals with logs printing. In `showTrace`, we already pretty print the contents of a log based on the abi. Then you could use `event log(string)` for your use case, and I a `log(bytes)` for mine

rainbreak:
No, you can't have both

```
event log(string);
event log(bytes);

function test_it() public {
  log("something");
}
```

this leads to `Error: No unique declaration found after argument-dependent lookup.`

MrChico:
in our two different test contract, I mean

PR #555, opened 2020-11-24
hevm trace improvements

rainbreak:


src/hevm/src/EVM/UnitTest.hs
```diff
@@ -670,9 +670,12 @@ formatTestLog events (Log _ args (topic:_)) =
     Just t -> case (Map.lookup (wordValue t) events) of
                    Nothing -> Nothing
                    Just (Event name _ _) -> case name of
-                     "logs" ->
+                     "log" ->
```
rainbreak, line 673: Like this @MrChico rather than 6fa91f1bca57037b40848c9b104f4a6675d4c69c

MrChico: I would rather keep `logs` as `bytes`, since that is the type they are given in test.sol

src/hevm/src/EVM/UnitTest.hs
```diff
@@ -670,9 +670,12 @@ formatTestLog events (Log _ args (topic:_)) =
...
                        Just $ formatSBytes args
 
+                     "logs" ->
+                       Just $ formatSString args
+
```
MrChico, line 678:
```suggestion
                     "log" ->
                       Just $ formatSString args

                     "logs" ->
                       Just $ formatSSBytes args

```

rainbreak:
@MrChico check out these vast simplifications.

src/hevm/src/EVM.hs
```diff
@@ -2120,6 +2118,11 @@ finishFrame how = do
       -- Pop to the previous level of the debug trace stack.
       popTrace
 
+      -- Pop the top frame.
+      assign frames remainingFrames
+      -- Install the state of the frame to which we shall return.
+      assign state (view frameState nextFrame)
+
```
MrChico, line 2125: nice

src/hevm/src/EVM/Format.hs
```diff
@@ -163,37 +199,71 @@ showTrace dapp trace =
     fullAbiMap = view dappAbiMap dapp
   in case view traceData trace of
     EventTrace (Log _ bytes topics) ->
-      case topics of
-        [] ->
-          mconcat
+      let logn = mconcat
+            [ "\x1b[36m"
+            , "log" <> (pack (show (length topics))) <> "("
+            , intercalate ", " (map (pack . show) topics)
```
MrChico, line 205: is there a `,` missing between the topic and the contents here?

src/hevm/src/EVM/Format.hs
```diff
@@ -163,37 +199,71 @@ showTrace dapp trace =
...
-                     , showValues [t | (t, NotIndexed) <- types] bytes
-                     -- todo: show indexed
-                     , "\x1b[0m"
-                     ] <> pos
-                 Nothing -> unknownTopic
-            Nothing -> unknownTopic
+          knownTopic name types = mconcat
+            [ "\x1b[36m"
+            , name
+            , showValues [t | (t, NotIndexed) <- types] bytes
```
MrChico, line 219: could use `unindexed`

src/hevm/src/EVM/UnitTest.hs
```diff
@@ -686,95 +686,58 @@ formatTestLog events (Log _ args (topic:_)) =
...
         Nothing -> Nothing
         Just (Event name _ types) ->
-          case (name, unindexed types) of
-            -- log(string)
-            ("log",  [AbiStringType]) ->
-              Just $ unquote $ showValue AbiStringType args
-            -- logs(bytes)
-            ("logs", [AbiBytesDynamicType]) ->
-              Just $ unquote $ showValue AbiStringType args
+          case (name <> parenthesise (abiTypeSolidity <$> (unindexed types))) of
```
MrChico, line 689: case maybeLitWord topic >>= \t1 -> Map.lookup (wordValue t1) events of

src/hevm/src/EVM/Format.hs
rainbreak: good catch, yes

src/hevm/src/EVM/UnitTest.hs
rainbreak: `logs` is now deleted again. Instead we have `log_bytes(bytes)`.

src/hevm/hevm.cabal
```diff
@@ -81,6 +81,7 @@ library
     ethjet/tinykeccak.h, ethjet/ethjet.h, ethjet/ethjet-ff.h, ethjet/blake2.h
   build-depends:
     QuickCheck                        >= 2.13.2 && < 2.15,
+    Decimal                           == 0.5.1,
```
MrChico, line 84: Can you add this dep to the stack yml as well?

rainbreak:
rebased

MrChico (approved):

PR #558, opened 2020-11-25
Update token---format

hujw77:
https://github.com/dapphub/dapptools/blob/6fa91f1bca57037b40848c9b104f4a6675d4c69c/src/token/libexec/token/token-supply#L6
https://github.com/dapphub/dapptools/blob/6fa91f1bca57037b40848c9b104f4a6675d4c69c/src/token/libexec/token/token-balance#L6
https://github.com/dapphub/dapptools/blob/6fa91f1bca57037b40848c9b104f4a6675d4c69c/src/token/libexec/token/token-allowance#L6

already tranfer result to `uint`

### DIFF
--- a/src/token/libexec/token/token---format
+++ b/src/token/libexec/token/token---format
@@ -1,5 +1,4 @@
 #!/usr/bin/env bash
 set -e
-x=$(seth --to-dec "$2")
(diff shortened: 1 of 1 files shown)


MrChico (approved):

MrChico:
Thanks @hujw77!

PR #559, opened 2020-11-26
include solc static builds in overlay

rainbreak:
For now we download the builds directly from the foundation. Later we
can make our own static builds.

Install all versions with

    nix-env -f . -iA solc-static-versions

Install a specific version with e.g.

    nix-env -f . -iA solc-static-versions.solc_0_4_12

solc binaries are then available in your path as e.g. `solc-0.4.12`

The static binaries are currently ~10MB each for linux and ~30MB each
for MacOS, making it feasible to install all versions.

Sizes: https://gist.github.com/rainbreak/3f2dd60faca44c644f2bc00b79b7a0b0


rainbreak:
If we wanted binary caching we could add `solc-static-versions` to release.nix. This wouldn't actually save us anything as we're just downloading a binary, but it would (possibly??) take some load off the solidity team's servers.

With the static binaries in place, we could look at making `dapp --use` just switch between static binary versions and remove the complicated use of nix. Similarly for `seth --use`. We would then have no gaps in our solc versions and things like `seth bundle-source` would work for more contracts.

xwvvvvwx (approved):
I love it :sparkling_heart:

MrChico:
This seems great. Is there any reason to not replace our current solcs with these static ones?

rainbreak:
I don't think so, except for perhaps reduced transparency as the source of these is whatever soliditylang.org says they are, rather than the build output in a github release.

The current nixpkgs solc would be preferable if it included z3 or cvc, but it doesn't. We could later look at making our own static builds with these included.

MrChico:
> The current nixpkgs solc would be preferable if it included z3 or cvc, but it doesn't. We could later look at making our own static builds with these included.

Hmm, I think they do? They are given as optional arguments, but I think they default to being included. At least the SMTChecker works fine with our currently built ones.

I would prefer to move over completely to the static builds

rainbreak:
> Hmm, I think they do? They are given as optional arguments, but I think they default to being included. At least the SMTChecker works fine with our currently built ones.

They are included if they are available on the system. The nixpkgs builds have z3 and cvc4 enabled by default. The solc-static-linux binaries are built in an Alpine docker environment, which does not include z3 or cvc4 as dependencies. Can you test out these static binaries with some smt stuff?

MrChico:
The static builds have z3 and cvc4 included, and the SMTChecker seems to work as expected

MrChico:
(I've mainly been using static builds via `dapp --use` for the last couple of months)

rainbreak:
> The static builds have z3 and cvc4 included, and the SMTChecker seems to work as expected

On Mac or Linux?

MrChico:
I believe on both

rainbreak:
The alpine build has dependencies defined here https://github.com/ethereum/solidity/blob/develop/scripts/install_deps.sh#L165 and added here https://github.com/ethereum/solidity/blob/develop/scripts/Dockerfile#L9, with

    apk add boost-dev boost-static build-base cmake git

I don't see how z3 can be included in the github built static binary.

xwvvvvwx:
Maybe they are just calling out to a `z3` / `cvc4` binary that is already present on your system @MrChico? I wonder if we can get away with just depending on the latest version of each in `dapp` and reusing them for all versions of `solc`?

MrChico:
I am able to use `SMTChecker` in a pure nix-shell on macos. I suggest one of you with a linux machine available try it as well:
```
pragma experimental SMTChecker;
contract Empty {
    constructor() payable {}
    fallback() external payable {}
}

contract Counter {
    event Hi(uint n);
    uint public count = 0;

    function increment(uint n) external {
        uint gasBefore = gasleft();
        for (uint i = 0; i < n; i++) {
            count = count + 1;
        }
        uint gasAfter = gasleft();
        emit Hi(gasBefore - gasAfter);
    }
}
```

xwvvvvwx:
I see `Warning: BMC analysis was not possible since no integrated SMT solver (Z3 or CVC4) was found.` when trying to compile the above file in an empty ubuntu container using static solc-0.7.4

xwvvvvwx:
same in a pure `nix-shell`

src/dapp/libexec/dapp/dapp---use
```diff
@@ -31,12 +38,14 @@ shift
 
 [[ "$#" -gt 0 ]] || usage
 
-if [ -z "$override" ]; then
-
-dapp "$@"
-
+if [[ -z "$bin" ]]; then
+  echo >&2 "${0##*/}: Could not find ${solc} in your path."
+  echo >&2 "${0##*/}: Try \`nix-env -f . -iA solc-static-versions.$solc\`"
```
xwvvvvwx, line 43: A local dapptools checkout is no longer the default workflow. This should probably be something like:

```suggestion
  echo >&2 "${0##*/}: Try \`nix-env -f https://github.com/dapphub/dapptools/archive/master.tar.gz -iA solc-static-versions.$solc\`"
```

rainbreak: Could we also use the `DAPPTOOLS` that is defined in the nix wrapper?

```
echo >&2 "${0##*/}: Try \`nix-env -f $DAPPTOOLS -iA solc-static-versions.$solc\`"
```

rainbreak:
@MrChico is the MacOS binary actually static? What do `file` and `ldd` show?

cameel:
> This seems great. Is there any reason to not replace our current solcs with these static ones?

Just a heads up - there's a flaw in pre-0.6.1 Mac builds (https://github.com/ethereum/solidity/issues/10183) and we'll be replacing them so if you're relying on the current checksum, please take that into account. The problem is pretty trivial on the surface - the version string used by these binaries slightly differs from builds for other platforms - but since it's embedded in the contract bytecode, contract verification would fail if performed using such a build.

I'm working on fixing that but this time I want to do full bytecode comparison run to avoid creating problematic binaries again and this is proving not that easy for earlier versions which do not have some of the CLI options.

> I don't see how z3 can be included in the github built solc-static-linux binary.

It's not. It wasn't in the alpine build and it's not in the new one on CircleCI. Here's the part of the config that builds the static Linux binary: https://github.com/ethereum/solidity/blob/v0.7.4/.circleci/config.yml#L462-L473

But this may change soon since we're considering enabling SMTChecker by default (https://github.com/ethereum/solidity/issues/10166).

BTW we've been slowly migrating off Travis and these builds were moved to CircleCI in 0.7.4. They're performed on Ubuntu now. Until 0.7.3 Travis was building on Alpine.

> is the MacOS binary actually fully static? What do file and ldd show?

It's not, but AFAIK it's not feasible to have a completely static one on macOS. If you run `otool`, it will show you the following dynamic libraries:

```
 	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 902.1.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1281.100.1)
```

MrChico:
hmm, so I guess the macOS binary was able to link to my local z3 then even though I was in a pure nix shell...?
Allowing both for a while seems like the best option then

MrChico (approved):

rainbreak:
Thank you for the helpful context @cameel!

rainbreak:
> hmm, so I guess the macOS binary was able to link to my local z3 then even though I was in a pure nix shell...?
> Allowing both for a while seems like the best option then

The macos binary has z3 included https://github.com/ethereum/solidity/blob/develop/.circleci/config.yml#L582. I'm not sure about cvc4. The linux static binary is the only build that explicitly has z3 disabled.

rainbreak:
One issue here then is that anyone that was previously using `dapp --use solc:x.y.z` on Linux with the SMTChecker, will now have failures (I get segfaults). As-is, they will have to either

1. name their binaries as `solc-x.y.z` and use `dapp --use solc-x.y.z`, or
2. give a direct path to the solc binary

The number of people this effects is probably small (maybe just you @xwvvvvwx ?!), but maybe there's a more user friendly way of doing this.

rainbreak:
I've chosen to relax the solc naming requirements, so affected users could even do `dapp --use solc`.

xwvvvvwx:
I don't really follow, how would I use the smt checker on linux with this new relaxed naming scheme?

xwvvvvwx:
fwiw I think the SMT checker is a really nice feature I think we should have it enabled in the solc builds that we ship with `dapp`

MrChico:
but if the user has `z3` installed the smt checker will work, no?

xwvvvvwx:
I think it also needs to be linked in as a library at build time?

rainbreak:
> I don't really follow, how would I use the smt checker on linux with this new relaxed naming scheme?

The `solc` installed by dapptools is still the z3 enabled nixpkgs version.

```
nix-env -f https://github.com/dapphub/dapptools/archive/master.tar.gz -iA solc-versions.solc_0_6_12
dapp --use solc build
```

rainbreak:
> fwiw I think the SMT checker is a really nice feature I think we should have it enabled in the solc builds that we ship with `dapp`

What would you prefer for now @xwvvvvwx - shall I make `dapp --use solc:x.y.z` use the `nix run` method again?

xwvvvvwx:
On balance I think having a wider range of solc versions available via `dapp --use` is more valuable than having the smt checker enabled for all of those versions.

I'm ok with leaving this PR as is.

rainbreak:
Follow https://github.com/ethereum/solidity/issues/10420 for more on z3 static builds.

PR #566, opened 2020-12-07
fix branching bug

MrChico:
The simplifications involved in #552  messed up the ability to select branches while debugging symbolic executions. Here's a simple fix


xwvvvvwx:
needs a changelog entry, but apart from that lgtm

MrChico:
will add soon

PR #568, opened 2020-12-08
UnitTest: tidy result in counter examples

xwvvvvwx:
This commit tidies the error messages shown for counter examples in symbolic tests.

- Failures due to an SMT timeout are now shown in the result field instead of in the calldata field
- Failures due to a DSTest assertion violation are now shown as such instead of the "32 symbolic bytes shown previously"

Note: the current text in the result field for DSTest assertion violations is perhaps a little too similar to that for an EVM assertion violation, I'm not sure what a better choice would be?

## Old output

SMT Timeout:

```
  Counter Example:

    result:   Return: 32 symbolic bytes
    calldata: SMT Timeout
```

DSTest Assertion Violation:

```
  Counter Example:

    result:   Return: 32 symbolic bytes
    calldata: prove_fail()
```

## New output

SMT timeout:

```
  Counter Example:

    result:   Failed: SMTTimeout
    calldata: unknown
```

DSTest Assertion Violation:

```
  Counter Example:

    result:   DSTest Assertion Violation
    calldata: prove_fail()
```

### DIFF
--- a/src/hevm/src/EVM/UnitTest.hs
+++ b/src/hevm/src/EVM/UnitTest.hs
@@ -581,7 +581,9 @@ symFailure UnitTestOptions {..} testName failures' = mconcat
     showRes vm = let Just res = view result vm in
                  case res of
                    VMFailure _ -> prettyvmresult res
-                   VMSuccess _ -> "DSTest Assertion Violation"
+                   VMSuccess _ -> if "proveFail" `isPrefixOf` testName
+                                  then "Successful execution"
+                                  else "DSTest Assertion Violation"
     mkMsg (vm, cd) = pack $ unlines
       ["Counter Example:"
       ,""
--- a/src/hevm/src/EVM/UnitTest.hs
+++ b/src/hevm/src/EVM/UnitTest.hs
@@ -583,7 +583,7 @@ symFailure UnitTestOptions {..} testName failures' = mconcat
                    VMFailure _ -> prettyvmresult res
                    VMSuccess _ -> if "proveFail" `isPrefixOf` testName
                                   then "Successful execution"
-                                  else "DSTest Assertion Violation"
+                                  else "Failed: DSTest Assertion Violation"
     mkMsg (vm, cd) = pack $ unlines
       ["Counterexample:"
       ,""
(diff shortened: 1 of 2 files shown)


src/hevm/src/EVM/UnitTest.hs
```diff
@@ -577,8 +578,10 @@ symFailure UnitTestOptions {..} testName failures' = mconcat
   , intercalate "\n" $ indentLines 2 . mkMsg <$> failures'
   ]
   where
-    showRes vm = let Just res = view result vm
-                 in prettyvmresult res
+    showRes vm = let Just res = view result vm in
+                 case res of
+                   VMFailure _ -> prettyvmresult res
+                   VMSuccess _ -> "DSTest Assertion Violation"
```
xwvvvvwx, line 584: Unless I'm missing something, it should be safe to assume that all vm's at this point that have not failed are the result of a ds test assertion violation?

MrChico: Hmm, hard to see the context with this diff, but what about proveFail cases?

xwvvvvwx: v good point.

xwvvvvwx: Think I have this handled now.

xwvvvvwx:
Updated to handle `proveFail`. 

Examples output:

```
contract TestRes is DSTest {
    function prove_assertion_failed() public {
        assertTrue(false);
    }

    function prove_smttimeout(uint x, uint y, uint z) public {
        if (x == 0 || y == 0 || z == 0 || x * y == 0 || x * y * z == 0) return;

        if ((x * y / z) * (x / y) / (x * y) == (x * x * x * y * z / x * z * y)) {
            assertTrue(false);
        } else {
            assertTrue(true);
        }
    }

    function proveFail_success() public {
        assertTrue(true);
    }
}
```

```
[FAIL] prove_smttimeout(uint256,uint256,uint256)
[FAIL] proveFail_success()
[FAIL] prove_assertion_failed()

Failure: prove_assertion_failed()

  Counterexample:

    result:   DSTest Assertion Violation
    calldata: prove_assertion_failed()


Failure: proveFail_success()

  Counterexample:

    result:   Successful execution
    calldata: proveFail_success()


Failure: prove_smttimeout(uint256,uint256,uint256)

  Counterexample:

    result:   Failed: SMTTimeout
    calldata: unknown
```

MrChico (approved):

PR #575, opened 2020-12-21
token: fix token --parse to include 0x prefix on uint hex data

livnev:


src/token/libexec/token/token---parse
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 set -e
 x=$(seth --to-hex "$1")
-printf %64s "${x#0x}" | tr ' ' 0
+echo "0x$(printf %64s "${x#0x}" | tr ' ' 0)"
```
MrChico, line 4: why isn't this whole script just `seth --to-uint256`?

MrChico: ah, right, you are essentially doing this in #576 . I like that

livnev:
closing in favour of #576

PR #576, opened 2020-12-21
token: simplify amount parsing

livnev:


MrChico:
I prefer this solution

PR #579, opened 2021-01-12
Dappdispatcher

MrChico:
Makes the `/libexec/dapp` a proper dispatcher, following the style of `/libexec/seth`. The main benefit is robustness of flag parsing and the fact that `dapp <cmd> --help` now works


src/dapp/libexec/dapp/dapp
```diff
@@ -6,6 +6,47 @@
...
+dapp <command> --help
+--
+  Compilation options:
+optimize             activate solidity optimizer
+legacy               compile using the '--combined-json' format
+extract              after building, write the .abi, .bin and .bin-runtime. Implies '--legacy'
+
+  Testing options:
+verbosity=<number>   verbosity of 'dapp test' output (0-3)
+v,verbose            sets verbosity to 1
```
xwvvvvwx, line 19: shouldn't we have `-vv` and `-vvv` here?

xwvvvvwx (approved):

src/dapp/libexec/dapp/dapp
xwvvvvwx: maybe worth a note in the changelog

MrChico: I ended up putting them back in anyway for now

PR #580, opened 2021-01-14
default to 0.8.6 / solc 0.8 style assertion support

d-xo:


xwvvvvwx:
OK, ci is failing, guess I need to go and update all our test contracts as well...

MrChico:
might need to restrict the ranges of abivalues to get this through

TransmissionsDev:
can we at least default to 0.7.6, 0.6.x feels pretty old at this point 😅

MrChico:
I support that

d-xo:
@MrChico solc 0.8.X treats all arithmetic overflow as an assertion violation, this feels like it will be a bit annoying for `hevm symbolic` checking. I think I'm probably going to default to ignoring assertions due to arithmetic overflow, and introduce a flag (e.g. `--check-overflow`) that would also report panics due to arithmetic overflow.

Alternatively we could introduce an `--assertions` flag and allow users to provide a list of the error codes that they want `hevm` to find (full list here: https://docs.soliditylang.org/en/v0.8.6/control-structures.html?highlight=Panic#panic-via-assert-and-error-via-require).

nix/hevm-tests/default.nix
```diff
@@ -1,6 +1,6 @@
 { pkgs }:
 let
-  solc = "${pkgs.solc-versions.solc_0_6_7}/bin/solc";
+  solc = "${pkgs.solc-static-versions.solc_0_8_6}/bin/solc-0.8.6";
   solidity = pkgs.fetchFromGitHub {
```
d-xo, line 4: would be nice to bump this at some point as well to get the latest smtchecker tests, but I left it for now since it's probably a pretty big job and this PR was already fairly time intensive...

d-xo:
Think this is good to go. Ended up going with an `--asertions` flag that lets users customize which assertions they want hevm to find. By default we look for everything except arithmetic overflow.

I ended up leaving most of the test contracts on 0.6.7 as dappsys doesn't support 0.8 and I was running into this annoying bytestring parse issue with 0.7.6 (I wanna leave that for another day).

MrChico:
ok, so this is more about finding assertions in solc0.8 than making it the default? Can we change the title otherwise I'll be confused

d-xo:
hmmmm yes, in the end it kinda turned into that. I did still make `solc-0.8.6` the default for `dapp`, but I just ended up leaving `0.6.7` as the default for the test contracts under `src/dapp-tests/pass` and `src/dapp-tests/fail`.

PR #582, opened 2021-01-18
update sbv

MrChico:
Fixes a common failure while using cvc4 for ds-test, described at https://github.com/LeventErkok/sbv/issues/574


MrChico:
With this change we can consider again making cvc4 the default solver

PR #584, opened 2021-01-21
bugfix for optimizer setting

Eenae:
`DAPP_BUILD_OPTIMIZE` env var was ignored because of a wrong data type in the comparison. As a result, the optimizer was always disabled. Fixed.


MrChico:
Ah, good catch. Thanks!

PR #594, opened 2021-01-29
hevm: allow symbolic arguments to cheatcodes

xwvvvvwx:
hevm cheat codes can now accept symbolic arguments, allowing symbolic jumps in time, or `store` / `load` for symbolic values. `roll` will still fail with `UnexpectedSymbolicArg` as the block number is currently not symbolic at the VM level.

### DIFF
--- a/src/hevm/src/EVM/ABI.hs
+++ b/src/hevm/src/EVM/ABI.hs
@@ -26,6 +26,7 @@
 -}
 
 {-# Language StrictData #-}
+{-# Language DataKinds #-}
 
 module EVM.ABI
   ( AbiValue (..)
@@ -46,6 +47,7 @@ module EVM.ABI
   , emptyAbi
   , encodeAbiValue
   , decodeAbiValue
+  , decodeBuffer
   , formatString
   , parseTypeName
   , makeAbiValue
@@ -56,7 +58,7 @@ module EVM.ABI
 import EVM.Types
 
 import Control.Monad      (replicateM, replicateM_, forM_, void)
-import Data.Binary.Get    (Get, runGet, label, getWord8, getWord32be, skip)
+import Data.Binary.Get    (Get, runGet, runGetOrFail, label, getWord8, getWord32be, skip)
 import Data.Binary.Put    (Put, runPut, putWord8, putWord32be)
 import Data.Bits          (shiftL, shiftR, (.&.))
 import Data.ByteString    (ByteString)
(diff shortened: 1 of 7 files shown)


src/hevm/src/EVM/ABI.hs
```diff
@@ -537,6 +540,28 @@ listP parser = between (char '[') (char ']') ((do skipSpaces
                                                   return a) `sepBy` (char ','))
 
 
+-- TODO: teach this to handle complex types for symbolic buffers
+decodeBuffer :: [AbiType] -> Buffer -> Maybe (Either [AbiValue] [SWord 256])
```
xwvvvvwx, line 544: The return type is a little gross here, dunno if there is a cleaner way... :thinking:

src/hevm/src/EVM/ABI.hs
```diff
@@ -537,6 +540,28 @@ listP parser = between (char '[') (char ']') ((do skipSpaces
...
+      fmap (\i -> fromBytes $ take 32 (drop (i*32) b)) [0..((length b) `div` 32 - 1)]
+
+containsComplexTypes :: [AbiType] -> Bool
+containsComplexTypes [] = False
+containsComplexTypes (hd : tl) = case hd of
+  AbiBytesDynamicType -> True
+  AbiStringType -> True
+  AbiArrayDynamicType _ -> True
+  AbiArrayType _ _ -> True
+  AbiTupleType _ -> True
```
MrChico, line 563: tuples are only complex if their arguments are; `AbiTupleType [a,b]` is equivalent to `[a, b]`

src/hevm/src/EVM/ABI.hs
MrChico: do we really need to special case the concrete case here? Can't we just have:
` decodeBuffer :: [AbiType] -> Buffer -> Maybe [SWord 256]`?

MrChico: If one needs concreteness that can be taken into account at the callsite. Current uses all do the same thing regardless of concreteness

src/hevm/src/EVM/ABI.hs
MrChico: Also, this function is actually just a copy of `abiKind a == Dynamic`

src/hevm/src/EVM/ABI.hs
MrChico: would also be nice to use this to clean up `getStaticAbiArgs` of `test.hs`

MrChico (approved):

MrChico:
do u wanna push a changelog scribble? Can do straight to master

xwvvvvwx:
done.

PR #606, opened 2021-02-09
hevm: add sign and addr cheatcodes

### LINKED ISSUE #549: `sign` cheatcode
Would be of great value imo

xwvvvvwx:
- sign allows you to sign a message with a given private key
- addr allows you derive the ethereum address for a given private key

taken together they should allow easy fuzzing for contracts that deal with signed messages.

As a side effect this PR also bumps `nixpkgs`, which means we're compiling with ghc 8.8.4, and can add `haskell-language-server` to the hevm `shell.nix`.

fixes #549


src/hevm/src/EVM.hs
```diff
@@ -1880,37 +1883,82 @@ type CheatAction = Word -> Word -> Buffer -> EVM ()
...
+                    assign (state . memory . word256At outOffset) (w256lit . num $ addr)
           _ -> vmError (BadCheatCode sig)
+
     ]
   where
-    action s f = (abiKeccak s, f (Just $ abiKeccak s))
+    action s tps f = (abiKeccak s, f (Just $ abiKeccak s) tps)
     mksym x = S (Todo "abidecode" []) x
 
+-- | Hack deterministic signing, totally insecure...
```
xwvvvvwx, line 1954: This is pretty sketchy. If we were doing this properly we would either:

1. Implement [rfc-6979](https://tools.ietf.org/html/rfc6979) (perhaps in cryptonite itself).
2. Add a new `Query` that lets us get random numbers

src/hevm/src/EVM.hs
```diff
@@ -1880,37 +1883,82 @@ type CheatAction = Word -> Word -> Buffer -> EVM ()
...
+              encoded = encodeAbiValue $
+                AbiTuple (RegularVector.fromList
+                  [ AbiUInt 8 v
+                  , AbiBytes 32 (encodeAbiValue . AbiUInt 256 . fromInteger $ sign_r s)
+                  , AbiBytes 32 (encodeAbiValue . AbiUInt 256 . fromInteger $ sign_s s)
+                  ])
+            in do
+              assign (state . returndata) (ConcreteBuffer encoded)
+              copyBytesToMemory (ConcreteBuffer encoded) (num . BS.length $ encoded) 0 outOffset
+          _ -> vmError (BadCheatCode sig),
```
xwvvvvwx, line 1929: One thing that's a little weird here is that we throw a `BadCheatCode` instead of an  `UnexpectedSymbolicArg` in case someone tries to call `hevm.sign` with a symbolic argument...

src/hevm/src/EVM/ABI.hs
```diff
@@ -540,6 +542,21 @@ listP parser = between (char '[') (char ']') ((do skipSpaces
                                                   skipSpaces
                                                   return a) `sepBy` (char ','))
 
+data AbiVals = NoVals | CAbi [AbiValue] | SAbi [SWord 256]
+
+decodeBuffer :: [AbiType] -> Buffer -> AbiVals
```
xwvvvvwx, line 547: This is actually not needed in the end, but I'm inclined to keep it because I think it's pretty useful. Do you agree @MrChico? I'm happy to remove it if you prefer to keep things lean...

src/hevm/src/EVM.hs
xwvvvvwx: no longer valid....

src/hevm/src/EVM.hs
xwvvvvwx: I think I'm going to implement #2 next week when we have a bit more free time...

default.nix
```diff
@@ -1,11 +1,11 @@
...
 
 let
-  rev = "5272327b81ed355bbed5659b8d303cf2979b6953";
+  rev = "8e78c2cfbae71720869c35b6710d652bf28b37cc";
   nixpkgs = builtins.fetchTarball {
-    name = "nixpkgs-release-20.03";
+    name = "nixpkgs-release-20.09";
     url = "https://github.com/nixos/nixpkgs/tarball/${rev}";
-    sha256 = "0182ys095dfx02vl2a20j1hz92dx3mfgz2a6fhn31bqlp1wa8hlq";
+    sha256 = "178dr6bz71lbv0ljynvvkrc2p0lwqmci482brkqdw9qfx3sc1a7f";
```
MrChico, line 8: I think this bump allows us to drop some of the custom packages we've been keeping. `z3` and `cvc4` are up to par or newer on `20.09` so we can drop our own versions. Looking at our other custom packages, we may even be able to drop `libff`, but that might be more housecleaning than what we wanna do in this PR

src/dapp-tests/integration/tests.sh
```diff
@@ -23,22 +23,22 @@ dapp_testnet() {
 
   dapp testnet --dir "$TMPDIR" &
   # give it a few secs to start up
-  sleep 90
+  sleep 900
```
```suggestion
  sleep 90
```

src/hevm/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # hevm changelog
 
+## Unreleased
+
+### Added
+
+- Two new cheatcodes were added: `sign(uint sk, bytes message)` and `addr(uint sk)`. Taken together
```
MrChico, line 7: might be good to mention the bump in nixpkgs / ghc too (also in `dapp`, `ethsign` and `seth`)

src/hevm/src/EVM.hs
```diff
@@ -1905,12 +1908,63 @@ cheatActions =
...
+                case digest of
+                  Nothing -> vmError (BadCheatCode sig)
+                  Just digest' -> do
+                    let s = ethsign priv digest'
+                        v = if (sign_s s) % 2 == 0 then 27 else 28
+                        encoded = encodeAbiValue $
+                          AbiTuple (RegularVector.fromList
+                            [ AbiUInt 8 v
+                            , AbiBytes 32 (encodeAbiValue . AbiUInt 256 . fromInteger $ sign_r s)
+                            , AbiBytes 32 (encodeAbiValue . AbiUInt 256 . fromInteger $ sign_s s)
```
MrChico, line 1931:
```suggestion
                            , AbiBytes 32 (word256Bytes . fromInteger $ sign_r s)
                            , AbiBytes 32 (word256Bytes . fromInteger $ sign_s s)
```
or something like this should be possible and simpler

src/hevm/src/EVM.hs
```diff
@@ -1905,12 +1908,63 @@ cheatActions =
               accessStorage a' (mksym slot) $ \res -> do
                 assign (state . returndata . word256At 0) res
                 assign (state . memory . word256At outOffset) res
+          _ -> vmError (BadCheatCode sig),
+
+      action "sign(uint256,bytes32)" $
+        \sig outOffset _ input -> case decodeStaticArgs input of
+          [sk, hash] ->
+            forceConcrete (mksym sk) $ \sk' ->
```
MrChico, line 1916: I don't really get what you are doing here, tbh. `forceConcrete (mksym sk) $ \sk' ->` seems like a very arcane noop

src/hevm/src/EVM.hs
```diff
@@ -1905,12 +1908,63 @@ cheatActions =
...
+                      pub = BS.concat [ encodeInt x, encodeInt y ]
+                      addr = w256lit . num . word256 . BS.drop 12 . BS.take 32 . keccakBytes $ pub
+                    assign (state . returndata . word256At 0) addr
+                    assign (state . memory . word256At outOffset) addr
           _ -> vmError (BadCheatCode sig)
+
     ]
   where
     action s f = (abiKeccak s, f (Just $ abiKeccak s))
     mksym x = S (Todo "abidecode" []) x
```
MrChico, line 1958: better to use `litWord` here imo, but I'm not sure I understand the need for `mksym` in the first place

src/hevm/src/EVM.hs
MrChico: nvm, I get it. Root problem is `decodeStaticArgs` destroys the whiff information

src/hevm/src/EVM.hs
MrChico: ah, ok. I get it now. I guess the root of the problem is that `decodeStaticArgs` forgets the whiff (should be captured as `frombuffer` whiff)

src/hevm/src/EVM.hs
```diff
@@ -1905,12 +1908,63 @@ cheatActions =
...
 
+-- | Hack deterministic signing, totally insecure...
+ethsign :: PrivateKey -> Digest Crypto.Keccak_256 -> Signature
+ethsign sk digest = go 420
```
MrChico, line 1962: is there a chance this could loop forever?

src/hevm/src/EVM/ABI.hs
MrChico: we can keep it around for a bit

src/hevm/src/EVM.hs
xwvvvvwx: hmmmmmmm, I'm not sure, it's pretty much exactly the same as the implementation in `cryptonite`  ([src](https://github.com/haskell-crypto/cryptonite/blob/cryptonite-v0.27/Crypto/PubKey/ECC/ECDSA.hs#L100)), just with a monotonically increasing `k` instead of a randomly selected one...

src/hevm/src/EVM.hs
xwvvvvwx: Decided to leave it as it is in the end. Signatures produced by `hevm.sign` will leak the private key, but that seems fine for our purposes....

default.nix
xwvvvvwx: yeah lets tackle that in a seperate PR.

MrChico: lets still drop z3 and cvc4 here tho

PR #613, opened 2021-02-17
Added support for DAPP_BUILD_OPTIMIZE_RUNS

hexonaut:
Allows setting the number of runs via env vars.


xwvvvvwx:
Can you document this with the other env vars in the `dapp` readme? otherwise lgtm. Thanks!!

hexonaut:
Updated. I think that "." after the `DAPP_BUILD_OPTIMIZE` was a typo, but lmk otherwise.

xwvvvvwx (approved):
Thanks :pray: :sparkles:

PR #615, opened 2021-02-24
dont use networkId as chainid

MrChico:


xwvvvvwx (approved):
What even is the network id?

MrChico:
see https://medium.com/@pedrouid/chainid-vs-networkid-how-do-they-differ-on-ethereum-eec2ed41635b for full details. On most networks, they are the same

PR #626, opened 2021-03-19
Admit symbolic bytecode

MrChico:
Allows for constructor arguments to be symbolic, as in [this example](https://github.com/dapphub/dapptools/blob/constructorArg/src/dapp-tests/pass/dsProvePass.sol#L38)

### DIFF
--- a/src/hevm/src/EVM.hs
+++ b/src/hevm/src/EVM.hs
@@ -283,13 +283,6 @@ data ContractCode
   | RuntimeCode Buffer  -- ^ "Instance" code, after contract creation
   deriving (Show)
 
-instance Eq ContractCode where
-  InitCode (ConcreteBuffer a) == InitCode (ConcreteBuffer b) = a == b
-  RuntimeCode (ConcreteBuffer a) == RuntimeCode (ConcreteBuffer b) = a == b
-  InitCode _ == RuntimeCode _ = False
-  RuntimeCode _ == InitCode _ = False
-  _ == _ = error "cannot compare symbolic values"
-
 -- | A contract can either have concrete or symbolic storage
 -- depending on what type of execution we are doing
 data Storage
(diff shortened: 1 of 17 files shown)


src/hevm/src/EVM.hs
```diff
@@ -554,23 +553,24 @@ exec1 = do
...
       case ?op of
 
         -- op: PUSH
         x | x >= 0x60 && x <= 0x7f -> do
           let !n = num x - 0x60 + 1
-              !xs = padRight n $ BS.take n (BS.drop (1 + the state pc)
-                                        (the state code))
+              !xs = case the state code of
+                      ConcreteBuffer b -> w256lit $ word $ padRight n $ BS.take n (BS.drop (1 + the state pc) b)
+                      SymbolicBuffer b -> readSWord' 0 $ padLeft' 32 $ take n $ drop (1 + the state pc) b
```
MrChico, line 569:
```suggestion
                      SymbolicBuffer b -> readSWord' 0 $ take n $ drop (1 + the state pc) b
```

src/hevm/src/EVM.hs
```diff
@@ -554,23 +553,24 @@ exec1 = do
...
       case ?op of
 
         -- op: PUSH
         x | x >= 0x60 && x <= 0x7f -> do
           let !n = num x - 0x60 + 1
-              !xs = padRight n $ BS.take n (BS.drop (1 + the state pc)
-                                        (the state code))
+              !xs = case the state code of
+                      ConcreteBuffer b -> w256lit $ word $ padRight n $ BS.take n (BS.drop (1 + the state pc) b)
+                      SymbolicBuffer b -> readSWord' 0 $ take n $ drop (1 + the state pc) b
```
MrChico, line 569:
```suggestion
                      SymbolicBuffer b -> readSWord' 0 $ padLeft' 32 $ take n $ drop (1 + the state pc) b
```

PR #630, opened 2021-03-29
Add Multi-File Etherscan Verification Option

naszam:
Description:

Add `--multiple` opt to `dapp-verify-contract`

### DIFF
--- a/src/dapp/libexec/dapp/dapp-mk-standard-json
+++ b/src/dapp/libexec/dapp/dapp-mk-standard-json
@@ -31,7 +31,7 @@ for f in files:
     tmpljson["sources"][f]={}
     if os.getenv('DAPP_VERIFY_CONTRACT_MULTIPLE') == "yes":
         tmpljson["sources"][f]["content"] = [f.read()]
-    else
+    else:
         tmpljson["sources"][f]["urls"] = [f]
 
 if os.getenv('DAPP_REMAPPINGS') is None or os.getenv('DAPP_REMAPPINGS') == "":
--- a/src/dapp/libexec/dapp/dapp-mk-standard-json
+++ b/src/dapp/libexec/dapp/dapp-mk-standard-json
@@ -29,7 +29,10 @@ for root, dirnames, filenames in os.walk(src):
 tmpljson["sources"]={}
 for f in files:
     tmpljson["sources"][f]={}
-    tmpljson["sources"][f]["urls"] = [f]
+    if os.getenv('DAPP_VERIFY_CONTRACT_MULTIPLE') == "yes":
+        tmpljson["sources"][f]["content"] = [f.read()]
+    else
+        tmpljson["sources"][f]["urls"] = [f]
 
 if os.getenv('DAPP_REMAPPINGS') is None or os.getenv('DAPP_REMAPPINGS') == "":
     tmpljson["settings"].pop("remappings", None)
--- a/src/dapp/libexec/dapp/dapp-verify-contract
+++ b/src/dapp/libexec/dapp/dapp-verify-contract
@@ -92,7 +92,7 @@ else
   optimized=0
 fi
 
-if  [[ $1 = "--multiple" ]]; then
+if  [[ $DAPP_VERIFY_CONTRACT_MULTIPLE == yes ]]; then
 
   # Standard Input JSON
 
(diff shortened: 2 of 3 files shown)


src/dapp/libexec/dapp/dapp-verify-contract
```diff
@@ -88,41 +92,87 @@ else
   optimized=0
 fi
 
-params=(
+if  [[ $1 = "--multiple" ]]; then
```
xwvvvvwx, line 95: Can you please rework this to use the unified option parsing code in `src/dapp/libexec/dapp/dapp`. You need to parse the `--multiple` flag using the infrastructure in place there and then conditionally set an environment variable (e.g. `DAPP_VERIFY_CONTRACT_MULTIPLE`) if the `--multiple` flag is present.

You can then branch on the envrionment variable here.

src/dapp/libexec/dapp/dapp-verify-contract
```diff
@@ -88,41 +92,87 @@ else
...
+    # NOTE: the Arguements typo is in etherscan's API
 
-$source
+    status=$(jshon <<<"$response" -e status -u)
+    guid=$(jshon <<<"$response" -e result -u)
+    message=$(jshon <<<"$response" -e message -u)
+    count=$((count + 1))
+
+    [[ $status = 1 ]] && break;
+  done
```
xwvvvvwx, line 136: This same code is used in both branches of the if statement, can you please refactor to remove the duplication (e.g. by pulling this block out of the `if` or extracting it into a seperate function)?

xwvvvvwx (commented):
Thanks for the submission! Very much appreciated :sparkles: :pray: 

I'm wondering if this should be the default verification method used in `dapp create --verify`, @MrChico do you have opinions here?

MrChico:
I don't think this should be the default verification method right now

src/dapp/libexec/dapp/dapp-verify-contract
naszam: Done. Let me know if it's ok!

src/dapp/libexec/dapp/dapp-verify-contract
naszam: Done. Let me know if it's ok! 
I'll send a pr on `src/dapp/libexec/dapp` with the env variable asap :slightly_smiling_face:

naszam:
If the proposed changes are ok, just need to explore the inputJSON a bit more, if working on the missing `libraries` field or trying to fix and use `mk-standard-json` instead to generate the inputJSON (that currently doesn't work out-of-the-box), then it will be ready for testing.

MrChico:
the changes so far look good! With `mk-standard-json` accomodated as discussed this should make for a nice improvement!

xwvvvvwx:
oh, also please update the readme to mention these new options

xwvvvvwx:
the close was an accident :woman_facepalming:

src/dapp/libexec/dapp/dapp-mk-standard-json
```diff
@@ -27,9 +27,19 @@ for root, dirnames, filenames in os.walk(src):
...
-    tmpljson["sources"][f]["urls"] = [f]
+
+if os.getenv('DAPP_VERIFY_CONTRACT_MULTIPLE') == "yes":
+    for f in files:
+        if f.endswith(".t.sol"):
+            continue
+        else:
+            tmpljson["sources"][f]={}
+            content = open(f, "r");
+            tmpljson["sources"][f]["content"] = [content.read()]
```
WilfredTA, line 38: `content.read()` instead of `[content.read()]` to enable successful compilation

transmissions11:
any blockers here? i much prefer multi-file verification

> I don't think this should be the default verification method right now

why so? 🤔

naszam:
Hi @transmissions11,

Thanks for your interest!

The only blocker so far is the `dapp-mk-standard-json` that doesn't work out-of-the-box for multi-file verification as it's used mainly in [dapp-build](https://github.com/dapphub/dapptools/blob/master/src/dapp/libexec/dapp/dapp-build#L66) to compile source code.
We tried to adjust it but we had a blocker with the remappings to get all the imported contract dependencies (cc: @WilfredTA).
The idea is to write a new script from scratch (probably in python) to generate a standard inputJSON for multi-file verification inspired by [truffle-plugin-verify](https://github.com/rkalis/truffle-plugin-verify/blob/master/verify.js#L215) or [hardhat-deploy](https://github.com/wighawag/hardhat-deploy/blob/master/src/etherscan.ts#L251) and then use it inside `dapp-verify-contracts` instead of `dapp-mk-standard-json`.
I think @e18r will give it a try, but feel free to jump on if you like!
Thanks!

transmissions11:
Ah interesting. The compilation stuff is a bit out of my league sadly. 

Thanks for all the work on this so far 🙏

Issue #619, opened 2021-03-04
Implement Berlin hard fork


MrChico:
https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/berlin.md

EIPS:
- [ ] [EIP-2565: ModExp Gas Cost](https://eips.ethereum.org/EIPS/eip-2565)
- [ ] [EIP-2315: Simple Subroutines for the EVM](https://eips.ethereum.org/EIPS/eip-2315)
There is some current back and forth regarding the specifics of whether jumps in and out of subroutines would be allowed after some feedback from the Solidity team. All core devs channel should be the best place to get the latest here.
- [ ] [EIP-2929: Gas cost increases for state access opcodes](https://eips.ethereum.org/EIPS/eip-2929)
- [ ] [EIP-2718: Typed Transaction Envelope](https://eips.ethereum.org/EIPS/eip-2718)
- [ ] [EIP-2930: Optional access lists](https://eips.ethereum.org/EIPS/eip-2930)
There will be an accompanying rpc call to generate the access list, WIP at: https://github.com/ethereum/go-ethereum/pull/22289. This should be used if available in `seth send` but can also be useful to optimize `seth run-tx`.

xwvvvvwx:
subroutines PR is here: https://github.com/dapphub/dapptools/pull/396

Probably just need rebasing.

MrChico:
subroutines is dropped now

PR #634, opened 2021-04-20
Berlin changes

laudiacay:
Draft PR for #619 ... Still in a bit of a messy state, but the bulk of the code's written, it just needs to be debugged & passing tests. 

Berlin tests that I am starting to work against as of right now are here: https://github.com/ethereum/tests/releases/tag/8.0.3


laudiacay:
I also ran linters against the codebase in 7349aa6 so it would stop crashing my editor- once this is passing tests, if y'all hate the linter output, I will just refactor to take that out.

xwvvvvwx:
@laudiacay super happy to see this PR, but the linter changes make this very hard to review, could you drop them from the PR please :pray:

src/hevm/src/EVM.hs
```diff
@@ -254,21 +256,24 @@ data FrameState = FrameState
...
-  , _substate        :: SubState
-  , _isCreate        :: Bool
-  , _txReversion     :: Map Addr Contract
+  { _gasprice            :: Word
+  , _txgaslimit          :: Word
+  , _origin              :: Addr
+  , _toAddr              :: Addr
+  , _value               :: SymWord
+  , _substate            :: SubState
+  , _accessList :: Map Addr [W256]
```
MrChico, line 265:
```suggestion
  , _accessList                     :: Map Addr [W256]
```

src/hevm/src/EVM.hs
```diff
@@ -2659,43 +2692,53 @@ mkCodeOps (SymbolicBuffer bytes) = RegularVector.fromList . toList $ go' 0 (stri
...
-    c_gascap =
-      if availableGas >= c_extra
-      then min xGas (allButOne64th (availableGas - c_extra))
-      else xGas
+  -> Bool -> Word -> Word -> Word -> Addr
+  -> EVM (Integer, Integer)
+costOfCall FeeSchedule {..} recipientExists xValue availableGas' xGas' target = do
+  acc <- accessAccountForGas target
+  let call_base_gas = if acc then g_warm_storage_read else g_cold_account_access
+  let availableGas = num availableGas'
```
MrChico, line 2700: I would not repeat the `let` for every line here

src/hevm/src/EVM/ABI.hs
```diff
@@ -24,93 +25,88 @@
...
 import qualified Text.Megaparsec.Char as P
+import Text.ParserCombinators.ReadP
 
 data AbiValue
-  = AbiUInt         Int Word256
-  | AbiInt          Int Int256
-  | AbiAddress      Addr
-  | AbiBool         Bool
-  | AbiBytes        Int BS.ByteString
+  = AbiUInt Int Word256
```
MrChico, line 86: I don't think we should let the linter dictate code alignment

MrChico:
Agree that dropping the linter changes make this a lot nicer. Some of the changes suggested by the linter are nice, like dropping redundant parens and simplifying expressions, but the I don't think we should adhere to its opinion on code alignment

src/hevm/src/EVM/Transaction.hs
```diff
@@ -143,10 +202,13 @@ initTx vm = let
     touched = if creation
               then [origin]
               else [origin, toAddr]
-    accessed = fromList $ [origin, toAddr] ++ [1..9]
+    accesslist = view (EVM.tx . EVM.accessList) vm
```
MrChico, line 205: Wouldn't it be more straightforward to initialize accessAddrs and accessedStoragekeys in `makeVm` rather than here?

xwvvvvwx:
Would be happy to review the linter stuff in a seperate pr fwiw

src/hevm/src/EVM/Transaction.hs
```diff
@@ -23,38 +23,60 @@ import qualified Data.Aeson.Types  as JSON
...
     txGasPrice :: W256,
     txNonce    :: W256,
     txR        :: W256,
     txS        :: W256,
     txToAddr   :: Maybe Addr,
     txV        :: W256,
-    txValue    :: W256
-  } deriving Show
+    txValue    :: W256,
+    txType     :: Maybe W256,
```
MrChico, line 41: could also make this an enumeration of known txTypes, it might make things simpler since we wouldn't have to deal with these extra `Maybe`s

src/hevm/src/EVM/Transaction.hs
```diff
@@ -23,38 +23,60 @@ import qualified Data.Aeson.Types  as JSON
...
+    txAccessList :: Maybe [AccessListEntry]
```
MrChico, line 42: can we treat the `Nothing` case as just the empty list?

src/hevm/src/EVM/Transaction.hs
```diff
@@ -8,53 +8,75 @@ import EVM.FeeSchedule
...
+    _ -> Nothing
   where v          = fromIntegral (txV tx)
         to'        = case txToAddr tx of
           Just a  -> BS $ word160Bytes a
           Nothing -> BS mempty
+        accessList = (maybe [] id) . txAccessList $ tx
+        rlpAccessList = BS.concat $ map (\accessEntry ->
+          rlpList [BS $ word160Bytes (accessAddress accessEntry), 
+          BS $ rlpList $ map rlpWord256 $ accessStorageKeys accessEntry
+          ]) accessList
```
MrChico, line 79:
```suggestion
        rlpAccessList = rlpList $ map (\accessEntry ->
          BS $ rlpList [BS $ word160Bytes (accessAddress accessEntry)
                     ,BS $ rlpList $ map rlpWord256 $ accessStorageKeys accessEntry]
          ) accessList
```

MrChico: needs confirmation, but I think something like this would make more sense to get the rlp encoding right

src/hevm/src/EVM/ABI.hs
laudiacay: undone 👯

src/hevm/src/EVM.hs
laudiacay: wow was not aware you could do that in haskell.... OCaml doesn't have that.

src/hevm/src/EVM/Transaction.hs
laudiacay: done!

laudiacay:
> @laudiacay super happy to see this PR, but the linter changes make this very hard to review, could you drop them from the PR please 🙏

done, yeah, i'll PR them elsewhere later with something that removes warnings

src/hevm/src/EVM/Transaction.hs
laudiacay: I think this is still wrong, I'm still getting a sender hash that doesn't exist in the prestate of the test. I'll mess with it and see what happens.

PR #635, opened 2021-04-20
nix: update hashes for solc static binaries

xwvvvvwx:
Seems as though the darwin binaries <=0.6.0 were updated in January due to https://github.com/ethereum/solidity/issues/10183.

This regenerates the derivations for the darwin solc binaries and adds new derivations for the most recent solc releases.

@MrChico can you try out a few solc versions from this branch on your mac to make sure everything is OK here?


MrChico:
This seems to work! Tested 0_5_16 and 0_4_12

MrChico (approved):

PR #637, opened 2021-04-30
Staticbuilds

MrChico:
adds a ci job that creates static binaries for hevm for macos and linux. The scripts are mainly adapted from echidna and uses stack for building. There's a mysterious failure in the tests that use cvc4 which I cannot reproduce locally (it's this old error returning https://github.com/LeventErkok/sbv/issues/510). Having started down this rabbit hole we should probably go all the way and bundle `dapp` and `seth` as well


xwvvvvwx:
Really happy to see this, but wondering why you copied the echidna scripts instead of using [static-haskell-nix](https://github.com/nh2/static-haskell-nix)?

xwvvvvwx:
Regarding dapp / seth, I would love to be able to distribute a static binary, but I'm not really sure how to go about it, perhaps [nix-bundle](https://github.com/matthewbauer/nix-bundle) is worth trying?

MrChico:
> Really happy to see this, but wondering why you copied the echidna scripts instead of using [static-haskell-nix](https://github.com/nh2/static-haskell-nix)?

well, I tried the nix way but this ended up being a lot simpler. A lot of the static-haskell-nix things also doesn't seem to work for macos

MrChico:
update:
Linux builds are now static by default, and come in two flavors; wrapped and unwrapped. Wrapped (the default `hevm` attribute) has z3, cvc4 and solc and everything coming from nix, as always. The only difference is that it now has all of its libraries statically linked.
The unwrapped hevm invokes z3 cvc4 and solc from the PATH, and this binary is the one we upload to ci with each release.

Although in principle it might be possible to go use nix for the pseudo-static macos build (the goal would basically be to statically link everything but libSystem), I was not able to get this to work. Instead, macos binaries are shipped with the libraries it (dynamically) links to, using a stack build adapted from echidnas ci. The nix hevm build should be unchanged.

MrChico:
I've decided to leave builds for `dapp` and `seth` out of this PR. It would still be very nice to have, but mañana mañana

.github/scripts/build-macos-release.sh
```diff
@@ -0,0 +1,32 @@
...
+    BINARY="$1"
+    MATCH="$2"
+    NEW="$3"
+    OLD=$(otool -L "$BINARY" | grep "$MATCH" | awk '{print $1}')
+    install_name_tool -change "$OLD" "$NEW" "$BINARY"
+    cp -n "$OLD" "$(dirname "$BINARY")/$(basename "$NEW")" || true
+}
+
+
+BUILD="$(mktemp -d)/hevm"
```
asymmetric, line 21: This is more commonly done with:

```suggestion
BUILD="$(mktemp -t hevm.XXXX)"
```

.github/scripts/install-cvc4.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+# Adapted from https://github.com/commercialhaskell/stack
```
asymmetric, line 3: Does this file need to mention the above repo's license, and/or include it somehow? Guess it depends how much of a derivative work this is.

haskell.nix
```diff
@@ -16,23 +16,26 @@ in self-hs: super-hs:
...
     restless-git = dontCheck "restless-git" (./src/restless-git);
     wreq = pkgs.haskell.lib.doJailbreak super-hs.wreq;
 
-    # we use a pretty bleeding edge sbv version
     sbv = sbv_prepatch.overrideAttrs (attrs: {
-      postPatch = ''
-      sed -i -e 's|"z3"|"${pkgs.z3}/bin/z3"|' Data/SBV/Provers/Z3.hs
-      sed -i -e 's|"cvc4"|"${pkgs.cvc4}/bin/cvc4"|' Data/SBV/Provers/CVC4.hs'';
+      postPatch =
+        if wrapped
```
asymmetric, line 29: You could use `lib.optionalString wrapped ...` here instead. Cf https://nixos.org/manual/nixpkgs/stable/#function-library-lib.strings.optionalString

haskell.nix
```diff
@@ -58,16 +61,29 @@ in self-hs: super-hs:
...
 
       enableSeparateDataOutput = true;
-      buildInputs = attrs.buildInputs ++ [pkgs.solc];
+      buildInputs = attrs.buildInputs ++ [pkgs.solc] ++ (if wrapped then [] else [pkgs.z3 pkgs.cvc4]);
       nativeBuildInputs = attrs.nativeBuildInputs ++ [pkgs.makeWrapper];
       configureFlags = attrs.configureFlags ++ [
           "--ghc-option=-O2"
-          ];
+      ] ++
+      (if stdenv.isDarwin then [] else [
```
asymmetric, line 79: Here you could use `stdenv.lib.optional !stdenv.isDarwin ...`

asymmetric (commented):
I don't have time to review or test this properly, so I've just left a couple of stylistic nits.

Additionally, will leave these 2 potentially relevant links, in case you haven't seen them:

- https://github.com/NixOS/nixpkgs/issues/43795
- https://github.com/nh2/static-haskell-nix

MrChico:
> Additionally, will leave these 2 potentially relevant links, in case you haven't seen them:

Yes, these were very helpful

.github/scripts/install-cvc4.sh
MrChico: it's a very simple script I doubt attribution is necessary

MrChico, line 3:
```suggestion
```

.github/scripts/install-solc.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+
+# Adapted from https://github.com/commercialhaskell/stack
```
MrChico, line 3:
```suggestion
```

.github/scripts/install-z3.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+# Adapted from https://github.com/commercialhaskell/stack
```
MrChico, line 3:
```suggestion
```

haskell.nix
```diff
@@ -58,16 +61,29 @@ in self-hs: super-hs:
...
+          "--enable-executable-static"
+          "--extra-lib-dirs=${pkgs.gmp.override { withStatic = true; }}/lib"
+          "--extra-lib-dirs=${pkgs.glibc.static}/lib"
+          "--extra-lib-dirs=${pkgs.libff}/lib"
+          "--extra-lib-dirs=${pkgs.ncurses.override {enableStatic = true; }}/lib"
+          "--extra-lib-dirs=${pkgs.zlib.static}/lib"
+          "--extra-lib-dirs=${pkgs.libffi.overrideAttrs (old: { dontDisableStatic = true; })}/lib"
+      ]);
```
MrChico, line 87:
```suggestion
      (stdenv.lib.lists.optionals !stdenv.isDarwin [
          "--enable-executable-static"
          "--extra-lib-dirs=${pkgs.gmp.override { withStatic = true; }}/lib"
          "--extra-lib-dirs=${pkgs.glibc.static}/lib"
          "--extra-lib-dirs=${pkgs.libff}/lib"
          "--extra-lib-dirs=${pkgs.ncurses.override {enableStatic = true; }}/lib"
          "--extra-lib-dirs=${pkgs.zlib.static}/lib"
          "--extra-lib-dirs=${pkgs.libffi.overrideAttrs (old: { dontDisableStatic = true; })}/lib"
      ]);
```

MrChico: hmm, I think I prefer the way its now

haskell.nix
```diff
@@ -16,23 +16,26 @@ in self-hs: super-hs:
...
-      sed -i -e 's|"z3"|"${pkgs.z3}/bin/z3"|' Data/SBV/Provers/Z3.hs
-      sed -i -e 's|"cvc4"|"${pkgs.cvc4}/bin/cvc4"|' Data/SBV/Provers/CVC4.hs'';
+      postPatch =
+        if wrapped
+        then
+          ''
+             sed -i -e 's|"z3"|"${pkgs.z3}/bin/z3"|' Data/SBV/Provers/Z3.hs
+             sed -i -e 's|"cvc4"|"${pkgs.cvc4}/bin/cvc4"|' Data/SBV/Provers/CVC4.hs
+          ''
+        else "";
```
MrChico, line 35:
```suggestion
      postPatch = stdenv.lib.optionalString wrapped
          ''
             sed -i -e 's|"z3"|"${pkgs.z3}/bin/z3"|' Data/SBV/Provers/Z3.hs
             sed -i -e 's|"cvc4"|"${pkgs.cvc4}/bin/cvc4"|' Data/SBV/Provers/CVC4.hs
          '';
```

MrChico:
Thanks @asymmetric

.github/scripts/install-cvc4.sh
```diff
@@ -0,0 +1,33 @@
...
+fetch_cvc4_macos() {
+  VER="$1"
+  wget "https://github.com/CVC4/CVC4/releases/download/$VER/cvc4-$VER-macos-opt"
+  chmod +x "cvc4-$VER-macos-opt"
+  mv "cvc4-$VER-macos-opt" "$HOME/.local/bin/cvc4"
+  echo "Downloaded cvc4 $VER"
+}
+
+if [ "$HOST_OS" = "Linux" ]; then
+    travis_retry fetch_cvc4_linux "1.8"
```
xwvvvvwx, line 30: I'm a little nervous that we hardcode the cvc4 version here, is there a way that we can make sure that the cvc4 version we install here is the same as the cvc4 version we get with the wrapped hevm via nix?

.github/scripts/install-solc.sh
```diff
@@ -0,0 +1,58 @@
...
+    wget "https://github.com/ethereum/solidity/releases/download/v$VER/solc-macos"
+    chmod +x solc-macos
+    mv solc-macos "$HOME/.local/bin/solc-$VER"
+    echo "Downloaded solc $VER"
+  else
+    echo "Skipped solc $VER, already present"
+  fi
+}
+
+fetch_all_solc_linux() {
```
xwvvvvwx, line 39: this seems to be unused?

.github/scripts/install-solc.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+
+
+set -eux
+
+mkdir -p $HOME/.local/bin;
+
+travis_retry() {
+  cmd=$*
+  $cmd || (sleep 2 && $cmd) || (sleep 10 && $cmd)
```
xwvvvvwx, line 10: why do we need this?

.github/workflows/release.yml
```diff
@@ -8,6 +8,18 @@ jobs:
     name: Create Release
     runs-on: ubuntu-latest
     steps:
+      # v12
+      - uses: cachix/install-nix-action@v12
+        with:
+          nix_path: nixpkgs=channel:nixos-unstable
```
xwvvvvwx, line 14: why do we need this?

xwvvvvwx (commented):
This looks generally good, I'm just worried about keeping everything in sync between the macos build scripts and the nix based builds. Is there a way that we can automatically select the libff / cvc4 etc. versions during the macos static build?

.github/workflows/release.yml
MrChico: because we're installing with nix-build the next line

.github/scripts/install-solc.sh
MrChico: removed

PR #642, opened 2021-05-19
identify code modulo `immutable` and fix stdjson bundling from etherscan

MrChico:
This PR lets us have srcmaps for contracts with immutable variables. If the hash based identification of a contract fails, we look at all the known contracts with immutable vars and see if our current bytecode matches any of them if we overwrite possible mutability locations with zeros.

There's also a fix for `seth bundle-source` to write all relevant contracts to disk when getting dealing with std-json artifacts from etherscan (i.e. multiple srcs, not flattened).

### DIFF
--- a/src/hevm/src/EVM/Solidity.hs
+++ b/src/hevm/src/EVM/Solidity.hs
@@ -16,6 +16,7 @@ module EVM.Solidity
   , CodeType (..)
   , Method (..)
   , SlotType (..)
+  , Reference(..)
   , methodName
   , methodSignature
   , methodInputs
@@ -59,7 +60,8 @@ import Data.SBV
 import Control.Applicative
 import Control.Monad
 import Control.Lens         hiding (Indexed, (.=))
-import Data.Aeson           (Value (..), ToJSON(..), (.=), object, encode)
+import Data.Aeson hiding (json)
+import Data.Aeson.Types
 import Data.Aeson.Lens
 import Data.Scientific
 import Data.ByteString      (ByteString)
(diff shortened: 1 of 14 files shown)


MrChico:
closes #593

src/hevm/src/EVM/Solidity.hs
```diff
@@ -106,6 +107,11 @@ data SlotType
 --  | StorageArray AbiType
   deriving Eq
 
+data Code = Code
```
xwvvvvwx, line 110: Is it possible to unify the `Code` types between here and in `Dapp.hs`? Seems a little weird to have these two almost identical types hanging around.

src/seth/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## unreleased
+
+### Fixed
+
+- seth bundle-source writes the contents of standard-json to the current directory to enable better sourcemaps for multiple files version of etherscan source.
```
xwvvvvwx, line 11:
```suggestion
- seth bundle-source writes the contents of standard-json to the current directory to enable better sourcemaps for multi-file etherscan source code.
```

src/hevm/src/EVM/Solidity.hs
MrChico: oops

xwvvvvwx (approved):

Issue #644, opened 2021-05-20
new test type: `invariantXXXX()`


MrChico:
randomly selects any possible public method in the set of contracts available after `setUp()` and calls it with randomly instantiates parameters. This process is repeated until `depth` is reached or an assertion is violated.

MrChico:
This is a less hands on approach than #485. Although the same thing could be synthesized in #485, it does not scale very well if multiple contracts are involved

MrChico:
introduced in #647

PR #647, opened 2021-05-20
ds-test invariants; testing functions run against arbitrary contract state

MrChico:
Concept art for #644. Seems to work pretty well except it doesn't have any replay functionality yet
EDIT; now implemented

### DIFF
--- a/src/dapp/CHANGELOG.md
+++ b/src/dapp/CHANGELOG.md
@@ -4,9 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## unreleased
+
+- A new type of test has been added invariant testing, 
+which run a test function against an arbitrary contract state. More information in the [README](./README.md###invariant-testing).
+
 ## [0.32.2] - 2021-03-22
 
-## Changed
+### Changed
 
 - updated `nixpkgs` to the `20.09` channel
 
--- a/src/dapp/CHANGELOG.md
+++ b/src/dapp/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## unreleased
 
 - A new type of test has been added invariant testing, 
-which run a test function against an arbitrary contract state. More information in the [README](./README.md###invariant-testing).
+which runs a test function against an arbitrary contract state. More information in the [README](./README.md###invariant-testing).
 
 ## [0.32.2] - 2021-03-22
 
--- a/src/dapp/CHANGELOG.md
+++ b/src/dapp/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## unreleased
 
 - A new type of test has been added invariant testing, 
-which runs a test function against an arbitrary contract state. More information in the [README](./README.md###invariant-testing).
+which runs a test function against a random reachable contract state. More information in the [README](./README.md###invariant-testing).
 
 ## [0.32.2] - 2021-03-22
 
--- a/src/dapp/README.md
+++ b/src/dapp/README.md
@@ -202,6 +202,8 @@ non-testing contract available after the `setUp()` function has been run, checki
 The `--depth` paramenter determines how many transactions deep each test will run, while the `--fuzz-runs` parameter 
 determines how many times the whole process is repeated.
 
+Note that a revert in any of the randomly generated call will not trigger a test failure. The goal of invariant tests is to find a state change that results in a violation of the assertions defined in the body of the test method, and since reverts do not result in a state change, they can be safely ignored. Reverts within the body of the `invariant*` test method will however still cause a test failure.
+
 Example:
 
 ```solidity
--- a/src/dapp/README.md
+++ b/src/dapp/README.md
@@ -16,6 +16,7 @@ that isn't available in `rpc`, such as [fuzz testing](#dapp-test-flags), symboli
   - [Unit testing](#unit-testing)
   - [Property based testing](#property-based-testing)
   - [Symbolically executed tests](#symbolically-executed-tests)
+  - [Invariant testing](#invariant-testing)
   - [Testing against RPC state](#testing-against-rpc-state)
   - [Deployment](#deployment)
 - [Configuration](#configuration)
--- a/src/dapp/README.md
+++ b/src/dapp/README.md
@@ -156,6 +156,8 @@ function test_withdraw(uint96 amount) public {
 }
 ```
 
+If a counterexample is found, it can be replayed or analyzed in the debugger using the `--replay` flag.
+
 ### Symbolically executed tests
 
 While property based testing runs each function repeatedly with new input values, symbolic execution leaves these 
@@ -188,6 +190,26 @@ Failure: proveFail_withdraw(uint256)
 which demonstrates that if we give the password `42`, it is possible to withdraw from the vault.
 
 For more reading on property based testing and symbolic execution, see [this tutorial on the Ethereum Foundation blog](https://fv.ethereum.org/2020/12/11/symbolic-execution-with-ds-test/).
+
+### Invariant testing
+
+While other forms of tests are always run against the post state of the `setUp()` function in the testing contract,
+it can be also be useful to check whether a property is satisfied at any possible contract state. This can be done with
+the `invariant*` testing type. When running an invariant test, hevm will invoke any state mutating function from any 
+non-testing contract available after the `setUp()` function has been run, checking the `invariant*` after each run.
+
+The `--depth` paramenter determines how many transactions deep each test will run, while the `--fuzz-runs` parameter 
+determines how many times the whole process is repeated.
+
+Example:
+
+```solidity
+function invariant_totalSupply() public {
+    assertEq(token.totalSupply(), initialTotalSupply);
+}
+```
+
+If a counterexample is found, it can be replayed or analyzed in the debugger using the `--replay` flag.
 
 ### Testing against RPC state
 
--- a/src/dapp/README.md
+++ b/src/dapp/README.md
@@ -195,7 +195,7 @@ For more reading on property based testing and symbolic execution, see [this tut
 ### Invariant testing
 
 While other forms of tests are always run against the post state of the `setUp()` function in the testing contract,
-it can be also be useful to check whether a property is satisfied at any possible contract state. This can be done with
+it can be also be useful to check whether a property is satisfied at every possible contract state. This can be done with
 the `invariant*` testing type. When running an invariant test, hevm will invoke any state mutating function from any 
 non-testing contract available after the `setUp()` function has been run, checking the `invariant*` after each run.
 
--- a/src/hevm/CHANGELOG.md
+++ b/src/hevm/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Added
 
 - Can now identify contracts with `immutable` by comparing bytecode modulo immutableReferences.
-- A new test runner for exploring arbitrary contract states.
+- A new test runner for checking invariants against random reachable contract states.
 
 ## 0.46.0 - 2021-04-29
 
--- a/src/hevm/CHANGELOG.md
+++ b/src/hevm/CHANGELOG.md
@@ -4,7 +4,8 @@
 
 ### Added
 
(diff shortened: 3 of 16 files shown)


MrChico:
This is now ready for review. Each tx uses an incremented timestamp and a random caller. Replay functionality is implemented. More info at https://github.com/dapphub/dapptools/tree/explore/src/dapp#invariant-testing

MrChico:
ci seems to fetch the wrong version of dss for some reason

src/hevm/src/EVM/UnitTest.hs
```diff
@@ -422,15 +444,122 @@ runTest opts@UnitTestOptions{..} vm (ConcreteTest testName, types) = liftIO $ ca
...
+ Stepper.evm $ assign (block . timestamp) (w256lit timestamp')
+ -- perform the call
+ bailed <- exploreStep opts' cd
+ Stepper.evm popTrace
+ let newHistory = if bailed then List history else List (thisCallRLP:history)
+     opts'' = opts {testParams = testParams {testTimestamp = timestamp'}}
+     carryOn = explorationStepper opts'' testName replayData newHistory (i - 1)
+ -- if we didn't revert, run the test function
+ if bailed
+ then carryOn
```
xwvvvvwx, line 522: Shouldn't we mark the test as failed if a call reverted? It's quite of a big semantic difference between the existing test types. If we're going to keep this behaviour perhaps we can highlight any reverts in the test output somehow?

src/hevm/src/EVM/Stepper.hs
```diff
@@ -57,6 +58,9 @@ data Action a where
   -- | Embed a VM state transformation
   EVM  :: EVM a   -> Action a
 
+  -- | Perform an IO action
+  IOAct :: StateT VM IO a -> Action a -- they should all just be this?
```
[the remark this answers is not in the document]
MrChico: its more direct, we don't have to define new `Query `s for every side effect we want to do

src/hevm/src/EVM/UnitTest.hs
```diff
@@ -422,15 +444,122 @@ runTest opts@UnitTestOptions{..} vm (ConcreteTest testName, types) = liftIO $ ca
...
     else fuzzRun opts vm testName types
 runTest opts vm (SymbolicTest testName, types) = symRun opts vm testName types
-
+runTest opts@UnitTestOptions{..} vm (ExploreTest testName, []) = liftIO $ case replay of
+  Nothing -> exploreRun opts vm testName []
+  Just (sig, cds) ->
+    if sig == testName
+    then exploreRun opts vm testName (decodeCalls cds)
+    else exploreRun opts vm testName []
+runTest _ _ (ExploreTest _, _) = error "TODO: support this too"
```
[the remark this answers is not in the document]
MrChico: yeah, would probably randomly instantiate the arguments that is used to call the invariant testing.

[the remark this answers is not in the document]
MrChico:
```suggestion
runTest _ _ (ExploreTest _, _) = error "invariant testing with arguments is not implemented (yet!)"
```

src/hevm/src/EVM/TTY.hs
```diff
@@ -274,6 +280,9 @@ isFuzzTest :: (Test, [AbiType]) -> Bool
 isFuzzTest (SymbolicTest _, _) = False
 isFuzzTest (ConcreteTest _, []) = False
 isFuzzTest (ConcreteTest _, _) = True
+isFuzzTest (ExploreTest _, []) = False
```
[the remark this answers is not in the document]
MrChico: oops

src/hevm/src/EVM/Stepper.hs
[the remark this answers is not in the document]
MrChico: yeah, that's what the comment suggests

src/hevm/src/EVM/UnitTest.hs
MrChico: keep in mind that this is the "exploration calls" i.e. calls we're making to attempt to change state in some valid way. If it reverts, then there's no state change for us to care about

src/hevm/src/EVM/UnitTest.hs
```diff
@@ -422,15 +444,122 @@ runTest opts@UnitTestOptions{..} vm (ConcreteTest testName, types) = liftIO $ ca
...
     else fuzzRun opts vm testName types
 runTest opts vm (SymbolicTest testName, types) = symRun opts vm testName types
-
+runTest opts@UnitTestOptions{..} vm (InvariantTest testName, []) = liftIO $ case replay of
+  Nothing -> exploreRun opts vm testName []
+  Just (sig, cds) ->
+    if sig == testName
+    then exploreRun opts vm testName (decodeCalls cds)
+    else exploreRun opts vm testName []
+runTest _ _ (InvariantTest _, _) = error "TODO: support this too"
```
MrChico, line 453:
```suggestion
runTest _ _ (InvariantTest _, _) = error "invariant testing with arguments is not implemented (yet!)"
```

src/dapp/CHANGELOG.md
```diff
@@ -4,9 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## unreleased
+
+- A new type of test has been added invariant testing, 
+which runs a test function against an arbitrary contract state. More information in the [README](./README.md###invariant-testing).
```
livnev, line 10: The expression "an arbitrary contract state" is used multiple times when describing this feature. To an unfamiliar reader, this could sound like "a specific arbitrary contract state", i.e. one that you get to specify, instead of an abstract/generic one. Maybe consider rephrasing (here and in other places) to replace arbitrary with "arbitrary symbolic" or just "generic"/"symbolic"? Also changing it to plural ("arbitrary contract states") could make this more clear.

src/hevm/CHANGELOG.md
```diff
@@ -4,7 +4,8 @@
 
 ### Added
 
-Can now identify contracts with `immutable` by comparing bytecode modulo immutableReferences
+- Can now identify contracts with `immutable` by comparing bytecode modulo immutableReferences.
+- A new test runner for exploring arbitrary contract states.
```
livnev, line 8: Also phrasing here

src/dapp/README.md
```diff
@@ -189,6 +192,26 @@ which demonstrates that if we give the password `42`, it is possible to withdraw
 
 For more reading on property based testing and symbolic execution, see [this tutorial on the Ethereum Foundation blog](https://fv.ethereum.org/2020/12/11/symbolic-execution-with-ds-test/).
 
+### Invariant testing
+
+While other forms of tests are always run against the post state of the `setUp()` function in the testing contract,
+it can be also be useful to check whether a property is satisfied at any possible contract state. This can be done with
```
livnev, line 198: s/any/every/ ?

src/hevm/src/EVM/UnitTest.hs
```diff
@@ -422,15 +444,122 @@ runTest opts@UnitTestOptions{..} vm (ConcreteTest testName, types) = liftIO $ ca
...
+           knownAbis :: Map Addr SolcContract
+           knownAbis =
+             -- exclude contracts without code
+             Map.filter (not . BS.null . view runtimeCode) $ 
+             -- exclude contracts without state changing functions
+             Map.filter (not . null . Map.filter mutable . view abiMap) $ 
+             -- exclude testing abis
+             Map.filter (isNothing . preview (abiMap . ix unitTestMarkerAbi)) $
+             -- pick all contracts with known compiler artifacts
+             fmap fromJust (Map.filter isJust $ Map.fromList [(addr, lookupCode (view contractcode c) dapp) | (addr, c)  <- Map.toList $ cs])
```
xwvvvvwx, line 486:
```suggestion
             fmap fromJust (Map.filter isJust $ Map.fromList [(addr, lookupCode (view contractcode c) dapp) | (addr, c)  <- Map.toList cs])
```

src/hevm/src/EVM/Stepper.hs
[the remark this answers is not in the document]
xwvvvvwx: I'm a little on the fence about this change. In a pre `IOAct` world we had pretty strict control over what kind of side effects can be triggered by EVM execution,  but with `IOAct` we allow unbounded side effects. While IOAct is definately simpler and more flexible, I do think that having tight control over side effects in the core is pretty nice. It also feels a little messy to leave the old `Query` types around side by side with the new `IOAct` approach.

I think for this PR all we really needed was a `pleaseGetRandom` query that calls `generate` on a `Gen` or something like that?

src/dapp/CHANGELOG.md
MrChico: it's not symbolic in this iteration, this is just a fuzzed thing. Maybe `random contract state` would be better

MrChico, line 10:
```suggestion
which runs a test function against a random reachable contract state. More information in the [README](./README.md###invariant-testing).
```

src/dapp/README.md
MrChico:
```suggestion
it can be also be useful to check whether a property is satisfied at every possible contract state. This can be done with
```

src/dapp/README.md
```diff
@@ -189,6 +192,26 @@ which demonstrates that if we give the password `42`, it is possible to withdraw
...
+the `invariant*` testing type. When running an invariant test, hevm will invoke any state mutating function from any 
+non-testing contract available after the `setUp()` function has been run, checking the `invariant*` after each run.
+
+The `--depth` paramenter determines how many transactions deep each test will run, while the `--fuzz-runs` parameter 
+determines how many times the whole process is repeated.
```
xwvvvvwx, line 203:
```suggestion
determines how many times the whole process is repeated.

Note that unlike the other test types, `invariant*` tests do not treat reverting calls as a failure. The goal of invariant tests is to find a state change that results in a violation of the assertions defined in the body of the test method, and since reverts do not result in a state change, they can be safely ignored.
```

src/dapp/CHANGELOG.md
livnev: Oh right, in that case yes adding "random" would help make this clear since I didn't understand that this was the case. 😅

src/hevm/CHANGELOG.md
MrChico, line 8:
```suggestion
- A new test runner for checking invariants against random reachable contract states.
```

src/hevm/src/EVM/Stepper.hs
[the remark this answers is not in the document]
xwvvvvwx: To be clear I'm not totally opposed to replacing `Query` with `IOAct`, I just think it's a moderately large shift in architectural philosophy, and I might prefer to make this change in a seperate PR where we also remove `Query` entirely.

src/dapp/README.md
MrChico: well, reverts in the testing function are still treated as failure. Just not any of the exploration calls

src/hevm/src/EVM/Stepper.hs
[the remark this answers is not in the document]
MrChico: Making a `GetRandom` `Query` here feels a little bureaucratic to me, since these calldata generation calls are kinda outside the realm of EVM execution, and more of in the domain of a "block execution" environment. Making this a Query would seem to result in storing all of the abi type information in the `result` of the VM as an intermediate failure, only to immediately resolve it with a fetcher. In the other cases, our `Query`s are at least something blocking the actual execution, so I think it doesn't feel too wild to allow `IOAct` here. But I'm down to experiment with removing `Query` entirely in a future pr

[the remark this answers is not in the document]
xwvvvvwx: yeah ok it's a fair point re: the beurocracy. I'm fine with this for now.

src/dapp/README.md
xwvvvvwx: ah true, what about:

```suggestion
determines how many times the whole process is repeated.

Note that a revert in any of the randomly generated call will not trigger a test failure. The goal of invariant tests is to find a state change that results in a violation of the assertions defined in the body of the test method, and since reverts do not result in a state change, they can be safely ignored. Reverts within the body of the `invariant*` test method will however still cause a test failure.
```

MrChico: nice, thx. Added

xwvvvvwx (approved):
lgtm :sparkling_heart:

MrChico:
thanks for review @xwvvvvwx

PR #653, opened 2021-06-01
dapp: rm mismatched hashes error

d-xo:
Is this bad? What was this check even for? I don't understand this file at all. halp pls :pray:


MrChico:
lets make it a warning instead

d-xo:
https://github.com/dapphub/dapptools/pull/677

PR #655, opened 2021-06-01
Update hevm deps and nixpkgs to 21.05

arcz:
Update some deps, especially `base16-bytestring` since its API changed and broke hevm in nixpkgs which I unbroke a while ago. Not sure if it's the problem on my end but I had to fallback to static solc in `overlay.nix` because none of versions are building on macOS.


MrChico:
Thanks! Didn't know there was a new release. Triggered ci

xwvvvvwx:
looks like `dapp` builds are failing due to some new shellcheck features...

arcz:
Should be fixed now

arcz:
@MrChico Not sure what happened to the ubuntu build, looks like a race condition?

xwvvvvwx:
reran CI and it's still failing so I guess it's not a race?

arcz:
yeah, I looked at the wrong place. I'm investigating this

xwvvvvwx:
wondering if bumping the sleep 90s to like 10 mins or something might help?

Perhaps nix needs to build a new patched geth version since we've bumped nixpkgs and that's taking more than 90s?

arcz:
bumped to 10 mins, let's see

xwvvvvwx:
It's passing!

Big thanks for all the work @arcz, very much appreciated 💖

PR #656, opened 2021-06-02
WIP: FFI interface calling arbitrary scripts as hevm cheat code

mhhf:


src/hevm/src/EVM.hs
```diff
@@ -27,6 +30,8 @@ import EVM.FeeSchedule (FeeSchedule (..))
 import Options.Generic as Options
 import qualified EVM.Precompiled
 
+import Debug.Trace
+
```
MrChico, line 34:
```suggestion
```

MrChico:
with a clean up and a flag `--ffi` this would be great to merge

src/hevm/src/EVM.hs
```diff
@@ -1951,7 +1961,30 @@ type CheatAction = Word -> Word -> Buffer -> EVM ()
...
+                      let encoded = ConcreteBuffer bs
+                      assign (state . returndata) encoded
+                      copyBytesToMemory encoded outSize 0 outOffset
+                      assign result Nothing
+                  in assign result (Just . VMFailure . Query $ (PleaseDoFFI cmd cont))
+                _ -> vmError (BadCheatCode sig)
+              _ -> vmError (BadCheatCode sig)
+          else
+            let msg = "ffi disabled: run again with --ffi if you want to allow tests to call external scripts"
+            in vmError . Revert $ abiMethod "Error(string)" (AbiString msg),
```
d-xo, line 1985: I can't seem to get this message to pretty print in the trace, not sure what I'm doing wrong :/

src/hevm/src/EVM.hs
```diff
@@ -110,6 +114,7 @@ data VM = VM
   , _burned         :: Word
   , _constraints    :: [(SBool, Whiff)]
   , _iterations     :: Map CodeLocation Int
+  , _options        :: VMOpts
```
d-xo, line 117: I ended up adding the options struct to `VM`, since I needed to check at runtime whether or not `allowFFI` was set. Not sure if there's a cleaner way?

MrChico: hmm, well, we don't need the entire thing. Why don't we just have `ffiEnabled :: Bool` instead?

src/hevm/src/EVM.hs
MrChico: cheat codes aren't handled exactly like a call is, there is no new frame allocated and we don't call `finishFrame` when returning. The revert message is added to the trace in the `finalize` subcall of `finishFrame`

src/hevm/src/EVM.hs
d-xo: yes true. we can always generalise to a `runtimeOpts` or similar if we need to later.

src/hevm/src/EVM.hs
d-xo: no I mean the revert is shown in the trace, but it's just a bunch of hex data instead of being pretty printed as a string.

I see this:

```
Failure: testFFI
  src/broked.sol:CheatCodes
   ├╴constructor
   └╴testFFI()
      └╴error Revert (0x00000000000000000000000000000000000000000000000000000000000000566666692064697361626c65643a2072756e20616761696e2077697468202d2d66666920696620796f752077616e7420746f20616c6c6f7720746573747320746f2063616c6c2065787465726e616c207363726970747300000000000000000000) (src/broked.sol:24)
```

when I want to see this:

```
Failure: testFFI
  src/broked.sol:CheatCodes
   ├╴constructor
   └╴testFFI()
      └╴error Revert ("ffi disabled: run again with --ffi if you want to allow tests to call external scripts") (src/broked.sol:24)
```

d-xo: I tried making an explicit call to `finishFrame` as below, but I still get the same output:

```haskell
            let msg = "ffi disabled: run again with --ffi if you want to allow tests to call external scripts"
            in finishFrame . FrameReverted . ConcreteBuffer $ abiMethod "Error(string)" (AbiString . encodeUtf8 $ msg),
```

d-xo: The message should have been encoded as a string in a tuple, not just a string.

d-xo:
> with a clean up and a flag `--ffi` this would be great to merge

Done, I think this is ready to merge :slightly_smiling_face:

src/hevm/src/EVM.hs
```diff
@@ -1951,7 +1962,30 @@ type CheatAction = Word -> Word -> Buffer -> EVM ()
...
+          vm <- get
+          if view EVM.allowFFI vm then
+            case decodeBuffer [AbiArrayDynamicType AbiStringType] input of
+              CAbi valsArr -> case valsArr of
+                [AbiArrayDynamic AbiStringType strsV] ->
+                  let
+                    cmd = (flip fmap) (V.toList strsV) (\case
+                      (AbiString a) -> unpack $ decodeUtf8 a
+                      _ -> "")
+                    cont = \bs -> do
```
MrChico, line 1976:
```suggestion
                    cont bs = do
```

src/hevm/src/EVM/ABI.hs
```diff
@@ -250,6 +250,7 @@ putAbi = \case
   AbiTuple v ->
     putAbiSeq v
 
+-- |Decode a sequence type (e.g. tuple / array). Will fail for non sequence types
```
MrChico, line 253:
```suggestion
-- | Decode a sequence type (e.g. tuple / array). Will fail for non sequence types
```

MrChico:
Please mention this in https://github.com/dapphub/dapptools/tree/master/src/hevm#cheat-codes and in the changelog and this is good to go!

d-xo:
> Please mention this in https://github.com/dapphub/dapptools/tree/master/src/hevm#cheat-codes and in the changelog and this is good to go!

was already done :slightly_smiling_face:

MrChico (approved):

MrChico:
> > Please mention this in https://github.com/dapphub/dapptools/tree/master/src/hevm#cheat-codes and in the changelog and this is good to go!
> 
> was already done 🙂

oops, yes. Great

PR #657, opened 2021-06-03
fix: fix the wrong parsing for 'vv' and 'vvv'

guanqun:


TransmissionsDev:
If this gets merged ideally some references to `--vvv` get added to the docs as right now it is not mentioned at all.

d-xo:
The issue here seems to be that `git rev-parse` doesn't accept short opts longer than one char so `-vv` and `-vvv` are both parsed as `-v`.

The PR here fixes this by changing `-vv` to `--vv` which tbh feels a little unsatisfying.

Not totally sure what the best way forward here is tbh. Seems like a really clean solution might involve rewriting our options parsing to use something other than `git rev-parse`, which would be rather annoying.

TransmissionsDev:
> The PR here fixes this by changing -vv to --vv which tbh feels a little unsatisfying.

Yeah fair. idk personally i think just flat out removing these and keeping things explicit with `--verbosity` is fine :p

d-xo:
I think I agree with that. Closing this in favour of: https://github.com/dapphub/dapptools/pull/676.

@guanqun apologies for the delayed response here, and big thanks so much for flagging the issue and submitting a PR! :pray: :sparkles:

guanqun:
nvm, it's glad to see there's a better fix.

PR #661, opened 2021-06-10
invariant testing fixes

MrChico:
Fixes (mostly) a performance regression in dapp debug introduced in #593 and introduces the ability to select which contracts to chose methods from during invariant testing. If a function `targetContracts()(address[])` is present in the current testing contract. See some examples in the testing files

### DIFF
--- a/src/dapp-tests/fail/invariantFail.sol
+++ b/src/dapp-tests/fail/invariantFail.sol
@@ -24,3 +24,55 @@ contract TestdappTest is DSTest {
         assertLt(testdapp.x(), 100);
     }
 }
+
+contract BrokenCoin {
+  mapping(address=>uint) public balanceOf;
+  constructor(uint amount) public {
+    balanceOf[msg.sender] = amount;
+  }
+
+  function transfer(address to, uint amount) public {
+    uint senderBal = balanceOf[msg.sender];
+    uint toBal = balanceOf[to];
+    require(senderBal >= amount);
+    senderBal += amount;
+    toBal -= amount;
+    balanceOf[msg.sender] = senderBal;
+    balanceOf[to] = toBal;
+  }
+}
+
+contract InvariantTest is DSTest {
+    BrokenCoin token;
+    User user;
+    address[] targetContracts_;
+
+    function targetContracts() public returns (address[] memory) {
+      return targetContracts_;
+    }
+    function setUp() public {
+        token = new BrokenCoin(100 ether);
+        user = new User(token);
+        token.transfer(address(user), 100 ether);
+        targetContracts_.push(address(user));
+    }
+
+    function invariantTestUserBal() public {
+        assertLe(token.balanceOf(address(user)), 100 ether);
+    }
+}
+
+contract User {
+  BrokenCoin token;
+  constructor(BrokenCoin token_) public {
+    token = token_;
+  }
+
+  function doTransfer(address to, uint amount) public {
+    token.transfer(to, amount);
+  }
+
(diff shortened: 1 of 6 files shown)


src/hevm/src/EVM/UnitTest.hs
```diff
@@ -222,12 +222,18 @@ tick x = Text.putStr x >> hFlush stdout
...
 -- | This is like an unresolved source mapping.
 data OpLocation = OpLocation
-  { srcCode :: ContractCode
+  { srcContract :: Contract
   , srcOpIx :: Int
-  } deriving (Show, Eq, Ord)
+  } deriving (Show)
+
+instance Eq OpLocation where
+  (==) (OpLocation a b) (OpLocation a' b') = b == b' && view contractcode a == view contractcode a'
```
xwvvvvwx, line 230: Aren't these instances a little weird? A `Contract` contains a lot more than just it's code...

src/dapp-tests/fail/invariantFail.sol
```diff
@@ -24,3 +24,55 @@ contract TestdappTest is DSTest {
...
+    toBal -= amount;
+    balanceOf[msg.sender] = senderBal;
+    balanceOf[to] = toBal;
+  }
+}
+
+contract InvariantTest is DSTest {
+    BrokenCoin token;
+    User user;
+    address[] targetContracts_;
```
xwvvvvwx, line 48: why not `address[] public targetContracts`?

src/hevm/src/EVM/UnitTest.hs
```diff
@@ -525,22 +534,40 @@ explorationStepper opts@UnitTestOptions{..} testName replayData (List history) i
...
+explorationStepper _ _ _ _ _ _  = error "malformed rlp"
+
+getTargetContracts :: UnitTestOptions -> Stepper [Addr]
+getTargetContracts UnitTestOptions{..} = do
+  vm <- Stepper.evm $ get
+  let Just contract = currentContract vm
+      theAbi = view abiMap $ fromJust $ lookupCode (view contractcode contract) dapp
+      setUp  = abiKeccak (encodeUtf8 "targetContracts()")
+  case Map.lookup setUp theAbi of
+    Nothing -> error "no method targetContracts"--return []
```
xwvvvvwx, line 546:
```suggestion
    Nothing -> return []
```

src/hevm/src/EVM/UnitTest.hs
MrChico: well, we only really care about the code here. Previous `OpLocation` was just code and index. We expand it here to be able to better identify contracts, but as far as the insertion to the multiset goes, only the code is relevant

src/dapp-tests/fail/invariantFail.sol
MrChico: because that makes a function `targetContracts(uint)(address)`

TransmissionsDev: did not know that trick!

src/hevm/src/EVM/UnitTest.hs
xwvvvvwx: I guess the only thing I'm worried about is that at some point in the future someone will use these instances for something else, and encounter some strange and hard to debug issue as a result.

I think I'm probably OK with this for now in the end though, all the additional checks that we would need for a full `Eq` instance seem like they would introduce quite a lot of overhead....

xwvvvvwx (approved):

PR #663, opened 2021-06-10
Fixup solc linux

xwvvvvwx:
Adds `solc-0.8.4` and uses our copy of z3 for linux builds only (since the approach here was broken on macos).

@MrChico @kjekac could one of you try running `nix-env -f https://github.com/dapphub/dapptools/archive/d194301f1c81ff8819c95823cebda68d771d283a.tar.gz -iA solc-static-versions.solc_0_8_4` before merge just to make sure I didn't mess anything up with the macos builds.


MrChico:
works for me

PR #676, opened 2021-06-26
rm mentions of -vv (was broken)

d-xo:
Purges all mention of the broken `-vv` flag.

see: https://github.com/dapphub/dapptools/pull/657

### DIFF
--- a/src/dapp/README.md
+++ b/src/dapp/README.md
@@ -196,7 +196,7 @@ For more reading on property based testing and symbolic execution, see [this tut
 
 While other forms of tests are always run against the post state of the `setUp()` function in the testing contract,
 it can be also be useful to check whether a property is satisfied at every possible contract state. This can be done with
-the `invariant*` testing type. When running an invariant test, hevm will invoke any state mutating function from any
+the `invariant*` testing type. When running an invariant test, hevm will invoke any state mutating function from all addresses returned by a call to `targetContracts()`, if such a function exists in the testing contracts. If no such method exists, it will invoke methods from any
 non-testing contract available after the `setUp()` function has been run, checking the `invariant*` after each run.
 
 The `--depth` paramenter determines how many transactions deep each test will run, while the `--fuzz-runs` parameter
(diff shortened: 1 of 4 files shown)


src/dapp/README.md
```diff
@@ -196,10 +196,10 @@ For more reading on property based testing and symbolic execution, see [this tut
 
 While other forms of tests are always run against the post state of the `setUp()` function in the testing contract,
 it can be also be useful to check whether a property is satisfied at every possible contract state. This can be done with
-the `invariant*` testing type. When running an invariant test, hevm will invoke any state mutating function from any 
+the `invariant*` testing type. When running an invariant test, hevm will invoke any state mutating function from any
```
MrChico, line 199:
```suggestion
the `invariant*` testing type. When running an invariant test, hevm will invoke any state mutating function from all addresses returned by a call to `targetContracts()`, if such a function exists in the testing contracts. If no such method exists, it will invoke methods from any
```

MrChico: Sneaking this change in here :)

MrChico (approved):

TransmissionsDev:
fyi there still seems to be mentions of -vv and -vvv in the latest `dapp` help menu:

```
❯ dapp
Usage: dapp <command> [<args>]
   or: dapp help <command>

Build, test, deploy and transact with Ethereum contracts from the comfort of the command line.

dapp <command> --help
--
  Compilation options:
optimize             activate solidity optimizer
legacy               compile using the '--combined-json' format
extract              after building, write the .abi, .bin and .bin-runtime. Implies '--legacy'

  Testing options:
verbosity=<number>   verbosity of 'dapp test' output (0-3)
v,verbose            sets verbosity to 1
vv                   sets verbosity to 2
vvv                  sets verbosity to 3
fuzz-runs=<number>   number of times to run fuzzing tests
replay=<string>      rerun a particular test case
m,match=<string>     only run test methods matching regex
cache=<string>       use the cache at directory
ffi                  allow the use of the ffi cheatcode (WARNING: allows test authors to execute arbitrary code on your machine)

  RPC options:
  
  ```

PR #677, opened 2021-06-28
Loosen remappings warning

d-xo:
This downgrades the dreaded mismatched packages error to a warning, and removes the useless `index.sol` remappings from the `dapp remappings` output.

### DIFF
--- a/src/dapp/libexec/dapp/dapp-remappings
+++ b/src/dapp/libexec/dapp/dapp-remappings
@@ -53,7 +53,7 @@ function findRemappings(prefix) {
 
 function ls(dir) {
   try {
-    return require("fs").readdirSync(dir)
+    return require("fs").readdirSync(dir).sort()
   } catch (error) {
     return []
   }
--- a/src/dapp/libexec/dapp/dapp-remappings
+++ b/src/dapp/libexec/dapp/dapp-remappings
@@ -11,7 +11,6 @@ Object.keys(pkg_src).sort((a, b) => a.length - b.length).sort(
   }
 ).forEach(name => {
   console.log(`${name}/=${pkg_src[name]}/`)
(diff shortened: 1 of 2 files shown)


src/dapp/libexec/dapp/dapp-remappings
```diff
@@ -31,23 +30,23 @@ function findRemappings(prefix) {
...
     if (pkg_src[name]) {
       if (hash != pkg_hash[name]) {
-        console.error(`${PROGRAM_NAME}: error: mismatching packages:`)
-        console.error(`${PROGRAM_NAME}: error: (1) ${pkg_src[name]}`)
-        console.error(`${PROGRAM_NAME}: error: (2) ${src}`)
-        process.exit(1)
+        console.error(`${PROGRAM_NAME}: warning: mismatching packages:`)
+        console.error(`${PROGRAM_NAME}: warning: (1) ${pkg_src[name]}`)
+        console.error(`${PROGRAM_NAME}: warning: (2) ${src}`)
+        console.error(`${PROGRAM_NAME}: warning: using ${pkg_src[name]}. You can override this using \`DAPP_REMAPPINGS\``)
```
d-xo, line 41: Is this always going to be the across different filesystems / invocations? Bit nervous that we're introducing some non determinism into the build process here....

MrChico: hmm, yeah, not sure

d-xo: added a sort to `ls()`, which I think should be enough to ensure we always have a consistent choice in case of conflicts?

MrChico: nice, sounds good to me!

MrChico (approved):

PR #679, opened 2021-07-01
nix: add shared haskell package set

d-xo:
act would just instantly segfault when built against static hevm, with this it works :)


nix/libff.nix
```diff
@@ -16,9 +16,9 @@ stdenv.mkDerivation rec {
 
   # CMake is hardcoded to always build static library which causes linker
   # failure for Haskell applications depending on haskellPackages.hevm on macOS.
-  postPatch = lib.optionalString stdenv.isDarwin ''
+  postPatch = if (!enableStatic) then ''
```
d-xo, line 19: this also defaults `libff` to shared on linux, which fixed `cabal v2-repl` for me.

PR #685, opened 2021-07-02
dapp: Better `.dapprc` docs

TransmissionsDev:


TransmissionsDev:
This is just my quick attempt at summarizing some of my findings about `dapprc`'s behavior. lmk how I can improve this :)

src/dapp/README.md
```diff
@@ -265,7 +265,9 @@ The `--verify` flag verifies the contract on etherscan (requires `ETHERSCAN_API_
 ## Configuration
 
 The commands of `dapp` can be customized with environment variables or flags.
-These variables can be set at the prompt or in a `.dapprc` file.
+These variables can be set at the prompt or in a `.dapprc` file. 
+
+Below is a non-comprehensive list of some common configuration options (more can be found in the sourcecode):
```
d-xo, line 270: Which variables are missing? Can we just add them to the list here instead of directing people to the source?

src/dapp/README.md
```diff
@@ -285,8 +287,28 @@ These variables can be set at the prompt or in a `.dapprc` file.
 | `DAPP_BUILD_OPTIMIZE`      | no                         | Activate Solidity optimizer                                                                                                                |
 | `DAPP_BUILD_OPTIMIZE_RUNS` | 200                        | Set the optimizer runs                                                                                                                     |
 
-A global (always loaded) config file is located in `~/.dapprc`.
-A local `.dapprc` can also be defined in your project's root, which overrides variables in the global config.
+A global (always loaded) config file is located in `~/.dapprc`. A local `.dapprc` can also be defined in your project's root, which overrides variables in the global config.
+
+Whenever you run a `dapp` command the most local `.dapprc` available will be executed as a shell script automatically. This means you must prefix your configuration options with the `export` keyword like so:
```
d-xo, line 292:
```suggestion
Whenever you run a `dapp` command the `.dapprc` files are sourced in order (global first, then the one in the current working directory, if it exists). If you wish to set configuration variables, you must use `export` as below:
```

d-xo (commented):
Thanks!

It would also be nice to note the precedence of the various configuration approaches. From reading the source I think we have:

1. command line flags
2. local `.dapprc`
3. global `.dapprc`
4. locally set environment variables

src/dapp/README.md
TransmissionsDev: > Which variables are missing? Can we just add them to the list here instead of directing people to the source?

Well there's like a *bunch* of variables that you can customize that are typically set with flags (verbosity, etc) that I'm not sure you guys want to encourage ppl using as config options? I can add if you do.

src/dapp/README.md
```diff
@@ -289,7 +289,7 @@ Below is a non-comprehensive list of some common configuration options (more can
 
 A global (always loaded) config file is located in `~/.dapprc`. A local `.dapprc` can also be defined in your project's root, which overrides variables in the global config.
 
-Whenever you run a `dapp` command the most local `.dapprc` available will be executed as a shell script automatically. This means you must prefix your configuration options with the `export` keyword like so:
+Whenever you run a `dapp` command the `.dapprc` files are sourced in order (global first, then the one in the current working directory, if it exists). If you wish to set configuration variables, you must use `export` as below:
```
TransmissionsDev, line 292: I would like to still mention that `.dapprc` is a shell script somehow, is there a concise way you would be okay with putting in? @d-xo

TransmissionsDev: oh nvm still mentioned below.

src/dapp/README.md
d-xo: hmmmmm, not sure which ones you mean, but I think I'm fine having all of them here. It's probably also worth linking to the [environment variables](https://github.com/dapphub/dapptools/blob/master/src/hevm/README.md#environment-variables) section in the hevm readme (or maybe even just adding them here?).

src/dapp/README.md
```diff
@@ -285,8 +287,28 @@ These variables can be set at the prompt or in a `.dapprc` file.
...
+Under the hood `.dapprc` is interpreted as a shell script, which means you can add additional scripting logic which will be run whenever you use `dapp`. For example if you wanted to fuzz for many iterations in CI and only a few locally you could add this to your `.dapprc`:
+
+```sh
+if [ "$CI" == "true" ]
+then 
+  export DAPP_TEST_FUZZ_RUNS=1000000 # In CI we want to fuzz for a long time.
+else
+  export DAPP_TEST_FUZZ_RUNS=1000 # When developing locally we only want to fuzz briefly. 
+fi
+```
```
d-xo, line 312:
```suggestion

### Precedence

Configuration options are read in the following order, with later reads overriding earlier ones

1. locally set environment variables
2. global .dapprc
3. local .dapprc 
4. command line flags
```

d-xo:
Thanks! ✨

TransmissionsDev:
oh @d-xo wasn't ready yet heh. was stil adding env vars to the config. should I pick it up in another pr?

TransmissionsDev:
Alright picked it up here: https://github.com/dapphub/dapptools/pull/691

d-xo:
oops, sorry @TransmissionsDev :grinning:

PR #687, opened 2021-07-02
Coverage Guided / Mutation Based Fuzzing

d-xo:
Still very much WIP (and currently based off a pretty outdated commit). 

This PR introduces a coverage guided / mutation based approach to fuzzing. All fuzz tests are now run with coverage enabled. We store each path seen, along with the input that we used to get there. When generating calldata for a fuzz run, we sometimes use the existing strategy (random), and sometimes choose instead to mutate one of the examples from the corpus of visited paths. 

This should hopefully help us get even deeper inside complex contracts.

Still TODO:

- [ ] rebase on master
- [ ] general tidyup
- [ ] generate coverage reports
- [ ] use the symbolic execution engine to prefill the corpus with interesting calldata
- [ ] apply a weighting to some examples (e.g. paths where the symbolic execution engine timed out can be targeted for fuzzing)


src/hevm/src/EVM/ABI.hs
```diff
@@ -134,6 +143,10 @@ data AbiType
   | AbiTupleType        (Vector AbiType)
   deriving (Read, Eq, Ord, Generic)
 
+instance ToJSON AbiType
```
d-xo, line 146: I think all these `ToJSON`/`FromJSON` instances are not neeeded anymore, they're just leftover from a time when I was serializing the corpus as json.

src/hevm/src/EVM/UnitTest.hs
```diff
@@ -101,6 +106,12 @@ data TestVMParams = TestVMParams
   , testChainId       :: W256
   }
 
+-- | For each tuple of (contract, method) we store the calldata required to
+-- reach each known path in the method
+type Corpus = Map (W256,Text) (Map [(W256, Int)] AbiValue)
```
d-xo, line 111: `W256` is the codehash

src/hevm/src/EVM/UnitTest.hs
```diff
@@ -191,23 +202,100 @@ checkFailures UnitTestOptions { .. } method bailed = do
...
+     then pure Pass
+     else pure $ Fail vm' (show . ByteStringS . encodeAbiValue $ args)
+  where
+    updateCorpus k1 k2 v c = case Map.lookup k1 c of
+      Nothing -> Map.insert k1 (Map.insert k2 v mempty) c
+      Just m' -> Map.insert k1 (Map.insert k2 v m') c
+
+type TraceIdState = (VM, [(W256, Int)])
+
+-- | This interpreter is similar to interpretWithCoverage, except instead of
```
d-xo, line 233: This comment is a lie, the hash accumulator was an old idea that actually ended up slowing things down. However I did end up going with a much faster data structure for the traces. Instead of a `MultiSet OpLocation` where `OpLocation` is a `Contract` and a bytecode index, we store insted a list of `(codehash, bytecode index)`. 

Using a codehash keeps the size of the corpus on disk small, and using a list means we can get `O(1)` insertion of new elements by just consing them onto the head of the list (instead of `O(log n)` for a MultiSet). 

With these optimizations in place the extra overhead introduced by the serilization / instrumentation is somewhere between 10 and 15 percent.

d-xo: Note that the use of codehash means we'll need some extra state if we want to show pretty coverage reports to the user (perhaps a mapping from codehash to `SolcContract` or smth in the `Corpus`?)

d-xo: we probably want to rewrite the existing coverage based interpreter to used the new data structures before merging, but I was still just trying to move fast and figure out what worked best so I used a seperate one to start with.

d-xo: oh also it's worth noting that using a `MultiSet` or a list instead of a `Set` for the traces means that traces that get further into a loop will be treated as a new trace, which seems like a nice thing.

transmissions11:
does this include constant mining? https://github.com/crytic/echidna/pull/262

PR #688, opened 2021-07-02
hevm: london hard fork

MrChico:
Still need to update the seth and dapp tooling, but here are the consensus changes


release.nix
```diff
@@ -19,7 +19,7 @@ let
     export PATH=${x.pkgs.hevm}/bin:${x.pkgs.jq}/bin:$PATH
     ${x.pkgs.hevm}/bin/hevm compliance \
       --tests ${ethereum-test-suite x} \
-      --skip "(Create2Recursive|Create1000|recursiveCreateReturn|underflowTest|walletRemoveOwnerRemovePending|Return5000|randomStatetest177|loopExp|loopMul)" \
+      --skip "(Create2Recursive|Create1000|recursiveCreateReturn|underflowTest|walletRemoveOwnerRemovePending|Return5000|randomStatetest177|loopExp|loopMul|FirstByte)" \
```
d-xo, line 22: why are we skipping `FirstByte`?

src/hevm/hevm-cli/hevm-cli.hs
```diff
@@ -749,14 +754,16 @@ vmFromCommand cmd = do
               then addr address (createAddress origin' (word nonce 0))
               else addr address 0xacab
 
-        vm0 miner ts blockNum diff c = EVM.makeVm $ EVM.VMOpts
+        vm0 baseFee miner ts blockNum diff c = EVM.makeVm $ EVM.VMOpts
```
d-xo, line 757: haskell question: why isn't the `baseFee` fetched from rpc above in scope here?

d-xo (commented):
consensus changes look great :sparkling_heart:

src/hevm/src/EVM.hs
```diff
@@ -377,6 +381,7 @@ data Block = Block
   , _number      :: Word
   , _difficulty  :: Word
   , _gaslimit    :: Word
+  , _baseFee     :: Word
```
d-xo, line 384: I guess this needs to be a `Maybe` to account for pre-1559 blocks?

release.nix
MrChico: Those tests do not have a `poststate`, just the poststate hash, which we are not deciphering at the moment

src/hevm/hevm-cli/hevm-cli.hs
MrChico: because variables defined in `where` cannot "look inside" and see vars defined in the monadic context.

src/hevm/src/EVM.hs
MrChico: we haven't bothered to make hevm backwards compatible, so I think this is fine as long as we signal that hevm is now on 1559

d-xo: I think it's gonna break `seth run-tx` on old blocks?

d-xo: Although yeah the prospect of worrying aobut backwards compatibility in hevm does not spark joy...

release.nix
d-xo: can we document that as a comment in this file? Would be awesome to have reasons for all the skips here actually, but that doesn't have to be done now.

d-xo:
Do we handle the ropsten issue (https://notes.ethereum.org/@timbeiko/ropsten-postmortem) correctly? I can't figure it out froma quick scan of the code...

src/hevm/src/EVM/Fetch.hs
```diff
@@ -103,12 +104,14 @@ fetchQuery n f q = do
 
 parseBlock :: (AsValue s, Show s) => s -> Maybe EVM.Block
 parseBlock j = do
+  traceM (show j)
```
MrChico, line 107:
```suggestion
```

src/hevm/src/EVM/Fetch.hs
```diff
@@ -7,6 +7,7 @@ module EVM.Fetch where
 import Prelude hiding (Word)
 
 import EVM.ABI
+import Debug.Trace
```
MrChico, line 10:
```suggestion
```

MrChico:
The seth helpers can be implemented in another pr. I think this is ready for review

haskell.nix
```diff
@@ -57,8 +39,7 @@ in self-hs: super-hs:
       configureFlags = attrs.configureFlags ++ [
           "--ghc-option=-O2"
       ] ++
-      (if stdenv.isDarwin then [] else
-        if shared then [] else [
```
d-xo, line 61: why remove the `shared` stuff here?

MrChico: mistake

haskell.nix
```diff
@@ -57,8 +39,7 @@ in self-hs: super-hs:
...
+      (if stdenv.isDarwin then [] else [
```
MrChico, line 42:
```suggestion
      (if stdenv.isDarwin then [] else
          if shared then [] else [
```

haskell.nix
```diff
@@ -58,7 +40,7 @@ in self-hs: super-hs:
           "--ghc-option=-O2"
       ] ++
       (if stdenv.isDarwin then [] else
-        if shared then [] else [
+          if shared then [] else [
```
MrChico, line 43:
```suggestion
       if shared then [] else [
```

haskell.nix
```diff
@@ -58,7 +40,7 @@ in self-hs: super-hs:
           "--ghc-option=-O2"
       ] ++
       (if stdenv.isDarwin then [] else
-        if shared then [] else [
+       if shared then [] else [
```
MrChico, line 43:
```suggestion
        if shared then [] else [
```

haskell.nix
MrChico: nice catch, thx

d-xo (approved):
looks good!

PR #690, opened 2021-07-03
Fix storage layout parsing

d-xo:
Fixes some issues with json parsing:

1. storage layout was being parsed as a string, when it is an object in the std-json output (and also in the combined json output post solc 0.8)
2. abis were being parsed as a string in the combined json output, but they are an object post solc 0.8

Note that this change currently breaks parsing of storage layout for combined-json output from solc < 0.8. I'm happy to make this work if we consider it important, but since it's a deprecated option, and I couldn't figure out a clean way to make it work after 20 minutes I just left it alone.

We could maybe also just consider removing the `combinedJSON` parsing code as part of this PR as well?

### DIFF
--- a/src/hevm/src/EVM/Solidity.hs
+++ b/src/hevm/src/EVM/Solidity.hs
@@ -334,8 +334,9 @@ readCombinedJSON json = do
       let
         theRuntimeCode = toCode (x ^?! key "bin-runtime" . _String)
         theCreationCode = toCode (x ^?! key "bin" . _String)
-        abis =
-          toList ((x ^?! key "abi" . _String) ^?! _Array)
+        abis = toList $ case (x ^?! key "abi") ^? _Array of
+                 Just v -> v                                       -- solc >= 0.8
+                 Nothing -> (x ^?! key "abi" . _String) ^?! _Array -- solc <  0.8
       in SolcContract {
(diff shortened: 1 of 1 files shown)


MrChico:
> Note that this change currently breaks parsing of storage layout for combined-json output from solc < 0.8. I'm happy to make this work if we consider it important, but since it's a deprecated option, and I couldn't figure out a clean way to make it work after 20 minutes I just left it alone.

Soft breakage, right? If so I'm fine with this

d-xo:
wdym soft breakage? with this PR we will have a hard error if we try to parse storage-layout from combined json < solc0.8

MrChico:
To me it looks like we would just have `storageLayout = Nothing`, and a soft error ,but I haven't run it

d-xo:
oh yes, that's probably true, I'm afk right now but I'll double check again this afternoon and merge if that's the case.

MrChico (approved):

MrChico:
hmm, I'm having some other problems in this department, looks like the std json that is being used around here is actually malformed. Investigating

d-xo:
oh, do you want to hold of merging this PR then?

MrChico:
actually, I realized my problems was I was not using the same version of solc (it is not imported in the nix-shell) With the solc specified in the overlay, this should be fine!

PR #697, opened 2021-07-13
fix debug bug

MrChico:
whenever one tries to debug a fuzz test which features a type which doesn't have a corresponding symbolic version (i.e. `bytes` or other dynamically sized types), hevm currently fails as we try to generate a symbolic argument (which will be thrown away regardless). This quick fix ensures that we only create the symbolic argument if it is actually a symbolic test

### DIFF
--- a/src/hevm/src/EVM/TTY.hs
+++ b/src/hevm/src/EVM/TTY.hs
@@ -623,7 +623,9 @@ initialUiVmStateForTest
   -> IO UiVmState
 initialUiVmStateForTest opts@UnitTestOptions{..} (theContractName, theTestName) = do
   let state' = fromMaybe (error "Internal Error: missing smtState") smtState
-  (buf, len) <- flip runReaderT state' $ SBV.runQueryT $ symCalldata theTestName types []
+  (buf, len) <- case test of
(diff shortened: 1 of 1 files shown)


src/hevm/src/EVM/TTY.hs
```diff
@@ -623,7 +623,9 @@ initialUiVmStateForTest
   -> IO UiVmState
 initialUiVmStateForTest opts@UnitTestOptions{..} (theContractName, theTestName) = do
   let state' = fromMaybe (error "Internal Error: missing smtState") smtState
-  (buf, len) <- flip runReaderT state' $ SBV.runQueryT $ symCalldata theTestName types []
+  (buf, len) <- case test of
```
d-xo, line 626: Can't we just create this in the `SymbolicTest` branche of the `case test of` below? I don't think it's used anywere else?

MrChico: unfortunately not because that is a `Stepper` type and not Symbolic. We could generalize EVMIO to be a `EVMSymbolic` or smth like that, but this felt like the easiest fix right now

d-xo (approved):
looks good!

PR #698, opened 2021-07-14
Soundness issue

MrChico:
Consider the following yul code, which asserts when the given calldata address has extra bytes beyond the 20 address bytes:
`bugdemo.yul`:
```
object "Token" {
  code {
   // Deploy the contract
   datacopy(0, dataoffset("runtime"), datasize("runtime"))
     return(0, datasize("runtime"))
  }
  object "runtime" {
   code {
    switch selector()
      case 0x70a08231 /* "balanceOf(address)" */ {
       pop(decodeAsAddress(0))
    }
  function selector() -> s {
     s := div(calldataload(0), 0x100000000000000000000000000000000000000000000000000000000)
  }
  function decodeAsAddress(offset) -> v {
     v := calldataload(add(4, mul(offset, 0x20)))
     if iszero(iszero(and(v, not(0xffffffffffffffffffffffffffffffffffffffff)))) {
        invalid()
     }
  }
}
}
}
```

When we supply `--sig balanceOf(address)`, calldata should be properly formed, and we should not end up violating the assertion, and yet:
```
hevm symbolic --code $(hevm exec --code $(erc20-example/solc-0.8.2 bugdemo.yul --strict-assembly | awk '/Binary representation:/{getline; print}') --gas 0xffff) --sig "balanceOf(address)"

Assertion violation found.
Calldata:
0x70a082310000000000000000000000000000000000000000000000000000000000000001
balanceOf(0x0000000000000000000000000000000000000001)
Caller:
0x0000000000000000000000000000000000000000
Callvalue:
0
```

I have tracked down this issue to this simplifcation rule, but why this is problematic is still a bit mysterious to me

### DIFF
--- a/src/hevm/src/EVM/Solidity.hs
+++ b/src/hevm/src/EVM/Solidity.hs
@@ -334,8 +334,9 @@ readCombinedJSON json = do
       let
         theRuntimeCode = toCode (x ^?! key "bin-runtime" . _String)
         theCreationCode = toCode (x ^?! key "bin" . _String)
-        abis =
-          toList ((x ^?! key "abi" . _String) ^?! _Array)
+        abis = toList $ case (x ^?! key "abi") ^? _Array of
+                 Just v -> v                                       -- solc >= 0.8
+                 Nothing -> (x ^?! key "abi" . _String) ^?! _Array -- solc <  0.8
       in SolcContract {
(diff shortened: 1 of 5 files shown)


d-xo:
Closing in favour of: https://github.com/dapphub/dapptools/pull/700

PR #700, opened 2021-07-16
hevm: Symbolic: correct semantics for Neg in whiffValue

d-xo:
`NOT` is bitwise negation, which is implemented in haskell with `complement` not `negate`. This was causing the soundness issue in: https://github.com/dapphub/dapptools/pull/698.


MrChico:
nice. This is a better fix!

MrChico (approved):

PR #703, opened 2021-07-19
More README improvements

TransmissionsDev:
Screen Shot 2021-07-19 at 9 35 50 AM

Originally made this PR was to write down how `ethsign import` can be used to generate a keystore file `seth` can ingest, but got a bit carried away and added some more env vars to the `dapp` table and cleaned up some grammar/spelling/casing 😅

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,16 +2,16 @@
 
 Hello!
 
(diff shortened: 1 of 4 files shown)


d-xo:
haha, thanks as always!!! :sparkling_heart:

PR #706, opened 2021-07-21
dapp/seth: Dead links and more config options

TransmissionsDev:
lol sorry i know i make like 8 of these a week but here u go:

1. add some SMT checker config options that were missing from the table
2. fix broken seth links introduced by @mds1's PR 😅

PR #707, opened 2021-07-21
hevm: Bump default smttimeout to 1 min (60000 ms)

TransmissionsDev:
The current default is just too low for `z3`, the 10 minute timeout used in `dapp-tests` is probably best for most projects.

(This PR includes the doc updates made in #706)


d-xo:
Not sure whats up with the diff, there seem to be a lot of unrelated changes from https://github.com/dapphub/dapptools/pull/706.

I think increasing the timeout after the switch to z3 makes sense, but 10 minutes seems very long. I'm worried that having to wait 20 or 30 minutes (the timeout is per smt query not per test so it can take a lot longer than 10 mins to get through one test if there are many branches to check) just to see your tests fail with a timeout the first time you run them is a bad experience. I think it might be better to have a shorter default and then users can increase it for the tests that really need it.

Do you have some examples of tests that need a 10 minute timeout? For the tests in ci it was just mostly tests involving a lot of non linear arithmetic.

TransmissionsDev:
yea fair, i don't think i have a test with a 10 minute query need tbh. i'll update to use a lower timeout (maybe 1-2 minutes?) 

i'd also like to update that error message, cuz for someone not familiar with dapptools i'm not sure they'd know to update the smt timeout env var just from "query timeout"

d-xo:
I'm fine with 1-2 minutes. Updating the error message is a great idea :slightly_smiling_face:

TransmissionsDev:
aight, added some info about the timeout in the docs, updated the error msg, and set the timeout to 1 min

src/dapp/README.md
```diff
@@ -197,6 +197,10 @@ Failure: proveFail_withdraw(uint256)
 
 which demonstrates that if we give the password `42`, it is possible to withdraw from the vault.
 
+When using symbolic execution for more complex tests, you may begin running into the SMT timeout, which bounds how long the hevm will
+try to execute a particular part of your code before bailing. If your tests begin failing with "`SMT Query Timeout!`", consider increasing
+it using the `--smttimeout` flag or `DAPP_TEST_SMTTIMEOUT` environment variable (the default timeout is 60000 ms).
```
d-xo, line 202: What do you think of this?

```suggestion
The symbolic execution engine is backed by an SMT solver. When symbolically executing more complex tests you may encounter test failures with an `SMT Query Timeout` message. In this case, consider increasing the smt timeout using the `--smttimeout` flag or `DAPP_TEST_SMTTIMEOUT` environment variable (the default timeout is 60000 ms). Note that this timeout is per smt query not per test, and that each test may execute multiple queries (at least one query for each potential path through the test method).
```

TransmissionsDev: ah much better, thx!

PR #708, opened 2021-07-21
dapp/seth: use solc-static for --use

TransmissionsDev:
i'm not sure if there's a reason we currently use `solc-versions` but trying this out because `solc-versions` is missing most of the modern solc builds that i depend on lol


d-xo:
I'm fine with this. @MrChico is this going to make life dificult for you on your vintage mac?

PR #710, opened 2021-07-22
hevm: symexec: verify: differentiate timeouts and success

d-xo:
the result returned from `verify` did not distinguish between a succesful run and a run where the solver timed out on the postcondition query, meaning that users of `verify` would treat a timeout as a success, and resulted in e.g. act printing `Q.E.D` even if the query timed out.

### DIFF
--- a/src/hevm/hevm-cli/hevm-cli.hs
+++ b/src/hevm/hevm-cli/hevm-cli.hs
@@ -516,12 +516,14 @@ assert cmd = do
       preState <- symvmFromCommand cmd
       let errCodes = fromMaybe defaultPanicCodes (assertions cmd)
       verify preState (maxIterations cmd) rpcinfo (Just $ checkAssertions errCodes) >>= \case
-        Right tree -> do
+        Cex tree -> do
           io $ putStrLn "Assertion violation found."
           showCounterexample preState maybesig
           treeShowing tree
           io $ exitWith (ExitFailure 1)
-        Left tree -> do
+        Perhaps -> do
+          io $ exitWith (ExitFailure 1)
+        Qed tree -> do
           io $ putStrLn $ "Explored: " <> show (length tree)
                        <> " branches without assertion violations"
           treeShowing tree
--- a/src/hevm/hevm-cli/hevm-cli.hs
+++ b/src/hevm/hevm-cli/hevm-cli.hs
@@ -420,16 +420,19 @@ equivalence cmd =
 
      void . runSMTWithTimeOut (solver cmd) (smttimeout cmd) (smtdebug cmd) . query $
        equivalenceCheck bytecodeA bytecodeB (maxIterations cmd) maybeSignature >>= \case
-         Right vm -> do io $ putStrLn "Not equal!"
-                        io $ putStrLn "Counterexample:"
-                        showCounterexample vm maybeSignature
-                        io exitFailure
-         Left (postAs, postBs) -> io $ do
+         Cex vm -> do
+           io $ putStrLn "Not equal!"
+           io $ putStrLn "Counterexample:"
+           showCounterexample vm maybeSignature
+           io exitFailure
+         Qed (postAs, postBs) -> io $ do
            putStrLn $ "Explored: " <> show (length postAs)
                        <> " execution paths of A and: "
                        <> show (length postBs) <> " paths of B."
            putStrLn "No discrepancies found."
-
+         Timeout () -> io $ do
+           putStrLn "Solver timeout!"
+           exitFailure
 
 -- cvc4 sets timeout via a commandline option instead of smtlib `(set-option)`
 runSMTWithTimeOut :: Maybe Text -> Maybe Integer -> Bool -> Symbolic a -> IO a
@@ -521,7 +524,7 @@ assert cmd = do
           showCounterexample preState maybesig
           treeShowing tree
           io $ exitWith (ExitFailure 1)
-        Perhaps -> do
+        Timeout _ -> do
           io $ exitWith (ExitFailure 1)
         Qed tree -> do
           io $ putStrLn $ "Explored: " <> show (length tree)
--- a/src/hevm/hevm-cli/hevm-cli.hs
+++ b/src/hevm/hevm-cli/hevm-cli.hs
@@ -431,7 +431,7 @@ equivalence cmd =
                        <> show (length postBs) <> " paths of B."
            putStrLn "No discrepancies found."
          Timeout () -> io $ do
-           putStrLn "Solver timeout!"
+           hPutStr stderr "Solver timeout!"
            exitFailure
 
 -- cvc4 sets timeout via a commandline option instead of smtlib `(set-option)`
--- a/src/hevm/hevm-cli/hevm-cli.hs
+++ b/src/hevm/hevm-cli/hevm-cli.hs
@@ -290,7 +290,7 @@ unitTestOptions cmd testFile = do
          Just url -> EVM.Fetch.oracle (Just state) (Just (block', url)) True
          Nothing  -> EVM.Fetch.oracle (Just state) Nothing True
     , EVM.UnitTest.maxIter = maxIterations cmd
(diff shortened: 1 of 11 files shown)


MrChico:
A similar prpblem exists if we reach `--max-iterations`. We should make sure all "inconclusive" results are not marked qed

src/hevm/hevm-cli/hevm-cli.hs
```diff
@@ -516,12 +519,14 @@ assert cmd = do
...
       verify preState (maxIterations cmd) rpcinfo (Just $ checkAssertions errCodes) >>= \case
-        Right tree -> do
+        Cex tree -> do
           io $ putStrLn "Assertion violation found."
           showCounterexample preState maybesig
           treeShowing tree
           io $ exitWith (ExitFailure 1)
-        Left tree -> do
+        Timeout _ -> do
+          io $ exitWith (ExitFailure 1)
```
MrChico, line 528:
```suggestion
          io $ do hPutStr stderr "Solver timeout!"
                       exitWith (ExitFailure 1)
```

src/hevm/hevm-cli/hevm-cli.hs
```diff
@@ -420,16 +420,19 @@ equivalence cmd =
...
+           showCounterexample vm maybeSignature
+           io exitFailure
+         Qed (postAs, postBs) -> io $ do
            putStrLn $ "Explored: " <> show (length postAs)
                        <> " execution paths of A and: "
                        <> show (length postBs) <> " paths of B."
            putStrLn "No discrepancies found."
-
+         Timeout () -> io $ do
+           putStrLn "Solver timeout!"
```
MrChico, line 434:
```suggestion
           hPutStr stderr "Solver timeout!"
```

src/hevm/src/EVM/Solidity.hs
```diff
@@ -299,15 +299,17 @@ readSolc fp =
 yul :: Text -> Text -> IO (Maybe ByteString)
 yul contract src = do
   (json, path) <- yul' src
-  let (Just c) = json ^?! key "contracts" ^?! key path ^? key contract
-  let bytecode = c ^?! key "evm" ^?! key "bytecode" ^?! key "object" . _String
+  let (Just f) = json ^?! key "contracts" ^? key path
+      (Just c) = if not (Text.null contract) then f ^? key contract else f ^? key "object"
```
MrChico, line 303:
```suggestion
      (Just c) = f ^? key $ if Text.null contract then "object" else contract
```

src/hevm/src/EVM/SymExec.hs
```diff
@@ -349,33 +353,39 @@ leaves (Node _ xs) = concatMap leaves xs
...
+verify :: VM -> Maybe Integer -> Maybe (Fetch.BlockNumber, Text) -> Maybe Postcondition -> Query VerifyResult
 verify preState maxIter rpcinfo maybepost = do
   smtState <- queryState
   tree <- doInterpret (Fetch.oracle (Just smtState) rpcinfo False) maxIter preState
   case maybepost of
     (Just post) -> do
       let livePaths = pruneDeadPaths $ leaves tree
-      -- can also do these queries individually (even concurrently!). Could save time and report multiple violations
+          -- have we hit max iterations for any path?
+          maxReached = or $ mapMaybe (flip maxIterationsReached maxIter) livePaths
```
MrChico, line 364: I don't think this will work? We will reach max iterations somewhere in the middle of a branch, but if we do we continue down to the leaf anyway, so the leaf won't remember that information. I think it might be better to add a Bool flag to `doInterpret` (and `interpret`) instead and make sure we mark max iterations there

src/hevm/src/EVM/SymExec.hs
```diff
@@ -349,33 +353,39 @@ leaves (Node _ xs) = concatMap leaves xs
...
-                  return $ Left tree
-        Unsat -> do io $ putStrLn "Q.E.D."
-                    return $ Left tree
-        Sat -> return $ Right tree
+                  return $ Timeout tree
+        Unsat -> do
+          if maxReached
+            then do
+              io $ putStrLn "WARNING: max iterations reached, execution halted prematurely"
+              io $ putStrLn "no postcondition violations discovered"
```
MrChico, line 378: we should maybe also fail here?

src/hevm/test/test.hs
```diff
@@ -620,20 +629,29 @@ main = defaultMain $ testGroup "hevm"
...
-            bPrgm = hex "6020600060003760005160008114601c5760028114602057fe6021565b6021565b5b506001600152"
+        Just aPrgm <- yul ""
+          [i|
+          {
+              calldatacopy(0, 0, 32)
+              switch mload(0)
+              case 0 { }
+              case 1 { }
+              default { invalid() }
+          }
```
MrChico, line 640: nice

src/hevm/src/EVM/SymExec.hs
d-xo: yes, you're right. I hadn't tested this properly :(

By mark do you just mean print a warning?

src/hevm/src/EVM/SymExec.hs
d-xo: yeah I wasn't quite sure what was best, I guess I feel like it's more of a warning than a hard error, so we shouldn't exit with a failure, but I'm fine to change it if you have strong opinions...

src/hevm/hevm-cli/hevm-cli.hs
d-xo: we already print this in `verify` so I left it out here...

src/hevm/src/EVM/SymExec.hs
d-xo: I suppose if we could maybe add a flag to the `_iterations` state in `VM` that would let us know if max iterations had ever been reached at any point along a branch?

d-xo: or perhaps we can just check every code location in the _iterations mapping for each vm?

MrChico: I'm thinking to just pass along a bool in the `doInterpret` function would be simplest?

d-xo:
@MrChico I think this is good to go. 

Afaict we are able to diffentiate between an execution that touches a location `maxIterations - 1` times and one that touches that location `maxIterations` times, since we always call the [`choosePath`](https://github.com/dapphub/dapptools/blob/873766fd0c3902b88444d3c023a44d5ff0352e3b/src/hevm/src/EVM.hs#L1625) continuation with a concrete `Case` whether or not we hit max iterations (and this continuation always increments the `iterations` field in the `Case` branch).

I also checked with a test contract and we do not get the warning regarding max iterations for the following contract if we call `hevm ` with `--max-iterations 7`, but we do with `--max-iterations 6` (the `JUMPI` for the loop is hit 6 times including the final check):

```solidity
contract MaxIter {
    function prove_loop(uint8 n) public pure {
        if (n > 5) return;

        uint count;
        for (uint8 i; i < n; i++) {
            count++;
        }
    }
}
```

Sorry for all the noise in this PR, it kinda kept growing and I ended up fixing a bunch of compiler warnings along the way...

src/hevm/src/EVM/SymExec.hs
```diff
@@ -347,36 +352,42 @@ leaves (Node x []) = [_vm x]
...
   case maybepost of
     (Just post) -> do
       let livePaths = pruneDeadPaths $ leaves tree
-      -- can also do these queries individually (even concurrently!). Could save time and report multiple violations
+          -- have we hit max iterations at any point in a given path
+          maxReached :: VM -> Bool
+          maxReached p = case maxIter of
+            Just maxI -> any (>= (fromInteger maxI)) (view iterations p)
+            Nothing -> False
+          --maxReached = or $ mapMaybe (flip maxIterationsReached maxIter) livePaths
```
MrChico, line 367:
```suggestion
```

MrChico (approved):

MrChico:
Looks great!

PR #711, opened 2021-07-26
fetch contract code, nonce and balance asynchronously

MrChico:
Not the optimization I set out to make, but this one is quite simple and still provides some marginal improvement.
Fetches account code, balance and nonce concurrently instead of sequentially. This provides around 30% speedup of a single account fetching. Against a larger example (`--match test_reward_wound_fully` of https://github.com/rainbreak/crop), the improvement is less pronounced (~1:52 to 1:50), but for the size of the diff it's maybe worth it still

### DIFF
--- a/src/hevm/hevm.cabal
+++ b/src/hevm/hevm.cabal
@@ -81,6 +81,7 @@ library
     ethjet/tinykeccak.h, ethjet/ethjet.h, ethjet/ethjet-ff.h, ethjet/blake2.h
   build-depends:
     QuickCheck                        >= 2.13.2 && < 2.15,
+    async,
     Decimal                           >= 0.5.1 && < 0.6,
     containers                        >= 0.6.0 && < 0.7,
     deepseq                           >= 1.4.4 && < 1.5,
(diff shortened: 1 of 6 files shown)


src/hevm/src/EVM/Fetch.hs
```diff
@@ -117,20 +123,19 @@ fetchWithSession url sess x = do
...
-  theCode    <- MaybeT $ fetch (QueryCode addr)
-  theNonce   <- MaybeT $ fetch (QueryNonce addr)
-  theBalance <- MaybeT $ fetch (QueryBalance addr)
-
-  return $
-    initialContract (EVM.RuntimeCode (ConcreteBuffer theCode))
-      & set nonce    (w256 theNonce)
-      & set balance  (w256 theBalance)
-      & set external True
+  in do
```
MrChico, line 130:
```suggestion
  in
```

src/hevm/hevm.cabal
```diff
@@ -81,6 +81,7 @@ library
     ethjet/tinykeccak.h, ethjet/ethjet.h, ethjet/ethjet-ff.h, ethjet/blake2.h
   build-depends:
     QuickCheck                        >= 2.13.2 && < 2.15,
+    async,
```
d-xo, line 84: I think we need version bounds here for hackage releases?

d-xo:
can you add a changelog entry as well pls :sparkling_heart:

src/hevm/hevm.cabal
MrChico: fixed

d-xo (approved):
looks good!

PR #716, opened 2021-07-29
hevm: add env var to control nonce

d-xo:
- Adds a `DAPP_TEST_NONCE` env var that allows users to control the nonce assigned to the test contract
- Adds unit tests for the various `DAPP_TEST_*` env vars

cc: @kmbarry1 you were asking for this a while back I believe?

### DIFF
--- a/nix/build-dapp-package.nix
+++ b/nix/build-dapp-package.nix
@@ -23,8 +23,9 @@ in
   , deps ? []
   , solc ? "${pkgs.solc}/bin/solc"
   , shouldFail ? false
-  , dappFlags ? ""
   , doCheck ? true
+  , dapprc ? ""
+  , testFlags ? ""
   , ... }@args:
     pkgs.stdenv.mkDerivation ( rec {
       inherit doCheck;
@@ -47,6 +48,7 @@ in
           '')
           passthru.libPaths;
       buildPhase = ''
+        source ${pkgs.writeText "dapprc" dapprc}
         mkdir -p out
         export DAPP_SOLC=${solc}
         export DAPP_REMAPPINGS="$REMAPPINGS"
@@ -59,7 +61,7 @@ in
       '';
 
       checkPhase = let
-        cmd = "DAPP_SKIP_BUILD=1 dapp test ${dappFlags}";
+        cmd = "DAPP_SKIP_BUILD=1 dapp test ${testFlags}";
       in
         if shouldFail
           then "${cmd} && exit 1 || echo 0"
--- a/src/hevm/CHANGELOG.md
+++ b/src/hevm/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Added
+
+- A new configuration variable `DAPP_TEST_NONCE` has been added that allows control over the nonce of the testing contract
+
 ### Changed
 
 - The configuration variable `DAPP_TEST_BALANCE_CREATE` has been renamed to `DAPP_TEST_BALANCE`
--- a/src/dapp-tests/default.nix
+++ b/src/dapp-tests/default.nix
@@ -114,11 +114,10 @@ let
     deps = [ ds-test ds-thing ];
   };
 
-  runTest = { dir, shouldFail, name, dappFlags?"" }: pkgs.buildDappPackage {
-    inherit name shouldFail;
+  runTest = { dir, shouldFail, name, dapprc ? "", testFlags ? "" }: pkgs.buildDappPackage {
+    inherit name shouldFail testFlags dapprc;
     solc=solc-0_6_7;
     src = dir;
-    dappFlags = "${dappFlags}";
     deps = [ ds-test ds-token ds-math ];
     checkInputs = with pkgs; [ hevm jq seth dapp solc ];
   };
@@ -128,7 +127,35 @@ in
       dir = ./pass;
       name = "dappTestsShouldPass";
       shouldFail = false;
-      dappFlags = "--max-iterations 50 --smttimeout 600000 --ffi";
+      testFlags = "--max-iterations 50 --smttimeout 600000 --ffi";
+    };
+
+    envVars = let
+      envVarTest = match : dapprc : runTest {
+        testFlags = "--match ${match}";
+        dir = ./env;
+        name = "dappTestEnvVar";
+        shouldFail = false;
+        inherit dapprc;
+      };
+      seth = "${pkgs.seth}/bin/seth";
+    in pkgs.recurseIntoAttrs {
+      # we get "hevm: insufficient balance for gas cost" if we run these together...
+      origin = envVarTest "origin.sol" "export DAPP_TEST_ORIGIN=$(${seth} --to-hex 256)";
+      rest = envVarTest "rest.sol" ''
+        export DAPP_TEST_BALANCE=$(${seth} --to-wei 998877665544 ether)
+        export DAPP_TEST_ADDRESS=$(${seth} --to-hex 256)
+        export DAPP_TEST_CALLER=$(${seth} --to-hex 100)
+        export DAPP_TEST_GAS_CREATE=$(${seth} --to-wei 4.20 ether)
+        export DAPP_TEST_GAS_CALL=$(${seth} --to-wei 0.69 ether)
+        export DAPP_TEST_NONCE=100
+        export DAPP_TEST_COINBASE=$(${seth} --to-hex 666)
+        export DAPP_TEST_NUMBER=420
+        export DAPP_TEST_TIMESTAMP=69
+        export DAPP_TEST_GAS_LIMIT=$(${seth} --to-wei 4206966 ether)
+        export DAPP_TEST_GAS_PRICE=100
+        export DAPP_TEST_DIFFICULTY=600
+      '';
     };
 
     shouldFail = let
@@ -136,7 +163,7 @@ in
         dir = ./fail;
         shouldFail = true;
         name = "dappTestsShouldFail-${match}";
-        dappFlags = "--match ${match} --smttimeout 600000";
+        testFlags = "--match ${match} --smttimeout 600000";
       };
     in pkgs.recurseIntoAttrs {
       prove-add = fail "prove_add";
@@ -155,7 +182,7 @@ in
       solc = solc-0_6_7;
       src = dss-src;
       name = "dss";
-      dappFlags = "--match '[^dai].t.sol'";
+      testFlags = "--match '[^dai].t.sol'";
       deps = [ ds-test ds-token ds-value ];
     };
   }
--- a/src/dapp-tests/env/origin.sol
+++ b/src/dapp-tests/env/origin.sol
@@ -1,0 +1,8 @@
+import "ds-test/test.sol";
+
+contract Env is DSTest {
+    // DAPP_TEST_ORIGIN
+    function testOrigin() public {
+        assertEq(tx.origin, address(256));
+    }
+}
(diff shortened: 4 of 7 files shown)


nix/build-dapp-package.nix
```diff
@@ -47,6 +48,7 @@ in
           '')
           passthru.libPaths;
       buildPhase = ''
+        source ${pkgs.writeText "dapprc" dapprc}
```
d-xo, line 51: @asymmetric do you know how I could actually write this file as a `.dapprc` into the source directory used for the nix build? I always ended up running into permissions errors.

asymmetric: Why do you want the dapprc to be in the source directory? You need it to be somewhere at runtime right? So it should be in the final store path, no?

Or do you actually need it at build-time because it changes the build behaviour?

d-xo: build time only

d-xo: Although now you mention it it should be in the current working directory instead of the $src dir so I think I was just trying to put it in the wrong place...

nix/build-dapp-package.nix
```diff
@@ -47,6 +48,7 @@ in
           '')
           passthru.libPaths;
       buildPhase = ''
+        ln -s ${pkgs.writeText "dapprc" dapprc} ./.dapprc
```
asymmetric, line 51: Instead of this, it could be nice to have the option to pass a `dapprc` explicitly as an argument (to `dapp build` I guess?).

Another (unrelated) nit is that `.dapprc` should probably be (first) looked up in `XDG_CONFIG_HOME`

kmbarry1:
> cc: @kmbarry1 you were asking for this a while back I believe?

Hey! Yeah, this gets partially at what I was thinking of. The even more powerful thing would actually be what you suggested in chat, a general `hevm` cheat code for setting any env value, i.e. `hevm.set("nonce", N)` or `hevm.set("currentAddress", 0x1234)`, etc.

d-xo:
Yeah definitely agree re: the cheat code, but that's a bit of a bigger job for another day.

kmbarry1:
> Yeah definitely agree re: the cheat code, but that's a bit of a bigger job for another day.

Sounds good--and I might still take a crack at it if ever I have some free time :sweat_smile: .

d-xo:
So I ended up adding the cheatcodes.

I played around with a cheatcode that could change the address of any contract (`hevm.file("address", usr, newAddress)`), but there are quite a few places in hevm where we expect addresses to stay constant (e.g. we make multiple calls into the same address when running unit tests), and it just felt like it was gonna being more trouble than it was worth to add.

In the end I added the following:

```solidity
/*
    allows setting the nonce and balance of an arbitrary account
*/
hevm.file(bytes32, address, uint)

/*
    allows setting the caller, origin, and coinbase
*/
hevm.file(bytes32, address)

/*
     allows setting gasPrice, tsGasLimit, blockGasLimit, difficulty
*/
hevm.file(bytes32, uint)

/*
    allows reading the nonce for any address
*/
hevm.look(bytes32, address)

/*
    allows reading the txGasLimit
*/
hevm.look(bytes32)

/*
    allows replacing the code at any address (except the currently executing contract)
*/
hevm.replace(address, bytes)
```

I think these give almost exactly the same level of control as the `DAPP_TEST*` env vars, but inside of unit tests via cheat codes. I wasn't completely sure about the interface for the `look` cheat codes, since they both only have one allowed value for the `bytes32` param, but I kinda liked the consistency of the interface, open to suggestions if someone has a better idea.

@kmbarry1 I think with `replace` and `file("nonce", val)` you should be able to do what you wanted with bytecode verification? I think you should be able to use replace to set the code of the adress you want to deploy from to a contract that just deploys the contract you want to verify and use `file` to set the nonce. Is that enough or did you need something more?

src/hevm/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Added
```
d-xo, line 5: document new cheatcodes....

TransmissionsDev:
hollllyyyy cow this is amazing @d-xo. only thing that i think is left out that i would kill to have is the ability to change the address of the test contract like you can with `DAPP_TEST_ADDRESS` but I know you said thats infeasible 😭

src/hevm/src/EVM.hs
```diff
@@ -2012,6 +2011,79 @@ cheatActions =
...
+                "nonce" -> forceConcrete val $ \v -> assign (env . contracts . ix usr . nonce) v
+                "balance" -> forceConcrete val $ \v -> assign (env . contracts . ix usr . balance) v
+                _ -> vmError (BadCheatCode sig)
+          _ -> vmError (BadCheatCode sig),
+
+      action "file(bytes32,address)" $
+        \sig _ _ input -> case decodeStaticArgs input of
+          [what', addr] -> forceConcrete what' $ \(C _ (utf8Word -> what)) ->
+            case what of
+              "caller" -> assign (state . caller) (SAddr . sFromIntegral . rawVal $ addr)
```
d-xo, line 2028: @MrChico I'm a little unclear on the details of the `frames` stack? do we need to change this anywhere else? I'm worried that this is going to mess up returning from the current frame?

src/hevm/src/EVM.hs
```diff
@@ -2552,15 +2624,14 @@ checkJump x xs = do
...
-        vmError BadJumpDestination
-    else vmError BadJumpDestination
+  if x < num (len theCode)
+     && 0x5b == (fromMaybe (error "tried to jump to symbolic code location") $ unliteral $ EVM.Symbolic.index (num x) theCode)
+     && OpJumpdest == snd (theCodeOps RegularVector.! (theOpIxMap Vector.! num x))
+  then do
+    state . stack .= xs
+    state . pc .= num x
+  else do
+    vmError BadJumpDestination
```
d-xo, line 2634: no semantic change here, just flattening the if statement since the `else` clauses are identical...

src/dapp-tests/pass/cheatCodes.sol
```diff
@@ -87,4 +102,54 @@ contract CheatCodes is DSTest {
...
+        hevm.file("nonce", who, n);
+        assertEq(hevm.look("nonce", who), n);
+    }
+
+    function testBalance(address who, uint bal) public {
+        hevm.file("balance", who, bal);
+        assertEq(who.balance, bal);
+    }
+
+    function testReplace() public {
```
d-xo, line 116: no idea why at all, but if I construct the target contract directly in the test instead of in the constructor this test always fails (and it does so before reaching the call to the cheatcode).... :thinking: 

```suggestion
    function testReplace() public {
        Old target = new Old();
```

MrChico: Hmm, strange indeed. Does it revert or what is the error?

src/hevm/src/EVM.hs
MrChico: You should see how this is handled in `finishFrame`. When returning from a call, we simply install the last callframe https://github.com/dapphub/dapptools/blob/master/src/hevm/src/EVM.hs#L2293

src/dapp-tests/pass/cheatCodes.sol
d-xo:
```
Failure: testReplace
  src/env.sol:CheatCodes
   ├╴constructor
   │  └╴create <unknown contract>@0xCe71065D4017F316EC606Fe4422e11eB2c47c246 <no source map>
   │     └╴← 63 bytes of code
   └╴testReplace()
      └╴error Revert 0x <no source map>
  ```

**Comment by d-xo** [comment]
> only thing that i think is left out that i would kill to have is the ability to change the address of the test contract like you can with DAPP_TEST_ADDRESS 

@TransmissionsDev Do you have an example of something that you could do with mutable adresses that you cannot do with mutable code?

File: src/hevm/src/EVM.hs (hunk 4 above)
Comment (d-xo) [reply src/hevm/src/EVM.hs:2028]: ok, so afaict this seems to be ok as is?

Comment (MrChico) [reply src/hevm/src/EVM.hs:2028]: yeah, it seems to be the expected behaviour to not modify the outer frame

File: src/hevm/src/EVM.hs [hunk 7]
```diff
@@ -2012,6 +2011,79 @@ cheatActions =
...
+                _ -> vmError (BadCheatCode sig)
+            _ -> vmError (BadCheatCode sig),
+
+      action "replace(address,bytes)" $
+        \sig _ _ input -> case decodeBuffer [AbiAddressType, AbiBytesDynamicType] input of
+          CAbi vals -> case vals of
+            [AbiAddress usr, AbiBytesDynamic (ConcreteBuffer -> newCode)] -> do
+              vm <- get
+              -- TODO: make error messages nice here
+              when (usr == (view (state . codeContract) vm)) $ vmError (BadCheatCode sig)
```
MrChico, line 2080: this is kind of an antipattern. We don't want to change anything after we run into a vmError. I think it would be better to change this to a `case` statement or use the continuation passing style applied elsewhere

d-xo: oh, execution will continue after calling `vmError`?

MrChico: the way you currently have it written, yes. `when` is just a shorthand for `if x then blah else return ()` so the state monad continues to the other statements after as well

TransmissionsDev:
> > only thing that i think is left out that i would kill to have is the ability to change the address of the test contract like you can with DAPP_TEST_ADDRESS
> 
> @TransmissionsDev Do you have an example of something that you could do with mutable adresses that you cannot do with mutable code?

no, just more convenient hehe. if i was working and wanted to adopt the address of some authorized user, much easier to just go `hemv.become(0xaa);  thing.auth();` vs having to deploy a new contract, that has a method to call auth, etc

will still be awesome tho, forgot about the `type(TestContract).creationCode` syntax, was imagining it would be harder to load in bytecode

src/hevm/src/EVM.hs
```diff
@@ -2012,6 +2011,81 @@ cheatActions =
...
+          CAbi vals -> case vals of
+            [AbiAddress usr, AbiBytesDynamic (ConcreteBuffer -> newCode)] -> do
+              vm <- get
+              if usr == (view (state . codeContract) vm)
+                 then vmError (BadCheatCode sig)
+                 else do
+                   assign (env . contracts . (ix usr) . codeOps) $ mkCodeOps newCode
+                   assign (env . contracts . (ix usr) . opIxMap) $ mkOpIxMap newCode
+                   assign (env . contracts . (ix usr) . contractcode) $ RuntimeCode newCode
+                   assign (cache . fetched . (ix usr) . contractcode) $ RuntimeCode newCode
```
MrChico, line 2085: not sure this is safe. The cache should only contain rpc queries

src/hevm/src/EVM.hs
```diff
@@ -2012,6 +2011,81 @@ cheatActions =
...
+              case what of
+                "txGasLimit" -> do
+                  vm <- get
+                  let out = litWord $ view (tx . txgaslimit) vm
+                  assign (state . returndata . word256At 0) out
+                  assign (state . memory . word256At outOffset) out
+                _ -> vmError (BadCheatCode sig)
+            _ -> vmError (BadCheatCode sig),
+
+      action "replace(address,bytes)" $
```
MrChico, line 2074: an alternative to this could be `move(src,dst)` which also moves storage (and nonce and balance?) from the src to dst.

transmissions11: that could get messy if you're updating an address's code multiple times tho right?

Issue #715, opened 2021-07-29
Redo dependency management


d-xo:
So I just had to update every single submodule in dappsys again since there was a one char change in `ds-test` and I just can't go on like this :sweat_smile:. Just wanted to get some ideas down for now as to the best way forward.

As I see it the biggest issue with out current dependency management setup is the way that `dapp remappings` forces a consistent version for each package in the dependency tree, resulting in a huge maintaince overhead every time a change to a low level dependency is made.

There are however many nice features about the current approach:

- federated: A package can be stored on any git host
- immutable: The hash of the package content is commited to in the git log
- auditable: Any version change is reflected in the commit. Dependency source code is available in the same repo as the top level project, and can be grepped through with ease.

How can we keep these nice features, while removing the pain associated with `dapp remappings`? I can see a few options:

1. **Remove dapp remappings**

The remappings make imports a little prettier, but we could get by just fine by using relative or absolute imports. This would keep all of the nice features, be almost no effort, and allow many package versions to exist in the same package tree.

The downsides are that this is a huge breaking change, and that imports wouldn't look so clean.

2. **Expand dapp remappings**

We could moditfy the import syntax supported by dapp remappings to allow users to specify a specific version, e.g. `import "ds-test/test.sol@<VERSION>` where `VERSION` is a reference to a git object (e.g. commit hash, branch or tag ref). This would allow uniquely specifying a specific package version in dependency tree. There is probably quite a lot of work involved to figure out the specifics here.

This has the advantage of being backwards compatible and keeping clean imports, but it quite a lot more work, and makes the dapp import syntax even more non standard, meaning users of other frameworks will have an even harder time consuming packages that have been created with dapp.

3. **Full nixifification**

We could start to more heavily leverage nix for dependency management. This would keep most of the advantages above (although grepping through dependency source would probably get harder), while also potentially allowing for nice features like parallel builds and multi compiler support. It's not super clear if we would be able to use nix to avoid the remappings issue. This would also be a huge breaking change and would tie us forever to nix as a runtime dependency, which may not be desirable.

MrChico:
Another advantage of the nixification approach may be deduplication. Another problem with our current approach that you can have a million folders in the `lib` dir

TransmissionsDev:
definitely remove remappings, 100%

lucas-manuel:
[...]

PR #719, opened 2021-08-02
dapp: package local remappings

d-xo:
As discussed in #715 , this modifies dapp remappings so that it now generates a unique set of remappings for each `src` folder in the dependency tree that points into the adjacent `lib` folder.

This allows for multiple versions of a given package in the dependency tree, and means `ds-pain` is happily obsolete.

The `context` field in the remappings format is backwards compatible to solc 0.4.0.

I'm still slightly on the fence if it's not better to just completely remove `dapp remappings` and force people to use relative paths for imports, but that would be a pretty major break to backwards compatibility, and I think probably ugly enough for larger projects to justify the continued existence of remappings for now.

There are a few breakages:

1. Projects will need to declare all their dependencies up front (e.g. if I `dapp install ds-token`, I only get `ds-token` available for import via remappings in the top level project, and must run a seperate `dapp install ds-test` to make `ds-test` available, even though `ds-test` is installed as a dep of `ds-token`). This means that projects that would compile with an older version of dapp will fail with with newer versions. Maybe it's worth introducing some legacy compatibility options.

2. Since remappings are included in the metadata hash, this will break bit for bit bytecode verification when comparing the same project built with two different dapp versions. I don't see this as a big issue.

Are there considerations I'm missing? Was there a reason for enforcing the same version of a package across the dependcy tree?

Still needs changelogs and docs before merge...


MrChico:
Exciting! I think this is a good approach to ameliorate the biggest pain point with a small change. Will investigate the consequences

TransmissionsDev:
oo love this

d-xo:
As discussed in https://github.com/dapphub/dapptools/issues/715#issuecomment-891319804, solc can throw errors for contracts that make use of multiple inheritance if multiple versions of the same contract exist in the inheritance tree.

I therefore added a processing step to the remappings generation that deduplicates identical package versions, and remaps them all to the same path. I also added another test case to https://github.com/dapphub/remappings-test that demonstrates the issue (and compiles successfuly with the deduplication step). Note that the deduplication step will only help for inheritance trees where the multiple contract instances are all genuinly from exactly the same package version (i.e. identical git hash).

d-xo:
Added:

- A legacy compatibility mode that allows imports from transitive dependencies if there is only a single version of that dependency in the tree
- A fallback routine `sha256sum` based routine for getting a hash of a dependencies contents if the dep is not a git repo

With the above two changes we can build a complex project like [`tinlake-tests`](https://github.com/dapp-org/tinlake-tests) without needing to use custom remappings :tada: 

For now the legacy mode is behind a flag instead of being the default since I think it's behaviour will be a little unpredictable (updating the version of a single dep deep in the dependency tree could break imports much higher up). Projects that need it can set it in a `.dapprc` or `shell.nix` file if required.

d-xo:
Added docs & changelog entry.

d-xo:
The biggest downside that I can see so far is that for complex projects the generated remappings can get quite huge. For example, if we run the new `dapp remappings` against `tinlake-tests`, we get [this](https://ipfs.io/ipfs/Qmbvg8WwzRTTmXHMa5MFoUXao5W4ps5vYwVd1ysrFUmvoo/remappings.txt), whereas with the current approach, we get [this](https://ipfs.io/ipfs/QmdEJ8aq4VcBLXJH7xQenVXukq3ARkwBMjpWcqfKxpHvWV/remappings.txt) (along with a ton of warnings related to mismatched package hashes....), which will make manually editing the generated remappings a lot harder (although hopefully this will not be needed nearly so much going forwards...).

d-xo:
Tested a bit with a few public repos, and it seems as though compatibility is generally good, with the exception of repos that depend on `dss-deploy`. This actually makes sense, since if I depend on `dss-deploy`, I probably also want to reuse the version of `dss` that it depends on. Starting to wonder if it doesn't make more sense to allow transitive imports by default :thinking: 

| repo                                                                 | default | --allow-transitive-imports |
|----------------------------------------------------------------------|---------|----------------------------|
| [dss](https://github.com/makerdao/dss/)                              | ✅       | ✅                          |
| [tinlake-tests](https://github.com/dapp-org/tinlake-tests)           | ❌       | ✅                          |
| [drai](https://github.com/mds1/drai)                                 | ✅       | ✅                          |
| [fuse-simulations](https://github.com/Rari-Capital/fuse-simulations) | ✅       | ✅                          |
| [dss-cdp-manager](https://github.com/makerdao/dss-cdp-manager)       | ❌       | ✅                          |
| [tinlake](https://github.com/centrifuge/tinlake)                     | ✅       | ✅                          |

livnev:
I was told issue #734 might be relevant to this PR.

src/dapp/libexec/dapp/dapp
```diff
@@ -138,50 +133,57 @@ shopt -s extglob
...
-    --accounts)        shift; export DAPP_TESTNET_ACCOUNTS=$(($1 - 1));;
-    --save)            shift; export DAPP_TESTNET_SAVE=$1;;
-    --load)            shift; export DAPP_TESTNET_LOAD=$1;;
-    --dir)             shift; export DAPP_TESTNET_gethdir=$1;;
-
+    --)                  shift; break;;
+    --extract)                  export DAPP_BUILD_EXTRACT=1;;
+    --optimize)                 export DAPP_BUILD_OPTIMIZE=1;;
+    --legacy)                   export DAPP_BUILD_LEGACY=1;;
+    --allow-transitive-imports) export DAPP_ALLOW_TRANSITIVE_IMPORTS=1;;
```
MrChico, line 140: Can we have a shorter alias? --trans?

MrChico (approved):

MrChico:
lfg

src/dapp/libexec/dapp/dapp
transmissions11: > Can we have a shorter alias? --trans?

ayy my own flag

xD

d-xo:
After discussing with @rainbreak over the last few days I think I'm gonna let this sit a little longer and experiment with some alternative approaches. In particular it may be better to make a full break and jump straight to a workflow involving a fully flattened directory hierarchy under `lib`, which would significantly reduce network and disk bloat, while also making workflows that involve local edits to dependencies much more predictable.

PR #720, opened 2021-08-02
Fix dapp use

d-xo:
The solc binaries from `solc-static-versions` have the version as a suffix (e.g. `solc-0.8.6`), meaning that they needed a bit of massaging to use them in `dapp --use`.

Instead of just using the package definition directly, we instead create a tiny wrapper package that renames the version prefixed binary to `solc` on demand.

I also squashed the warnings about `stdenv.lib` as part of this pr.

### DIFF
--- a/src/dapp/CHANGELOG.md
+++ b/src/dapp/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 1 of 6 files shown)


MrChico:
Looks great!

MrChico (approved):

TransmissionsDev:
> The solc binaries from `solc-static-versions` have the version as a suffix (e.g. `solc-0.8.6`), meaning that they needed a bit of massaging to use them in `dapp --use`.

oh oops sorry this was totally introduced by my PR the other day that switched to static 😓 

thx for cleaning up my mess

PR #724, opened 2021-08-06
seth: fix seth --use

d-xo:
same as https://github.com/dapphub/dapptools/pull/720 but for seth as well


TransmissionsDev:
bumping :p

PR #738, opened 2021-08-13
`seth`: add ENS support

mds1:
Adds the following:
- `seth namehash <name>` to get the ENS namehash of a name
- `seth resolve-name <name>` to resolve an ENS name to an address
- `seth lookup-address <address>` to lookup the ENS name an address reverse resolves to

Also adds tests for the ENS methods which are required to run against a forked mainnet. These may break in the future if the resolution for those names changes. To simplify things, I hardcoded a mainnet RPC URL using the same community provided Infura endpoint that ethers uses, but I'm open to other suggestions here

ENS names are converted to lowercase before hashing, but note this is not the complete [normalization process](https://docs.ens.domains/contract-api-reference/name-processing#normalising-names), so users must ensure the ENS names they enter are properly formatted

h/t @livnev for some bash ENS snippets

### DIFF
--- a/src/dapp-tests/integration/tests.sh
+++ b/src/dapp-tests/integration/tests.sh
@@ -294,9 +294,9 @@ test_hexdata_3() {
 test_hexdata_3
 
 # SETH ENS TESTS
-# Tests for resolve-name and lookup-address may break in the future as they are dependent on mainnet state
+# Tests for resolve-name and lookup-address use a Rinkeby name that's been registered for 100 years and will not be changed
 # Infura ID source: https://github.com/ethers-io/ethers.js/blob/0d40156fcba5be155aa5def71bcdb95b9c11d889/packages/providers/src.ts/infura-provider.ts#L17
-ETH_RPC_URL=https://mainnet.infura.io/v3/84842078b09946638c03157f83405213
+ETH_RPC_URL=https://rinkeby.infura.io/v3/84842078b09946638c03157f83405213
 
 test_namehash_1() {
     local output
@@ -321,8 +321,8 @@ test_namehash_3
 
 test_namehash_4() {
     local output
-    output=$(seth namehash ricmoo.xyz)
-    [[ $output = "0x7d56aa46358ba2f8b77d8e05bcabdd2358370dcf34e87810f8cea77ecb3fc57d" ]] || error
+    output=$(seth namehash seth-test.eth)
+    [[ $output = "0xc639cb4715c456d2cc8523ee5568222dbae3551e2a42c61a7da4db3ec28ab9e9" ]] || error
 }
 test_namehash_4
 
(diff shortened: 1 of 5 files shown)


src/dapp-tests/integration/tests.sh
```diff
@@ -292,3 +292,73 @@ test_hexdata_3() {
   [[ $(seth --to-hexdata 0xCA:0xfe:0x) = "0xcafe" ]] || error
 }
 test_hexdata_3
+
+# SETH ENS TESTS
+# Tests for resolve-name and lookup-address may break in the future as they are dependent on mainnet state
+# Infura ID source: https://github.com/ethers-io/ethers.js/blob/0d40156fcba5be155aa5def71bcdb95b9c11d889/packages/providers/src.ts/infura-provider.ts#L17
+ETH_RPC_URL=https://mainnet.infura.io/v3/84842078b09946638c03157f83405213
```
livnev, line 299: I'm a bit concerned about these mainnet tests that can easily break some day. Are there any ENS names which are guaranteed to be there forever? Using historical block isn't really an option either since it would require an archive node.

src/seth/libexec/seth/seth-lookup-address
```diff
@@ -0,0 +1,37 @@
...
+  seth --fail "${0##*/}: error: using chain ID $chainid, but ENS lookups are only supported for chain IDs 1, 3, 4, and 5"
+fi
+
+# get namehash for <address>.addr.reverse
+addressin=$(seth --to-hexdata $1) # make address lowercase
+namein=$(echo $addressin.addr.reverse | cut -c3-)
+namehash=0x$(seth namehash $namein | cut -c3-)
+
+# get name from the resolver
+ENS_REGISTRY=0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e # same on all supported networks
```
livnev, line 26: I would prefer it if this could be overriden by setting `SETH_ENS_REGISTRY`.

src/seth/libexec/seth/seth-namehash
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+### seth-namehash -- returns the ENS namehash of the provided name
+### Usage: seth namehash <name>
```
livnev, line 3: Maybe worth adding an example that the usage is `seth namehash vitalik eth` and not `seth namehash vitalik.eth`

src/seth/libexec/seth/seth-resolve-name
```diff
@@ -0,0 +1,26 @@
...
+chainid=$(seth chain-id)
+if [[ $chainid != @(1|3|4|5) ]]; then
+  seth --fail "${0##*/}: error: using chain ID $chainid, but ENS lookups are only supported for chain IDs 1, 3, 4, and 5"
+fi
+
+# get namehash
+namehash=$(seth namehash $1)
+
+# resolve name
+ENS_REGISTRY=0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e # same on all supported networks
```
livnev, line 23: ditto here

livnev (changes requested):
Looks great overall! Thanks!

src/seth/libexec/seth/seth-namehash
mds1: Ah, I actually tweaked your implementation so the usage is `seth namehash vitalik.eth`, since it's more intuitive that way 

https://github.com/mds1/dapptools/blob/879c44645fe8d2fbcaeb0a99672bcbfb9406e162/src/seth/libexec/seth/seth-namehash#L19-L21

src/dapp-tests/integration/tests.sh
mds1: Yea agreed, the current approach isn't ideal. There are some names registered for a really long time, but none that we can guarantee won't change their resolver configuration. I think most robust solution would be to register `seth-test.eth` on Rinkeby for 100 or so years, configure the resolver, then never  touch it again. I don't mind setting that up if that works for you

livnev: That sounds like a good solution to me.

src/seth/libexec/seth/seth-lookup-address
mds1: Done in 2e3f8a94b5701e628ed35f64cd898719690ccfcc

src/seth/libexec/seth/seth-resolve-name
mds1: Done in 2e3f8a94b5701e628ed35f64cd898719690ccfcc

src/dapp-tests/integration/tests.sh
mds1: Done in c69851381ec39537ff6a740ef8e70d02d72deab7

livnev (approved):
LGTM.

MrChico:
Wonderful, thanks @mds1!

PR #743, opened 2021-08-16
Add support for solc 0.8.7

mds1:
Adds solc 0.8.7 to `solc-static-versions.nix`, generated using `./make-solc-static.sh`


d-xo:
Thanks!

TransmissionsDev:
Anyone wanna make a follow up PR to make this the default too? :p

mds1:
Thought about doing that in this PR, but it seemed `block.basefee` always returned zero even when running against a forked mainnet. I didn't look into it much, but if 0.8.7 isn't fully supported by HEVM it might make sense to keep 0.8.6 the default

TransmissionsDev:
> Thought about doing that in this PR, but it seemed `block.basefee` always returned zero even when running against a forked mainnet. I didn't look into it much, but if 0.8.7 isn't fully supported by HEVM it might make sense to keep 0.8.6 the default

ah gotcha 👍

PR #745, opened 2021-08-19
hevm: UnitTest: check invariant before calling methods from targetContracts

### LINKED ISSUE #723: run invariants once without executing other steps
If every call the fuzzer makes as part of an invariant test reverts, the invariant won't ever get run, even if it always fails

We should run the invariant once at the start or end of invariant testing with the fresh post setUp state to ensure the invariant actually checks out at all

d-xo:
Closes #723


MrChico (approved):

TransmissionsDev:
🎊

PR #758, opened 2021-09-01
ci: fixed macos artifact upload

bernard-wagner:
This PR fixes the path of the `macOS-hevm.tar.gz` to correctly include it in the releases.

The Github action for the macOS `hevm` binary is currently not working.

```
// https://github.com/dapphub/dapptools/runs/3255462191
🤔 Pattern 'hevm-macOS.tar.gz' does not match any files.
```

Example pipeline: https://github.com/bernard-wagner/dapptools/actions/runs/1189824034

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -105,13 +105,17 @@ jobs:
       - uses: actions/download-artifact@master
         with:
           name: hevm-macOS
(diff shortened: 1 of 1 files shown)


d-xo:
Awesome, thanks so much ✨🤩

PR #762, opened 2021-09-05
Changelog fixes

transmissions11:


MrChico:
lets make a seth release

transmissions11:
Can we get one for hevm too? The async stuff was breaking for the Maker peeps right? others could run into that as well if they're not pulling from master

MrChico:
since this is a monorepo we're actually always releasing everything when we do a release. So yeah, time for a dapptools release

transmissions11:
oh yup

https://github.com/dapphub/dapp-tools/blob/c8628ba28d2db66c881ab58e68ce64c9ebf4fd57/install#L21

PR #764, opened 2021-09-05
Include a default SPDX License Identifier

nanexcool:


transmissions11:
Forgive me if I'm missing something but why quote the license comment?

transmissions11:
Also is unlicensed a good choice? Thought we wanted to go for something permissive

transmissions11:
Also thoughts on defaulting to a permissive license?

nanexcool:
> Forgive me if I'm missing something but why quote the license comment?

ah yeah, I think the quotes are not needed, I'll remove.

> Also thoughts on defaulting to a permissive or copyleft license?

I'm fine with anything. I'd default to MIT maybe, but anything is better than having to copy paste from the warning message :P

transmissions11:
MIT sounds good. one thing idk about is technically DSTest is GPL... does that mean test files have to be GPL as well? i mean no one is gonna enforce that but should at least think about it if this is gonna be what the tool is recommending by default :p

d-xo:
@nanexcool please change to GPL and I will merge :slightly_smiling_face:

nanexcool:
> @nanexcool please change to GPL and I will merge 🙂

done!

MrChico:
so agressive, I love it

PR #766, opened 2021-09-05
fix: do not set the `gas` field when estimating gas

### LINKED ISSUE #765: Gas Estimation broken for contract deployments >4m gas
# Problem
If you try to estimate gas consumption for a contract that requires >4m gas to deploy, the command fails with `gas required exceeds allowance (4000000)`. Example command:

```
seth estimate --create YOUR_BYTECODE  --rpc-url http://localhost:8545
```

# Diagnosis
I suspect it's due to [this](https://github.com/dapphub/dapptools/blob/master/src/seth/libexec/seth/seth-estimate#L36) line, which [populates](https://github.com/dapphub/dapptools/blob/master/src/seth/libexec/seth/seth---send-params#L21) the `gasLimit` field with the value 4m gas (in my env at least).

# Suggested Fix
We should still use the `gasPrice` and `from` from `seth --send-params`, and ignore the `gas` param

gakonst:
Fixes #765 

When estimating gas, provide the `NO_GAS` flag to `seth --send-params` so that the `gas` field is not populated, and we can let the node do gas estimation for us.

### DIFF
--- a/src/seth/CHANGELOG.md
+++ b/src/seth/CHANGELOG.md
@@ -22,7 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix `--use` bug
 - `seth bundle-source` writes the contents of standard-json to the current directory to enable better sourcemaps for multi-file etherscan source code.
-- `seth estimate` now properly calculates the gas limit for transactions which may require >4m gas
+- `seth estimate` no longer sets a gas limit when estimating gas
 
 ## [0.10.1] - 2021-03-22
 
--- a/src/seth/CHANGELOG.md
+++ b/src/seth/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix `--use` bug
 - `seth bundle-source` writes the contents of standard-json to the current directory to enable better sourcemaps for multi-file etherscan source code.
+- `seth estimate` now properly calculates the gas limit for transactions which may require >4m gas
 
 ## [0.10.1] - 2021-03-22
 
--- a/src/seth/libexec/seth/seth---send-params
+++ b/src/seth/libexec/seth/seth---send-params
@@ -18,9 +18,14 @@ param() {
 }
 
 param ETH_FROM       from      --address
-param ETH_GAS        gas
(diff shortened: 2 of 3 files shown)


src/seth/libexec/seth/seth---send-params
```diff
@@ -18,10 +18,15 @@ param() {
 }
 
 param ETH_FROM       from      --address
-param ETH_GAS        gas
 param ETH_NONCE      nonce
 param ETH_VALUE      value     --wei
 
+# skip setting ETH_GAS if the NO_GAS var is provided, useful e.g. when
+# doing gas estimation since the gas field should be empty
+if [[ -z $NO_GAS ]]; then
```
d-xo, line 26: Can we rename this to `SETH_PARAMS_NO_GAS` or similar

d-xo (commented):
Nice catch! Thanks a lot for the fix :sparkling_heart: 

Could you please add an entry under `fixed` in the [changelog](https://github.com/dapphub/dapptools/blob/master/src/seth/CHANGELOG.md#fixed)?

gakonst:
@d-xo done!

src/seth/CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix `--use` bug
 - `seth bundle-source` writes the contents of standard-json to the current directory to enable better sourcemaps for multi-file etherscan source code.
+- `seth estimate` now properly calculates the gas limit for transactions which may require >4m gas
```
d-xo, line 25:
```suggestion
- `seth estimate` no longer sets a gas limit when estimating gas
```

I'm actually not sure where the 4 million was coming from (and couldn't reproduce locally). I suspect you had `ETH_GAS` set (either in your environment or in some `.dapprc` file)?

gakonst: Fairly sure I didn't have the `ETH_GAS` env var set! Sorry a bit low on time to try to repro it again.

In any case, the PR should be required _in case_ somebody wants to estimate gas and they have `ETH_GAS` set, right?

d-xo: Yea the 4 million is curious, but the PR is good regardless ☺️

d-xo:
Thanks!

PR #771, opened 2021-09-06
feat: expand sign and send cases

gakonst:
Previously, if ETH_RPC_ACCOUNTS was set, seth-send would always use the node's
accounts as its signers. Now, even if it's set, it'll first check if the specified
ETH_FROM is in the keystore, and use it if so.

### DIFF
--- a/src/seth/README.md
+++ b/src/seth/README.md
@@ -189,6 +189,10 @@ the `ETH_RPC_ACCOUNTS` variable or use the `--rpc-accounts` flag.
 This probably means you need to use Geth's or Parity's account
 management tools to "unlock" your account.
 
+If both `ETH_RPC_ACCOUNTS` and `ETH_FROM` are set, Seth will first check if the
+provided account is in the keystore and only use the RPC node's signer if the
+account does not exist.
+
 Note: Seth uses the [`ethsign`] tool for signing transactions.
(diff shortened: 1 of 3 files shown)


d-xo:
This looks great, just needs some docs:

1. Update the [key management](https://github.com/dapphub/dapptools/tree/master/src/seth#key-management-and-signing) section of the readme to reflect the new behaviour
2. Update the changelog

src/seth/README.md
```diff
@@ -189,6 +189,10 @@ the `ETH_RPC_ACCOUNTS` variable or use the `--rpc-accounts` flag.
 This probably means you need to use Geth's or Parity's account
 management tools to "unlock" your account.
 
+If both `ETH_RPC_ACCOUNTS` and `ETH_FROM` are set, Seth will first check if the
+provided account is in the keystore and only use the RPC node's signer if the
+account does not exist.
+
```
d-xo, line 195:
```suggestion
If both `ETH_RPC_ACCOUNTS` and `ETH_FROM` are set, `seth` will first check if the
provided account can be found in the keystore, or in any connected hardware
wallet, and only use the RPC node's signer if the account does not exist.
```

d-xo:
Thanks!

PR #772, opened 2021-09-06
fix: use default keystore when importing

gakonst:
Previously we would always require a keystore to be provided with `ethsign --import`, when that should not be required.

### DIFF
--- a/src/ethsign/ethsign.go
+++ b/src/ethsign/ethsign.go
@@ -41,7 +41,7 @@ func rawSignHash(data []byte) []byte {
   return crypto.Keccak256([]byte(msg))
 }
 
-func getWallets(c *cli.Context, defaultKeyStore string) []accounts.Wallet {
+func getWallets(c *cli.Context) []accounts.Wallet {
   backends := []accounts.Backend{}
 
   path := c.String("key-store")
@@ -72,7 +72,7 @@ func getWallets(c *cli.Context, defaultKeyStore string) []accounts.Wallet {
 
 func getWalletData(c *cli.Context, defaultHDPaths cli.StringSlice, defaultKeyStore string, from common.Address) (*accounts.Account, string, accounts.Wallet, error) {
 
-  wallets := getWallets(c, defaultKeyStore)
+  wallets := getWallets(c)
 
   var wallet accounts.Wallet
   var acct *accounts.Account
@@ -229,7 +229,7 @@ func main() {
         },
       },
       Action: func(c *cli.Context) error {
-        wallets := getWallets(c, defaultKeyStore)
+        wallets := getWallets(c)
         for _, x := range(wallets) {
           if x.URL().Scheme == "keystore" {
             for _, y := range(x.Accounts()) {
--- a/src/ethsign/ethsign.go
+++ b/src/ethsign/ethsign.go
@@ -41,16 +41,13 @@ func rawSignHash(data []byte) []byte {
   return crypto.Keccak256([]byte(msg))
 }
 
-func getWallets(c *cli.Context, defaultKeyStores cli.StringSlice) []accounts.Wallet {
+func getWallets(c *cli.Context, defaultKeyStore string) []accounts.Wallet {
   backends := []accounts.Backend{}
 
-  paths := c.StringSlice("key-store")
-
-  for _, x := range(paths) {
-    ks := keystore.NewKeyStore(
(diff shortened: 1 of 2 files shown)


src/ethsign/ethsign.go
```diff
@@ -649,24 +649,16 @@ func main() {
       Name: "import",
       Usage: "import hexadecimal private key into keystore",
       Flags: []cli.Flag{
-        cli.StringFlag{
-          Name:  "keystore",
+        cli.StringSliceFlag{
+          Name:  "key-store",
```
d-xo, line 653: I think I would prefer to keep the naming of the flag here compatible with previous versions.

d-xo (commented):
changelog entry pls :)

src/ethsign/ethsign.go
gakonst: Gotcha, was thinking you'd want to make it consistent with all other commands though?

d-xo: ah yes, good point, fine to leave this as is then.

src/ethsign/ethsign.go
```diff
@@ -649,24 +649,16 @@ func main() {
...
-
-        for _, required := range requireds {
-          if c.String(required) == "" {
-            return cli.NewExitError("ethsign: missing required parameter --"+required, 1)
-          }
-        }
-
         ks := keystore.NewKeyStore(
-          c.String("keystore"),
+          c.String("key-store"),
```
d-xo, line 661: So this is broken. Since the type of the arg changed to a slice, `c.String('key-store")` produces something that looks like `[....]`, so we end up including `[` in the path where we create the keystore.

I guess we need to have some logic that iterates over the `key-store` slice and picks the first path that exists? Or maybe we just need to keep it as a string and remove the default value of `defaultKeyStores`, I'm not really sure if having multiple possible key stores makes sense in this context?

d-xo: In the second case I guess we would fail if `ETH_KEYSTORE` is not set?

src/ethsign/ethsign.go
```diff
@@ -41,16 +41,13 @@ func rawSignHash(data []byte) []byte {
   return crypto.Keccak256([]byte(msg))
 }
 
-func getWallets(c *cli.Context, defaultKeyStores cli.StringSlice) []accounts.Wallet {
+func getWallets(c *cli.Context, defaultKeyStore string) []accounts.Wallet {
```
d-xo, line 44: I think this param is actually unused? Perhaps we can remove it from `getWallets` and `getWalletData`?

src/ethsign/ethsign.go
```diff
@@ -41,16 +41,13 @@ func rawSignHash(data []byte) []byte {
...
-
-  for _, x := range(paths) {
-    ks := keystore.NewKeyStore(
-      x, keystore.StandardScryptN, keystore.StandardScryptP)
-    backends = append(backends, ks)
-  }
+  path := c.String("key-store")
+  ks := keystore.NewKeyStore(
+      path, keystore.StandardScryptN, keystore.StandardScryptP)
+      backends = append(backends, ks)
```
d-xo, line 50: nitpick: alignment.

```suggestion
    path, keystore.StandardScryptN, keystore.StandardScryptP)
  backends = append(backends, ks)
```

gakonst: The whole file needs a `gofmt`, so figured i'd avoid doing it myself

src/ethsign/ethsign.go
```diff
@@ -73,9 +70,9 @@ func getWallets(c *cli.Context, defaultKeyStores cli.StringSlice) []accounts.Wal
   return manager.Wallets()
 }
 
-func getWalletData(c *cli.Context, defaultHDPaths cli.StringSlice, defaultKeyStores cli.StringSlice, from common.Address) (*accounts.Account, string, accounts.Wallet, error) {
+func getWalletData(c *cli.Context, defaultHDPaths cli.StringSlice, defaultKeyStore string, from common.Address) (*accounts.Account, string, accounts.Wallet, error) {
```
d-xo, line 73:
```suggestion
func getWalletData(c *cli.Context, defaultHDPaths cli.StringSlice, from common.Address) (*accounts.Account, string, accounts.Wallet, error) {
```

d-xo: I think this is unused here as well?

d-xo:
Thanks!

PR #773, opened 2021-09-07
feat: allow specifing version to `dapp install` url

gakonst:
As title, right now we cannot specify a rev/branch/tag when installing a version. 

We allow providing one after an "@". we clone the repo in a dir that preserves the version/tag so that we can know what version we've installed at a glance.

e.g. if you wanted to install an older version of OZ, you'd do `dapp install https://github.com/OpenZeppelin/openzeppelin-contracts@release-v3.3`

Example (`dapp upgrade` still works and updates to the latest master):

```
mkdir test-deps
cd test-deps
dapp init
dapp install rari-capital/solmate@92f9f3ff441d3e2e07b4c8764632659c2557467e
dapp upgrade
```

### DIFF
--- a/src/dapp/libexec/dapp/dapp---parse-deps
+++ b/src/dapp/libexec/dapp/dapp---parse-deps
@@ -8,8 +8,8 @@ for x; do
   if [[ ${x:?} =~ ^[^/]+/([^/]+)$ ]]; then
     echo "$x" "${BASH_REMATCH[1]}" "https://github.com/$x"
   # matches strings like foo@bar, foo:bar or /foo/bar
-  elif [[ $x =~ .*[@:/](.*) ]]; then
-    name=${BASH_REMATCH[1]%.*}
+  elif [[ $x =~ .*[:/](.*) ]]; then
+    name=${BASH_REMATCH[1]}
     echo "$x" "${name:?}" "$x"
   else
     echo "$x" "$x" "https://github.com/dapphub/$x"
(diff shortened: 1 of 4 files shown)


src/dapp/libexec/dapp/dapp---parse-deps
```diff
@@ -8,8 +8,8 @@ for x; do
   if [[ ${x:?} =~ ^[^/]+/([^/]+)$ ]]; then
     echo "$x" "${BASH_REMATCH[1]}" "https://github.com/$x"
   # matches strings like foo@bar, foo:bar or /foo/bar
-  elif [[ $x =~ .*[@:/](.*) ]]; then
-    name=${BASH_REMATCH[1]%.*}
+  elif [[ $x =~ .*[:/](.*) ]]; then
```
gakonst, line 11: Wasn't clear to me why the removed things are needed and were blocking me from getting things working. Pls advise if we need an alternative way to do this or if this looks good

transmissions11:
been talking to @gakonst about this outside of github but looks like went to bed. this PR is dope but not in favor of including the branch/commit in the submodule dir name:

- leads to ugly imports (if ppl want to avoid they have to use remappings)
- if people wan't to upgrade to a new commit they have to rename all their imports 
- adds weird dynamics to `dapp upgrade`

gakonst:
Points made, PR no longer has version-specific folder names, and I've tested that it works nicely with `dapp upgrade`.

d-xo:
Thanks!

src/dapp/libexec/dapp/dapp---parse-deps
transmissions11: @d-xo I'm curious about this as well. What was the purpose originally?

d-xo: Honestly I have no idea

transmissions11: lolol oke hope nothing breaks

d-xo: same hehe, but since `parse-deps` is only used in `dapp install` (and `dapp clone` which we can probs just delete), and `dapp install <dep>@<ref>` just always errored out in old versions I figured we were probably OK here.

gakonst: Yeah same thought why it was ok

PR #775, opened 2021-09-08
hevm: correctly set basefee and gas limit when running with --rpc

d-xo:
The block gas limit and block basefee are now correctly set when running tests via rpc. 

@mds1 this should fix the issue you were seeing [here](https://github.com/dapphub/dapptools/pull/743#issuecomment-899481642).

Thanks @transmissions11 for spotting this! :pray:


MrChico (approved):

MrChico:
nice catch

PR #778, opened 2021-09-08
dapp: Ignore non-directories in dapp-remappings

transmissions11:
Files from MacOS finder like .DS_Store and other junk can end up in the libs folder and mess up remappings, this should fix

### DIFF
--- a/src/dapp/CHANGELOG.md
+++ b/src/dapp/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Dapp debug respects `DAPP_LINK_TEST_LIBRARIES`
-- Dapp remappings ignores files containing `.` in `DAPP_LIB`
+- Dapp remappings ignores files (not directories) in `DAPP_LIB`
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


src/dapp/libexec/dapp/dapp-remappings
```diff
@@ -17,6 +17,11 @@ function findRemappings(prefix) {
   ls(`${prefix}/${process.env.DAPP_LIB}`).forEach(name => {
     var lib = `${prefix}/${process.env.DAPP_LIB}`
     var src = `${lib}/${name}/${process.env.DAPP_SRC}`.replace(/^.\//, "")
+    
+    // If the file has an extension, ignore.
+    if (name.includes(".")) {
```
d-xo, line 22: should we maybe early return if it's not a directory instead?

transmissions11: yeah probably more reliable will fix

transmissions11: done!

src/dapp/CHANGELOG.md
```diff
@@ -16,7 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Dapp debug respects DAPP_LINK_TEST_LIBRARIES
+- Dapp debug respects `DAPP_LINK_TEST_LIBRARIES`
+- Dapp remappings ignores non-directories in `DAPP_LIB`
```
d-xo, line 20: Can you move this to an `## Unreleased` heading pls

transmissions11: oop

d-xo:
:sparkling_heart:

PR #779, opened 2021-09-08
dapp: Minor comment fix

transmissions11:


gakonst (approved):
oops - good catch thanks.

PR #782, opened 2021-09-10
use eth instead of wei as default for `seth --to-wei`

endophysics:
This will make `seth --to-wei` use `eth` as the default unit instead of `wei`. This makes more sense, given the helptext `convert an ETH amount into wei`


gakonst (approved):
supportive

d-xo:
Looks great! Thanks a lot :)

Please just add a changelog entry (under a new `## Unreleased` heading)

endophysics:
Updated the changelog

d-xo:
Thanks!

PR #785, opened 2021-09-12
hevm: skip unit test contracts with no code

### LINKED ISSUE #769: Do not run tests in abstract contracts
Writing `abstract contract MyTest is DSTest {` derails the testing suite:

```
Running 1 tests for src/test/MyTest.t.sol:MyTest
hevm: Data.Binary.Get.runGet at position 0: not enough bytes
uint256
bool
  error, called at libraries/binary/src/Data/Binary/Get.hs:351:5 in binary-0.8.8.0:Data.Binary.Get
```

AFAIK there's no way to tell the testing suite to ignore a contract. It's included for running even before attempted compilation based on the ABI and there's no way to hide `IS_TEST`.

d-xo:
## Description

Skips execution for unit test contracts with no code. Fixes #769.

## Checklist

- [ ] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/dapp-tests/pass/abstract.sol
+++ b/src/dapp-tests/pass/abstract.sol
@@ -1,0 +1,13 @@
+pragma solidity ^0.6.7;
+
+import {DSTest} from "ds-test/test.sol";
+
+// should not be run (no code)
+abstract contract MyTest is DSTest {}
+
+// should run and pass
+contract TestMy is MyTest {
+    function testTrue() public {
+        assertTrue(true);
+    }
+}
--- a/src/dapp-tests/pass/abstract.sol
+++ b/src/dapp-tests/pass/abstract.sol
@@ -3,7 +3,11 @@ pragma solidity ^0.6.7;
 import {DSTest} from "ds-test/test.sol";
 
 // should not be run (no code)
-abstract contract MyTest is DSTest {}
+abstract contract MyTest is DSTest {
+    function testAbstract() public {
+        assertTrue(true);
+    }
+}
 
 // should run and pass
 contract TestMy is MyTest {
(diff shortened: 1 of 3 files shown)


src/dapp-tests/pass/abstract.sol
```diff
@@ -0,0 +1,13 @@
+pragma solidity ^0.6.7;
+
+import {DSTest} from "ds-test/test.sol";
+
+// should not be run (no code)
+abstract contract MyTest is DSTest {}
```
CodeSandwich, line 6:
```suggestion
abstract contract MyTest is DSTest {
    function abstractTest() public {
        assertTrue(true);
    }
}
```
Could we add an abstract test? This is exactly what I needed abstract tests for.

src/hevm/src/EVM/Dapp.hs
```diff
@@ -140,7 +140,9 @@ findUnitTests match =
       Nothing -> []
       Just _  ->
         let testNames = unitTestMethodsFiltered (regexMatches match) c
-        in [(view contractName c, testNames) | not (null testNames)]
+        in if BS.null (view runtimeCode c)
+           then []
+           else [(view contractName c, testNames) | not (null testNames)]
```
MrChico, line 145:
```suggestion
        in [(view contractName c, testNames) | not (BS.null (view runtimeCode c) && not (null testNames)]
```

MrChico: I think something like this should work, and look cleaner immo

src/dapp-tests/pass/abstract.sol
d-xo, line 6:
```suggestion
abstract contract MyTest is DSTest {
    function testAbstract() public {
        assertTrue(true);
    }
}
```

PR #788, opened 2021-09-13
seth: Use raw jq output in 4byte

transmissions11:
## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


d-xo:
Could you add an automated test that would have failed without this change please ([here](https://github.com/dapphub/dapptools/blob/master/src/dapp-tests/integration/tests.sh))

transmissions11:
> Could you add an automated test that would have failed without this change please ([here](https://github.com/dapphub/dapptools/blob/master/src/dapp-tests/integration/tests.sh))

Kk will do. 

Btw does anyone know why when passing the `-r` flag to the verbose mode curl it seems to have no effect? (Still prints a colored output and everything?)

transmissions11:
alright added tests @d-xo

PR #790, opened 2021-09-13
Fix wei calculations when sending value

nanexcool:
PR #782 messed up value calculations! Before sending `-V 1` would send 1 wei, now it sends 1 ETH.

### DIFF
--- a/src/seth/libexec/seth/seth---send-params
+++ b/src/seth/libexec/seth/seth---send-params
@@ -7,7 +7,7 @@ param() {
       if [[ $3 = --address ]]; then
         value=$(seth --to-address "${!1}")
       elif [[ $3 = --wei ]]; then
-        value=$(seth --to-wei "${!1}")
+        value=$(seth --to-wei "${!1}" wei)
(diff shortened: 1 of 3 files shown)


transmissions11:
Yikes good catch 😬

What's the point of using --to-wei here at all here btw? Isn't it a noop?

nanexcool:
> Yikes good catch 😬
> 
> What's the point of using --to-wei here at all here btw? Isn't it a noop?

lol yeah probably, I just went through every file quickly coz I was freaking out hahaha

also will have to add tests to the integration suite so this doesn't happen again

d-xo:
> What's the point of using --to-wei here at all here btw? Isn't it a noop?

I think it would let you do `seth send -V 1 eth`?

d-xo:
reverted the offending commit in https://github.com/dapphub/dapptools/pull/791

d-xo:
I'm still OK with merging something that changes the default for `--to-wei` btw, as long as it's been thoroughly tested and we've checked all locations where it's currently used to make sure we're not breaking something.

PR #796, opened 2021-09-15
seth: non checksum addresses supported in ETH_FROM

d-xo:
## Description

Fixes an issue in #771, where if `ETH_RPC_ACCOUNTS` was set, and `ETH_FROM` was not a checksummed address, the signing request would be incorrectly forwarded to the rpc node, even if `ETH_FROM` was present in a local keystore.

The issue is fixed by using a case insensitive grep when checking if `ETH_FROM` is present in a local keystore.

## Checklist

- [ ] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/dapp-tests/integration/tests.sh
+++ b/src/dapp-tests/integration/tests.sh
@@ -49,9 +49,51 @@ dapp_testnet() {
 
   # dynamic fee transaction (EIP-1559)
   seth send "$A_ADDR" "on()" --gas 0xffff --password /dev/null --from "$ACC" --keystore "$TMPDIR"/8545/keystore --prio-fee 2gwei --gas-price 10gwei
+
+  # clean up
+  killall geth
 }
 
 dapp_testnet
+
+# checks that seth send works with both checksummed and unchecksummed addresses
+seth_send_address_formats() {
+  TMPDIR=$(mktemp -d)
+
+  dapp testnet --dir "$TMPDIR" &
+  # give it a few secs to start up
+  sleep 180
+  read -r ACC BAL <<< "$(seth ls --keystore "$TMPDIR/8545/keystore")"
+
+  lower=$(echo "$ACC" | tr '[:upper:]' '[:lower:]')
+  export ETH_GAS=0xffff
+
+  zero=0x0000000000000000000000000000000000000000
+
+  # with checksummed
+  tx=$(seth send "$zero" --from "$ACC" --password /dev/null --value "$(seth --to-wei 1 ether)" --keystore "$TMPDIR"/8545/keystore --async)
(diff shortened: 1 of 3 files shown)


src/seth/CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `seth abi-encode` command returns the ABI encoded values without the function signature
 - `seth index` command returns the slot number for the specified mapping type and input data
 
+### Fixed
+
+- Adress lookup no longer fails if `ETH_RPC_ACCOUNTS` is set, and `ETH_FROM` is an unchecksummed address
```
d-xo, line 19:
```suggestion
- Address lookup no longer fails if `ETH_RPC_ACCOUNTS` is set, and `ETH_FROM` is an unchecksummed address
```
```suggestion
- Adress lookup no longer fails if `ETH_RPC_ACCOUNTS` is set, and `ETH_FROM` is an unchecksummed address
```

src/dapp-tests/integration/tests.sh
```diff
@@ -49,10 +49,52 @@ dapp_testnet() {
 
   # dynamic fee transaction (EIP-1559)
   seth send "$A_ADDR" "on()" --gas 0xffff --password /dev/null --from "$ACC" --keystore "$TMPDIR"/8545/keystore --prio-fee 2gwei --gas-price 10gwei
+
+  # clean up
+  killall geth
```
MrChico, line 54: don't you wanna do the same thing thats in the trap?
https://github.com/dapphub/dapptools/pull/796/files#diff-f18f8b69e1e1cfa0cc72e097dec217d074ca723df9a72f22f7c997e39ceb8100R6

d-xo: Somehow this broke `mktemp` in the subsequent tests. Not sure why. I redid the `dapp testnet` stuff in #797 anyway.

src/dapp-tests/integration/tests.sh
```diff
@@ -49,10 +49,52 @@ dapp_testnet() {
...
 
 dapp_testnet
 
+# checks that seth send works with both checksummed and unchecksummed addresses
+seth_send_address_formats() {
+  TMPDIR=$(mktemp -d)
+
+  dapp testnet --dir "$TMPDIR" &
+  # give it a few secs to start up
+  sleep 180
```
gakonst, line 65: FYI, you can make this like 3-10 seconds and it still works well enough. Should make CI faster

gakonst (approved):

PR #797, opened 2021-09-15
dapp-tests: use bash_unit as a test runner

d-xo:
## Description

Makes use of [`bash_unit`](https://github.com/pgrange/bash_unit) as a unit test runner for the dapp_tests. This makes working with these tests a little more ergonomic and means we can:

- esaily run a subset of the tests
- have slightly more readable assertions
- get clearer error messages when an assertion does not hold

If you want to hack on the tests, you can `cd` into `src/dapp-tests` and entering the nix-shell. From there you can run `bash_unit integeration/tests.sh`. To run a subset of the tests, you can run `bash_unit -p <REGEX> integration/tests.sh`.

I also refactored a little bit, and now the tests are organised so that there is a shared instance of `dapp testnet` running in the background for every test, as well as a helper function `fresh_account` that generates and funds a new address.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


d-xo:
wrote a mini fuzz testing framework and fuzzed a few easy roundtrip properties. Found a bug in `--from-fix`, and another in `--to-fix`.

PR #799, opened 2021-09-17
fix create with 1559 bug

MrChico:
## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/ethsign/ethsign.go
+++ b/src/ethsign/ethsign.go
@@ -398,7 +398,7 @@ func main() {
 					} else if txtype == types.DynamicFeeTxType {
 						tx = types.NewTx(&types.DynamicFeeTx{
 							Nonce:     nonce,
-							To:        &common.Address{},
+							To:        nil,
 							Value:     value,
 							Gas:       gasLimit,
 							GasFeeCap: gasPrice,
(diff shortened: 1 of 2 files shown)


src/ethsign/ethsign.go
```diff
@@ -398,7 +398,7 @@ func main() {
 					} else if txtype == types.DynamicFeeTxType {
 						tx = types.NewTx(&types.DynamicFeeTx{
 							Nonce:     nonce,
-							To:        &common.Address{},
+							To:        nil,
```
d-xo, line 401: curious what the difference is

d-xo (approved):
Don't really understand why the fix works, but the test passes so lgtm I guess

src/ethsign/ethsign.go
MrChico: Based on @hexonaut err report I'm guessing common.Address{} is the zero address, which is not the same as not having an address which is needed for creates. There was a comment in geth that led me to believe this is the rught way to do it, and it seems to work in the test

d-xo: ah ok yeah. I can also see this reading through the signer code in geth (e.g. [here](https://github.com/ethereum/go-ethereum/blob/401354976bb44f0ad4455ca1e0b5c0dc31d9a5f5/signer/core/apitypes/types.go#L103).

PR #803, opened 2021-09-20
Enhance gas price

OdysLam:
## Description

- Make gas prices more human friendly 
- Show the user both the market price and the provider's price for gas
- Color for the provider's price is adjusted based on market price, so the user can visually see the difference. 

## possible improvements

- Take arguments to print only provider or only market view
- Take arguments to export prices in a manner that can be consumed by other commands (e.g seth estimate to show size + deployment cost)

Very open to suggestions and feedback :)

### DIFF
--- a/src/dapp/README.md
+++ b/src/dapp/README.md
@@ -551,6 +551,8 @@ Spins up a geth testnet.
 
     dapp-verify-contract -- verify contract source on etherscan
     Usage: dapp verify-contract <path>:<contractname> <address> [constructorArgs]
+    
+Example: `dapp verify-contract src/auth/authorities/RolesAuthority.sol:RolesAuthority 0x9ed0e..`
 
 Requires `ETHERSCAN_API_KEY` to be set.
 
--- a/src/seth/libexec/seth/seth-gas-price
+++ b/src/seth/libexec/seth/seth-gas-price
@@ -1,3 +1,47 @@
 #!/usr/bin/env bash
 set -e
-seth rpc eth_gasPrice
+
+if command -v tput > /dev/null 2>&1; then
+  if [ $(($(tput colors 2> /dev/null))) -ge 8 ]; then
+    # Enable colors
+    TPUT_RESET="$(tput sgr 0)"
+    TPUT_YELLOW="$(tput setaf 3)"
+    TPUT_RED="$(tput setaf 1)"
+    TPUT_BLUE="$(tput setaf 4)"
+    TPUT_GREEN="$(tput setaf 2)"
+    TPUT_WHITE="$(tput setaf 7)"
+    TPUT_BOLD="$(tput bold)"
+  fi
+fi
+
+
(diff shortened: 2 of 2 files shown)


d-xo:
> For this, I prefer to concatenate in 2 commands that instead of returning a value, they define an env variable. market and provider. The reason being that a curl will bring the entire market view. Env variable: SETH_TX_FAST_GAS, etc.

I'm not sure I understand, can you provide some concrete examples.

> What are you thinking about surfacing the following functionality:
> 
> As a user I want to know how much it will cost me to deploy my smart contract
> 
> This is a combination of seth estimate and seth gas-price. I have implemented it in this repo as an extension of @gakonst deployment script. It could be part of dapp test or another command.

I think this is a nice idea, but I think it's a little to specific to belong in core dapp. I think it makes more sense as a seperate tool.

OdysLam:
> I'm not sure I understand, can you provide some concrete examples.

- `seth gas-price market`: Populate `$ETH_TX_GAS_FAST,` `$ETH_TX_GAS_RAPID`, etc
- `seth gas-price provider`: Populate `$ETH_TX_GAS_PROVIDER`

> I think this is a nice idea, but I think it's a little to specific to belong in core dapp. I think it makes more sense as a separate tool.

Is it too specific? I mean, if you are deploying a smart contract, you do want to know when it costs 1 ETH and when 3 ETH to deploy it, right? It could even be automated with a chron job so that it's deployed at the gas-price that you want. 

At least, that was how I approached it. Since I am super new, I am very curious to see how teams approach this issue, or perhaps they don't care about deployment cost? Making this an issue for the hobbyist, but not the professional in a team.

d-xo:
> - `seth gas-price market`: Populate `$ETH_TX_GAS_FAST`, `$ETH_TX_GAS_RAPID`, etc
> - `seth gas-price provider`: Populate `$ETH_TX_GAS_PROVIDER`

I think this feels a little stateful and somewhat hard to compose. I would prefer keeping one command and allowing users to specify which output they want (e.g. `seth gas-price rapid`).

> Is it too specific? I mean, if you are deploying a smart contract, you do want to know when it costs 1 ETH and when 3 ETH to deploy it, right? It could even be automated with a chron job so that it's deployed at the gas-price that you want.
> 
> At least, that was how I approached it. Since I am super new, I am very curious to see how teams approach this issue, or perhaps they don't care about deployment cost? Making this an issue for the hobbyist, but not the professional in a team.

I think you['ve changed my mind. A `dapp create --estimate-cost` or similar would be pretty nice actually.

OdysLam:
I highly respect those who can change their mind. I hope it's as useful as I think it will be.

transmissions11:
oof GasNow is shutting down on October 18th so need to find another gas price api 😅

OdysLam:
Yeah, I need to get around to implementing this. Perhaps: https://www.blocknative.com/

transmissions11:
how about ethgasprice? https://ethgasstation.info/api/ethgasAPI.json

says they require an API key but works fine for me without

PR #805, opened 2021-09-21
dapp: build: strip escape chars and colorize output

### LINKED ISSUE #659: 'dapp build'  -   include the colors in the std output json for solc
Dears, first of all, I am a big fan of dapptools!

I would like to ask if it would be  possible to include the colors in the std output json. Solc has an amazing color output that helps a lot to differentiate between warnings, errors, etc....

d-xo:
## Description

- strips '\\' chars from the output
- colors errors in red, and warnings in yellow

Fixes  #659 

## Images

image

image

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/dapp/CHANGELOG.md
+++ b/src/dapp/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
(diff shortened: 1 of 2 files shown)


MrChico:
so slick! Does it work even with @leonardoalt s terminal?

PR #813, opened 2021-09-22
seth/dapp --use: look for systemwide solc installs

d-xo:
## Description

`seth --use` and `dapp --use` will now find copies of `solc` that have been installed systemwide from `solc-versions` or `solc-static-versions` on nixos systems, as well as the imperitive versions installed via `nix-env`.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog


MrChico:
hmmm, I thought the current way of doing this was meant to do this already? And also in a way that allowed non nixos users to find installed solc versions...

d-xo:
The current way just finds stuff installed via nix-env, but not installed via configuration.nix.

I think this new way should work anywhere that has nix, but I don't have systems like that to test on :/

PR #818, opened 2021-09-23
Fix seth sig

### LINKED ISSUE #817: seth won't generate a function signature with `bytes<n>` as an argument
- When running `seth sig "foo(bytes4)"` I get this error `hevm: could not parse abi argument: 0 : bytes4`
- This is the same for any variation of `bytes` e.g. `bytes`, `bytes32`, etc...
- Strangely arrays of `bytes` work just fine e.g. `seth sig "foo(bytes4[])"` returns `0xe44c352f`

d-xo:
## Description

- automatically pad input for `bytesX` types in `hevm abiencode`
- use correct syntax for string / bytesX types in `seth sig`
- fuzz `seth sig` against 4byte.directory

Fixes #817 

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog


mds1:
Ah good catch with the 4byte fuzzing to handle string types too

One UX comment is that it'd be nice to be able to do `seth calldata "foo(string)()" "abc"` instead of `seth calldata "foo(string)()" '"abc"'` (i.e. remove the need for double quotes), but that's bit tangential to this PR anyway

d-xo:
looks like we also need to handle tuples... :grimacing: 

https://github.com/dapphub/dapptools/runs/3830742936#step:5:1252

d-xo:
Even if we update to a new ethers, I we will have to extend the abi parser in hevm to support tuple types ([here](https://github.com/dapphub/dapptools/blob/1729e1d87319fc0112e18536d9bc600cb1f3f623/src/hevm/src/EVM/ABI.hs#L511)). I'm hitting this with `seth calldata` and tuple types:

``
$ seth calldata 'createIncentive(tuple (address,address,uint256,uint256,address) hi,uint256)' '(0x0, 0x0, 1, 1, 0x0)' 1
hevm: tuple types not supported
  error, called at src/EVM/ABI.hs:511:34 in hevm-0.48.1-KdjyJ1kXYK7IQeiBwTCoFo:EVM.ABI
```

d-xo:
I think I'm going to merge this as is, as it already fixes quite a few issues with the current `set sig` implementation, and then we can tackle tuples in a seperate pr since it seems to be a bit of a bigger job. Do you agree @mds1?

mds1:
@d-xo Yep that seems reasonable to me!

PR #829, opened 2021-10-04
dapp-mutate

d-xo:
## Description

This adds a simple mutation testing framework to dapptools based on [`universalmutator`](https://github.com/agroce/universalmutator)

Mutation testing can be thought of as a kind of reverse fuzzing, where instead of generating random inputs to our test functions, we make random mutations (i.e. introduce bugs) to the source code of the system under test, and then run the test suite to see if the mutation is detected.

The idea is that mutations that are undetected by the test suite should highlight behaviours that are not well specified by the test suite and hopefully provide some insight to drive improvements to the test suite (or to uncover areas that should be targeted for special attention during audit).

I have been using [`dss`](https://github.com/makerdao/dss/) as a testbed while developing, and despite the generally high coverage of the test suite there, `dapp mutate` was able to uncover many bugs that would have slipped through the test suite (e.g. subtle changes to behaviour in math functions, flipped / altered conditions in important require statements).

The workflow currently looks like this:

1. Run `dapp mutate gen`. This will iterate over every non test file under `DAPP_SRC` and generate mutated versions. Solc is invoked as a part of this process to ensure that the mutated version compiles.
2. Run `dapp mutate filter`. This will iterate over every generated mutant and check to see if applying the mutation would be detected by the test suite.

While `dapp mutate filter` is running, users can call `dapp mutate status` to get an overview of the current progress, or `dapp mutate show-diffs` to see the mutations that were not detected by the test suite.

Filter in particular is especially time consuming, and it's probably worth running it overnight or perhaps disabling some particulary time consuming tests. This seems acceptable since users will probably not want to run `dapp mutate` on every commit / change, but would rather run the analysis periodically to check the status of their test suite.

Still needs docs / changelogs and maybe some tests before merge.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/overlay.nix
+++ b/overlay.nix
@@ -159,4 +159,24 @@ in rec {
   secp256k1 = super.secp256k1.overrideDerivation (_: {
     dontDisableStatic = true;
   });
+
+  universalmutator = with super.python3Packages; buildPythonPackage rec {
+    pname = "universalmutator";
+    version = "1.0.12";
+
+    format = "wheel";
+
+    propagatedBuildInputs = [ python-Levenshtein setuptools ];
+
+    src = fetchPypi {
+      inherit pname version format;
+      python = "py3";
+      sha256 = "sha256-HHJuoqlKPV3aus+fZE+pAzeoiTNPrXO+Put5TyCmAOc=";
+    };
+
+    meta = with lib; {
+      maintainers = with maintainers; [ xwvvvvwx ];
+      description = "regex based mutation testing tool";
+    };
+  };
 }
--- a/src/dapp/default.nix
+++ b/src/dapp/default.nix
@@ -1,6 +1,7 @@
 { lib, stdenv, fetchFromGitHub, makeWrapper, glibcLocales
 , coreutils, git, gnused, gnumake, hevm, jshon, jq, nix
-, nodejs, perl, python3, seth, shellcheck, solc, tre, dapptoolsSrc }:
+, nodejs, perl, python3, seth, shellcheck, solc, tre
+, dapptoolsSrc, parallel, universalmutator }:
 
 stdenv.mkDerivation rec {
   name = "dapp-${version}";
@@ -16,7 +17,7 @@ stdenv.mkDerivation rec {
   postInstall =
     let
       path = lib.makeBinPath [
-        coreutils git gnused gnumake hevm jshon jq nix nodejs perl seth solc tre python3
+        parallel coreutils git gnused gnumake hevm jshon jq nix nodejs perl seth solc tre python3 universalmutator
       ];
     in
       ''
--- a/src/dapp/libexec/dapp/dapp
+++ b/src/dapp/libexec/dapp/dapp
@@ -22,7 +22,7 @@
 ### cache=<string>       use the cache at directory
 ### ffi                  allow the use of the ffi cheatcode (WARNING: allows test authors to execute arbitrary code on your machine)
 ### coverage             print coverage data
-
+###
 ###   RPC options:
 ### rpc                  fetch remote state via ETH_RPC_URL
 ### rpc-url=<string>     fetch remote state via <url>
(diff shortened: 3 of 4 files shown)


MrChico:
wowwww very cool

I'm wondering if it might be a good idea to break `dapp mutate gen`, `dapp mutate show-diff` and `dapp mutate filter` into separate files, and then make a general convenient `dapp mutate --iterations=<number>` command which does the generation followed by the filtering, with a flag showing the diff.

d-xo:
ah yes true that would probably be more convenient. We're a little constrained by the interface provided by universalmutator unfortunately (which doesn't e.g. allow for the generation of a specific number of mutants), but I'll have a play around and see if I can make it work...

wminshew:
> Filter in particular is especially time consuming, and it's probably worth running it overnight or perhaps disabling some particulary time consuming tests. This seems acceptable since users will probably not want to run dapp mutate on every commit / change, but would rather run the analysis periodically to check the status of their test suite.

> ah yes true that would probably be more convenient. We're a little constrained by the interface provided by universalmutator unfortunately (which doesn't e.g. allow for the generation of a specific number of mutants), but I'll have a play around and see if I can make it work...

@gakonst something something rust re-write candidate?

gakonst:
@d-xo very exciting work - thank you for sparheading this. @wminshew it's possible but I haven't done any research on that yet.

MrChico:
> which doesn't e.g. allow for the generation of a specific number of mutants

Not sure if you misunderstood me or not, but to clarify: with `--iterations` I was referring to the number of times the process of generating mutants and filtering them would be repeated, not the number of mutants created.

agroce:
> @gakonst something something rust re-write candidate?

my guess is filtering is expensive purely because it goes over a large number of mutates and runs a bunch of tests, so the only way to seriously speed it up would be to make _every test_ run faster -- which probably isn't a rust rewrite candidate?

(this is the author of universalmutator chiming in)

agroce:
> which doesn't e.g. allow for the generation of a specific number of mutants

hmm.  one thing to do, that might be better than total random selection, is to 1) generate mutants then 2) prioritize them and cut off at N?  analyze_mutants doesn't take an N argument to run N mutants, but it can take a file, and prioritize takes an optional N and can return only the top N mutants, and dump those into a file

wminshew:
> > @gakonst something something rust re-write candidate?
> 
> my guess is filtering is expensive purely because it goes over a large number of mutates and runs a bunch of tests, so the only way to seriously speed it up would be to make _every test_ run faster -- which probably isn't a rust rewrite candidate?
> 
> (this is the author of universalmutator chiming in)

I am admittedly out of my league here but my interpretation of this [thread/demo](https://twitter.com/gakonst/status/1443691651311054848?s=20) was that tests already ~are running significantly faster in the [incomplete] rust version? happy to be corrected & have a better understanding of the interplay going forward !

agroce:
In that case, yes, speeding up tests really really helps, since you basically get N mutants * (all (relevant) tests runtime)

If you break on test failure, of course, that's a maximum and gets better the sooner you detect failures!

agroce:
The main issue for _generating_ only N mutants in universalmutator is that right now they would be for the first <N lines only.  I think I could add a mode to randomize the line processing order, if that would be useful?

d-xo:
> hmm. one thing to do, that might be better than total random selection, is to 1) generate mutants then 2) prioritize them and cut off at N? analyze_mutants doesn't take an N argument to run N mutants, but it can take a file, and prioritize takes an optional N and can return only the top N mutants, and dump those into a file

I hadn't noticed the `prioritize` command before, what are the heurisitcs used to prioritize one mutant above another?

MrChico:
I was thinking one iteration is one round of mutate, filter, and that one may do multiple of those. But maybe its not more useful to do mutate, filter, mutate, filter... than to just create more mutations in the first place

PR #831, opened 2021-10-07
Fix example in Readme

julianmrodri:
## Description

Fixes and example in a test in the Readme file for `dapp`.

## Checklist

- [N/ ] tested locally
- [N/A] added automated tests
- [X ] updated the docs
- [N/A] updated the changelog


transmissions11:
Could have sworn someone patched an example just like this a couple weeks ago

Good catch!

Issue #686, opened 2021-07-02
Debugger shows `<unknown>:Nothing (?)` while debugging frequently


TransmissionsDev:
While debugging recently I've noticed that the code panel goes completely blank. Is this normal?

<img width="1557" alt="Screen Shot 2021-07-01 at 7 29 06 PM" src="https://user-images.githubusercontent.com/26209401/124211797-9b4cf080-daa2-11eb-81a2-9299c3973083.png">

If you'd like a reproduction case clone this branch, run `make`, and then run `dapp debug` and select `VaultFactoryTest::test_deploy_vault()`. After stepping through the trace for a little bit you will notice the code panel go blank:

https://github.com/rari-Capital/vaults/tree/v1

MrChico:
is it blank all the time for this contract? Or just in some sections?

TransmissionsDev:
> is it blank all the time for this contract? Or just in some sections?

Just some sections is what's odd.

tcoulter:
I get this a lot as well, sometimes in very large blocks. 

Similar reproduction steps: 

1. Clone this repo: https://github.com/tcoulter/sai
2. Run `dapp debug`
3. Choose the test `CageTest::testCageExistAfterPeriod()` (it happens frequently there, though you'll likely reproduce it with many of these tests)
4. Press `n` to step through until you see the same error as the OP. 

From my experience, it seems related to contract creation, as it appears to happen a lot during the non-constructor code related to contract creation. This is only a hunch however. 

As a feature request, it'd be great if I could step over anything unknown.

MrChico:
> As a feature request, it'd be great if I could step over anything unknown.

you can step to the next piece of source code with shift-N, or step forward through the bytecode without entering subcalls with control-n

tcoulter:
In the case above, I do use shift-N. When it reaches anything unknown, it steps through the instructions as if I was just pressing N alone.

TransmissionsDev:
> In the case above, I do use shift-N. When it reaches anything unknown, it steps through the instructions as if I was just pressing N alone.
> 
> Edit: I guess I didn't specify that above. Perhaps that's a bug!

yup same. should have specified as well oop 😅 

doesn't feel good to be stepping thru things smoothly with `N` and then hit a 2 second patch of nothingness

TransmissionsDev:
@MrChico @d-xo bumping because running into this while debugging some complex tests rn and its slowing me down quite a bit heh

MrChico:
[...]

PR #832, opened 2021-10-07
hevm: tty: skip null src maps when jumping with shift+n

d-xo:
## Description

I *think* this fixes https://github.com/dapphub/dapptools/issues/686, but since the source map format and parser are very mysterious to me I'm not really sure that I've done the right thing.

Afaict the issue was that we sometimes get a sourcemap back from `currentPosition` that looks like this:

```
Just (SM {srcMapOffset = -1, srcMapLength = -1, srcMapFile = -1, srcMapJump = JumpRegular, srcMapModifierDepth = 0})
```

Which is treated as distinct by the check in `isNextSourcePosition`. I'm assuming the `-1` in this case is a sentinel value meaning the given source pc does not have a corresponding position in the provided source map? Perhaps there is a cleaner fix here where we make this position unrepresentable? @MrChico thoughts?

cc @transmissions11 for testing.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/hevm/CHANGELOG.md
+++ b/src/hevm/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Test contracts with no code (e.g. `abstract` contracts) are now skipped
 - Replay data for invariant tests is now displayed in a form that does not cause errors when used with `dapp test --replay`
+- Stepping through the debugger with `shift + n` now always skips positions in the bytecode that do not have a corresponding position in the solidity source map
 
 ## [0.48.1] - 2021-09-08
 
--- a/src/hevm/src/EVM/TTY.hs
+++ b/src/hevm/src/EVM/TTY.hs
@@ -838,7 +838,12 @@ isNextSourcePosition
 isNextSourcePosition ui vm =
   let dapp' = dapp (view uiTestOpts ui)
       initialPosition = currentSrcMap dapp' (view uiVm ui)
-  in currentSrcMap dapp' vm /= initialPosition
+      currentPosition = currentSrcMap dapp' vm
+  in (not . isNullPosition $ currentPosition) && (currentPosition /= initialPosition)
+
+isNullPosition :: Maybe SrcMap -> Bool
+isNullPosition (Just (SM{..})) = srcMapOffset == -1 && srcMapLength == -1 && srcMapFile == -1
+isNullPosition Nothing = True
 
 isNextSourcePositionWithoutEntering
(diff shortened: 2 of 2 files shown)


transmissions11:
ayy tysm!

sadly doesn't seem to work yet, ran this and still got screens of nothing :(

```
❯ duppgrade skip-null-srcmaps
Updating dapptools to dapphub/dapptools at commit skip-null-srcmaps!
replacing old 'dapp-0.34.1'
replacing old 'hevm-0.48.1'
replacing old 'seth-0.11.0'
replacing old 'ethsign-0.17.0'
installing 'dapp-0.34.1'
installing 'hevm-0.48.1'
installing 'seth-0.11.0'
installing 'ethsign-0.17.0'
❯ dapp debug
+ dapp build
+ dapp clean
+ rm -rf out
Warning: Contract code size exceeds 24576 bytes (a limit introduced in Spurious Dragon). This contract may not be deployable on mainnet. Consider enabling the optimizer (with a low "runs" value!), turning off revert strings, or using libraries.
  --> src/test/Vault.t.sol:12:1:
   |
12 | contract VaultsTest is DSTestPlus {
   | ^ (Relevant source part starts here and spans across multiple lines).

Warning: Contract code size exceeds 24576 bytes (a limit introduced in Spurious Dragon). This contract may not be deployable on mainnet. Consider enabling the optimizer (with a low "runs" value!), turning off revert strings, or using libraries.
  --> src/test/VaultFactory.t.sol:11:1:
   |
11 | contract VaultFactoryTest is DSTestPlus {
   | ^ (Relevant source part starts here and spans across multiple lines).


+ hevm dapp-test --debug --dapp-root=. --json-file=out/dapp.sol.json --verbose 1 --fuzz-runs 100 --smttimeout 500000
```



Screen Shot 2021-10-07 at 5 16 35 PM

transmissions11:
btw just to clarify the reproduction cases a bit more:

- it comes up with `shift + N` and `shift + P`
- it does not come up with `control + n` but does come up sometimes with `control + p`

transmissions11:
the issue is super pronounced in that repo because hevm doesn't seem to recognize the `Vault` contract at all (hence why they have no label the debugger), so no sourcemaps can be pulled for it

Screen Shot 2021-10-07 at 5 30 05 PM

Screen Shot 2021-10-07 at 5 29 05 PM

PR #833, opened 2021-10-08
Fail gracefully if nix is not present

d-xo:
## Description

This is the most minimal change that removes our hard runtime dependency on nix.

I made two changes:

1. `dapp testnet` no longer runs in a  nix-shell. We use our patched geth everywhere now. The patch is tiny and afaict `dapp testnet` is the only place we actually use geth, so I think this is fine. Was there some reason for using a nix-shell in `dapp testnet` that I'm missing?

2. We now error out with a nice message if `nix` is not found when calling `seth --use` or `dapp --use`. We should lift this restriction in a seperate pr by either integrating an existing solc version manager, or writing our own (I think it should be fairly trivial to do this with curl & jq).

I am actually not sure how to test this locally since I don't know how to get this version of dapptools onto a machine that does not have nix installed :sweat_smile: 

cc @odyslam I think this should be everything you need to get started with a brew package.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


d-xo:
Hi @odyslam do you mean falling back to solc-select or similar in case nix is not installed? I think that would be sensible, but I would not make that a requirement for the packaging work.

odyslam:
Very timely response @d-xo , as I intend to pick it up this weekend. 

Okay cool, basically I wanted to scope the solution space so that I don't waste time in sth that is not acceptable. 

Looking forward in contributing back to this amazing tool :)

PR #850, opened 2021-10-30
dapp: Gas snapshotting

transmissions11:
## Checklist

- [x] tested locally
- [x] added automated tests
- [x] updated the docs
- [x] updated the changelog


jennypollack (approved):
lgtm

d-xo:
fine with this in general, would be great to have a test though, perhaps something structured in a similar way to [`test_custom_solc_json`](https://github.com/dapphub/dapptools/blob/master/src/dapp-tests/integration/tests.sh#L204)?

transmissions11:
> fine with this in general, would be great to have a test though, perhaps something structured in a similar way to [`test_custom_solc_json`](https://github.com/dapphub/dapptools/blob/master/src/dapp-tests/integration/tests.sh?rgh-link-date=2021-11-09T12%3A01%3A57Z#L204)?

done!

PR #853, opened 2021-11-04
Add Polygon Support

ncitron:
## Description
Add Polygon support for `seth chain`, `seth etherscan-source`, and `dapp verify-contract`. Please let me know how I should update the changelog as well as I am not too sure about that.

Also, it appears that for most chains, `seth etherscan-source` is broken. This is because most of the `ETHERSCAN_API_URL` set at the `seth` entrypoint has the /api portion included, but `seth etherscan-source` expects it to not be included. This is breaking my Polygon integration, as well as several others (I tested myself to confirm that it does not work for mainnet or kovan too but I beleive all but arbitrum are broken). Am I misunderstanding something here? I'd be happy to open a new PR to fix it.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [x] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/dapp/libexec/dapp/dapp-verify-contract
+++ b/src/dapp/libexec/dapp/dapp-verify-contract
@@ -20,6 +20,8 @@ set -e
 [[ $2 ]] || (dapp verify-contract --help >&2 && exit 1)
 
 chain=$(seth chain)
+echo "chain: $chain"
+
 case "$chain" in
   ethlive|mainnet)
     export ETHERSCAN_API_URL=https://api.etherscan.io/api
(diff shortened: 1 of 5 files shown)


MrChico:
Amazing, thanks!

d-xo:
> Also, it appears that for most chains, seth etherscan-source is broken. This is because most of the ETHERSCAN_API_URL set at the seth entrypoint has the /api portion included, but seth etherscan-source expects it to not be included. This is breaking my Polygon integration, as well as several others (I tested myself to confirm that it does not work for mainnet or kovan too but I beleive all but arbitrum are broken). Am I misunderstanding something here? I'd be happy to open a new PR to fix it.

Yeah this seems to be a valid bug, a pr to fix it would be very welcome :)

PR #855, opened 2021-11-07
automatic gas estimation

ncitron:
## Description
Adds automatic gas estimation to `seth mktx` if `ETH_GAS` is not set. Since `seth mktx` is the underlying for `seth send` and `dapp create`, dapptools users will likely never have to input gas limits manually.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [x] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/seth/libexec/seth/seth-mktx
+++ b/src/seth/libexec/seth/seth-mktx
@@ -18,7 +18,7 @@ args=(
   --nonce "${ETH_NONCE:-$(seth nonce "$ETH_FROM")}"
   --chain-id "$(seth chain-id)"
   --gas-price "${ETH_GAS_PRICE:-$(seth gas-price)}"
-  --gas-limit "${ETH_GAS:-200000}"
+  --gas-limit "${ETH_GAS:-$(seth estimate $ETH_FROM $data)}"
   --value "$value"
   --data "${data:-0x}"
 )
--- a/src/seth/libexec/seth/seth-mktx
+++ b/src/seth/libexec/seth/seth-mktx
@@ -18,7 +18,7 @@ args=(
   --nonce "${ETH_NONCE:-$(seth nonce "$ETH_FROM")}"
   --chain-id "$(seth chain-id)"
   --gas-price "${ETH_GAS_PRICE:-$(seth gas-price)}"
-  --gas-limit "${ETH_GAS:-$(seth estimate $ETH_FROM $data)}"
+  --gas-limit "${ETH_GAS:-$(seth estimate $ETH_FROM ${$data:-0x})}"
   --value "$value"
   --data "${data:-0x}"
 )
--- a/src/seth/libexec/seth/seth-mktx
+++ b/src/seth/libexec/seth/seth-mktx
@@ -23,7 +23,6 @@ if [[ -z "$ETH_GAS" ]]; then
   jshon+=(-n {})
   [[ $TO ]] && jshon+=(-s "$TO" -i to)
   jshon+=(-s "$data" -i data)
-  # shellcheck disable=SC2207
   jshon+=($(SETH_PARAMS_NO_GAS=1 seth --send-params))
   jshon+=(-i append)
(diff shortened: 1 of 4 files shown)


MrChico:
Very nice, been wanting this for some time. I have fixed ci on master, could you rebase so we can run it again?

src/seth/libexec/seth/seth-send
```diff
@@ -24,7 +24,7 @@ set -e
 [[ $ETH_FROM ]] ||
 seth --fail "${0##*/}: error: \`ETH_FROM' not set; send from which account?"
 [[ $ETH_GAS ]] ||
-echo >&2 "${0##*/}: warning: \`ETH_GAS' not set; using default gas amount"
+echo >&2 "${0##*/}: warning: \`ETH_GAS' not set; using estimated gas amount"
```
mds1, line 27: Thoughts on removing this warning? I think this is typically the expected behavior when sending transactions that don’t specify a gas limit, so a warning seems unnecessary

ncitron: Yeah, I agree. Will change now

ncitron:
Looks like everything is passing now. One thing that was pointed out in the chat was that if you are using `ETH_RPC_ACCOUNT`, this gas estimation will not be used since it never uses the `seth mktx` command. Looking a bit deeper, it seems that if you use an RPC account, gas prices are never even provided, at will instead use the node's default gas amount. Should we attempt to address that before merging, or can this be quashed in a future PR?

ncitron:
This should be good to go now. I've made a couple of changes so that instead of directly calling eth_estimateGas, I make all gas estimations through `seth estimate` (which I made a slight modification to).

I also made sure to do this gas estimation in both `seth send` and `seth mktx`. This is because if you use an RPC signer, just adding the estimation in `seth mktx` is not sufficient.

src/seth/libexec/seth/seth-estimate
```diff
@@ -15,7 +16,7 @@
 ### With `-F <sender>', simulate calling <receiver> from <sender>
 ### With `-V <value>', simulate transferring <value> to <receiver>.
 set -e
-[[ $2 ]] || seth --fail-usage "$0"
+[[ $1 ]] || seth --fail-usage "$0"
```
d-xo, line 19: why?

src/seth/libexec/seth/seth-mktx
```diff
@@ -13,20 +13,34 @@ if [[ $2 ]]; then
   data=$(seth calldata "${@:2}")
 fi
 
+data=${data:-0x}
```
d-xo, line 16: can we put this in an `else` for the above `if` statement?

src/seth/libexec/seth/seth-mktx
```diff
@@ -13,20 +13,34 @@ if [[ $2 ]]; then
...
+
+if [[ -z "$SETH_CREATE" ]]; then
+  TO=$(seth --to-address "$1")
```
d-xo, line 19: Why only for `--create`?

d-xo (commented):
This is a very nice UX upgrade, thanks a lot! We definitely need tests before merging, (you can add tests to [this file](https://github.com/dapphub/dapptools/blob/2fcfa9cce68a430226dc6dddbeb593925279104e/src/dapp-tests/integration/tests.sh)).

I'm a little worried that there may be some error (either in seth, or the node doing the estimating) that may result in us automatically setting a very high gas value. I think we should at least display the value that we get back from `seth estimate` and potentially also ask for user confirmation if we're running in an interactive terminal?

src/seth/libexec/seth/seth-estimate
ncitron: I added a new option for `seth estimate` which allows the form `seth estimate --create <data>`. This is useful in our context when we want to estimate the cost of a transaction on a much lower level after everything has already been encoded. Had I not modified the above command, it would have failed whenever using this new form since it only has one argument.

PR #860, opened 2021-11-12
Release candidate

jennypollack:
## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/dapp/CHANGELOG.md
+++ b/src/dapp/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Unreleased
+## [0.35.0] - 2021-11-12
 
 ### Added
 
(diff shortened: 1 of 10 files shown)


d-xo:
lgtm. As we discussed in chat, I would probably merge all the commits into one, and then tag that same commit four times to avoid the situation we had last release with inconcistent version numbers in the release picked by the dapptools installer.

d-xo:
lgtm. As we discussed in chat, I would probably merge all the commits into one, and then tag that same commit four times to avoid the situation we had last release with inconcistent version numbers in the release picked by the dapptools installer.

MrChico:
why multiple tags? Wouldn't that cause multiple releases?

d-xo:
Don't we normally do one release for each tool?

MrChico:
there's not so much point in doing that

d-xo:
sure, I agree, but if we want to have one version for all tools we need to implement a new version numbering scheme and I don't think it's worth blocking this release to figure that out, or did you have something else in mind?

MrChico:
no, I don't think we need to change version numbers either. What I'm saying is that right now we are usually bumping all of the numbers at the same time, making one tag, and therefore one release. At least thats what I've been doing. The install scripts always use the latest release anyway, so it's not like ppl are gonna get the different tools from the different releases

d-xo:
Ah I see what you mean now. Actually I always tagged each version of each tool separately when I cut a release, but I guess a single tag is fine too 🤷‍♀️

PR #866, opened 2021-11-15
fix&add model checker settings

leonardoalt:
Not sure whether this is wanted, and if yes not sure whether this is the best format.
This is what I'm using locally so wanted to share.
My main concern is making `dapp build` run the smtchecker, this might surprise some. Maybe it should be a different mode of `dapp test`?

## Missing

- [ ] run the table of contents update
- [ ] added automated tests
- [ ] updated the changelog


MrChico:
>  My main concern is making dapp build run the smtchecker, this might surprise some. Maybe it should be a different mode of dapp test?
I think it makes sense

MrChico (approved):

PR #929, opened 2022-03-23
nix: rm leftover dapp2 refs

### LINKED ISSUE #921: 'dapp2' missing, at /nix/store/.../nix/solidity-package.nix:23:19
Shouldn't this line be removed after `dapp2` was deleted:
https://github.com/dapphub/dapptools/blob/master/nix/solidity-package.nix#L23?

Commit: https://github.com/dapphub/dapptools/commit/850ce46528ec2c1425c0de63e24e86767471849b

d-xo:
Should fix #921 

Can you confirm @gbalabasquer ?

### DIFF
--- a/nix/solidity-package.sh
+++ b/nix/solidity-package.sh
@@ -17,7 +17,7 @@ mapfile -t files < <(find "$dir/src" -name '*.sol')
 json_file="out/dapp.sol.json"
 (set -x; solc $REMAPPINGS "${opts[@]}" $solcFlags "${files[@]}" > "$json_file")
 
-if [[ "$doCheck" == 1 ]] then
+if [[ "$doCheck" == 1 ]]; then
   DAPP_OUT=out dapp test
 fi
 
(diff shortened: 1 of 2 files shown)


nix/solidity-package.sh
```diff
@@ -17,8 +17,8 @@ mapfile -t files < <(find "$dir/src" -name '*.sol')
 json_file="out/dapp.sol.json"
 (set -x; solc $REMAPPINGS "${opts[@]}" $solcFlags "${files[@]}" > "$json_file")
 
-if [[ "$doCheck" == 1 ]] && command -v dapp2-test-hevm >/dev/null 2>&1; then
-  DAPP_OUT=out dapp2-test-hevm
+if [[ "$doCheck" == 1 ]] then
```
gbalabasquer, line 20: I'm getting:
```
line 20: syntax error near unexpected token `then'
```

I think this should fix it:

```suggestion
if [[ "$doCheck" == 1 ]]; then
```

gbalabasquer, line 20: I'm getting:
```
line 20: syntax error near unexpected token `then'
```

I think this should fix it:

```suggestion
if [[ "$doCheck" == 1 ]]; then
```

gbalabasquer: @d-xo ^^

d-xo: thanks!

PR #938, opened 2022-04-19
Bump nixpkgs and hevm dependencies

arcz:
## Description
hevm is currently broken in nixpkgs because `text-format` doesn't support the new GHC that was bumped in this PR (https://github.com/NixOS/nixpkgs/pull/160733). Fortunately, hevm appears to not use `text-format` so it can be removed from dependencies. To make hevm compatible with the upcoming 22.05 release of nixpkgs, I bumped nixpkgs to the latest unstable (it will branch off to become 22.05 soon https://github.com/NixOS/nixpkgs/issues/165792) and updated the code to work with the newer version of GHC and dependencies (aeson in particular).

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


nix/solc-static.nix
```diff
@@ -40,7 +40,8 @@ stdenv.mkDerivation rec {
     sha256 = "${sha256}";
   };
 
-  nativeBuildInputs = [ autoPatchelfHook makeWrapper ];
+  nativeBuildInputs = [ makeWrapper ]
+    ++ lib.optional stdenv.isLinux autoPatchelfHook;
```
d-xo, line 44: why?

src/hevm/src/EVM/Solidity.hs
```diff
@@ -714,8 +715,8 @@ astIdMap = foldMap f
     f :: Value -> Map Int Value
     f (Array x) = foldMap f x
     f v@(Object x) =
-      let t = foldMap f (HMap.elems x)
-      in case HMap.lookup "id" x of
+      let t = foldMap f (KeyMap.elems x)
+      in case KeyMap.lookup "id" x of
```
d-xo, line 719: why?

src/hevm/src/EVM/Types.hs
```diff
@@ -434,17 +434,17 @@ readN s = fromIntegral (read s :: Integer)
 readNull :: Read a => a -> String -> a
 readNull x = fromMaybe x . Text.Read.readMaybe
 
-wordField :: JSON.Object -> Text -> JSON.Parser W256
+wordField :: JSON.Object -> Key -> JSON.Parser W256
```
d-xo, line 437: Is this related to an aeson version bump?

d-xo:
Thanks! Looks like ci is failing when trying to build cln. Not sure what's going on there...

Does this also bump us to a newer ghc version with support for arm macs (fine if not, just asking to understand).

nix/solc-static.nix
arcz: the build failed for me, it tried to patch the macos binary

src/hevm/src/EVM/Solidity.hs
arcz: this is because of the aeson update

src/hevm/src/EVM/Types.hs
arcz: yes

arcz:
This should enable M1 support if it was not working already. We are building M1 releases in Echidna which obviously uses hevm and everything works. I hope fix the build over the weekend.

d-xo:
great news :)

arcz:
I'm still on it, I'm struggling to get the CI tests working after all the updates

arcz:
@d-xo I merged your branch and fixed the build but I don't know why this test fails:
```
	Running test_4byte ... FAILURE
 expected [transfer(address,uint256)] but was []
```
It first failed on my local build but then it worked fine. Not sure what's going on there.
The static build is indeed broken and I'm not sure how to fix it.

PR #943, opened 2022-06-03
Add support for compilation via Yul IR

CodeSandwich:
## Description

Adds support for compilation via Yul IR. Solidity doesn't consider it experimental since 0.8.13.

## Checklist

- [ ] tested locally (TBH I have no idea how to do that :grimacing:)
- [ ] added automated tests
- [x] updated the docs
- [x] updated the changelog


src/dapp/README.md
```diff
@@ -314,6 +314,7 @@ variables](../hevm/README.md#environment-variables).
 | `DAPP_REMAPPINGS`          | `$(dapp remappings)`       | [Solidity remappings](https://docs.soliditylang.org/en/latest/using-the-compiler.html#path-remapping)                                              |
 | `DAPP_BUILD_OPTIMIZE`      | `0`                        | Activate Solidity optimizer (`0` or `1`)                                                                                                           |
 | `DAPP_BUILD_OPTIMIZE_RUNS` | `200`                      | Set the optimizer runs                                                                                                                             |
+| `DAPP_VIA_IR`              | `0`                        | Change compilation pipeline to go through the Yul intermediate representation (`0` or `1`)                                                                                                           |
```
d-xo, line 317: I think the default should be `1` here?

d-xo (commented):
Thanks! 

To test locally I think you should be able to run `dapp mk-standard-json` adnd check the contents of the resulting file.

src/dapp/README.md
CodeSandwich: The default behavior is unchanged by this PR, it's to use the legacy pipeline, which can be expressed as `DAPP_VIA_IR=0`. We don't want to suddenly switch all the users to using Yul IR, because it's not a drop-in replacement.

CodeSandwich:
> To test locally I think you should be able to run dapp mk-standard-json adnd check the contents of the resulting file.

The biggest problem is compilation and setting up the environment, I have no idea how to do that. I have no experience with Haskell, Nix or the pipeline they create and there are no instructions in README for script kiddies.

src/dapp/README.md
d-xo: but isn't the default set to `1` here: https://github.com/dapphub/dapptools/pull/943/files#diff-b2869294406b9b133b905fc0cb95a92a490154bbc16a1de23bc4bc91692973e4R144 ?

CodeSandwich:
I've tried it out and it works! `dapp build --via-ir` triggers compilation via Yul and so does `export DAPP_VIA_IR=1`. `dapp mk-standard-json` with `DAPP_VIA_IR=1` does produce a JSON with an extra `"viaIR": true,`, which makes sense.

The actual compilation and trial showed me a few places where I haven't adjusted code for support of `--via-ir`, I've fixed it and force-pushed to prevent leaving junk in the commit history, it's a tiny PR anyway.

d-xo:
Thanks!!

PR #956, opened 2022-08-26
New simplifier rules

msooseth:
## Description
More rewrite rules, especially read->write.


src/hevm/src/EVM/SymExec.hs
```diff
@@ -359,6 +362,12 @@ simplify e = if (mapExpr go e == e)
     go o@(SHR a v)
       | a == (Lit 0) = v
       | otherwise = o
+    go o@(Add a b)
+      | a == (Lit 0) = b
+      | otherwise = o
+    go o@(Sub a b)
+      | a == (Lit 0) = b
```
d-xo, line 369: don't think this is correct? 0 - b   isn't the same as b

msooseth: OOoooppps, sorrry!

PR #967, opened 2022-12-27
Flake

asymmetric:
Adds a `flake.nix`. ~The main difference is that it imports `hevm` from https://github.com/ethereum/hevm, because it's already a flake, and getting our `hevm` to work with our flake was starting to become time-consuming.~


d-xo (approved):
lgtm

PR #969, opened 2023-02-02
♻️  add solc `0.8.18` support

pcaversaccio:
## Description

Add support for solc `0.8.18`. The content was generated using `$ ./nix/make-solc-static.sh`.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

Cc: @d-xo

### DIFF
--- a/nix/solc-static-versions.nix
+++ b/nix/solc-static-versions.nix
@@ -73,7 +73,7 @@ rec {
     solc_0_8_15 = { version = "0.8.15"; path = "solc-linux-amd64-v0.8.15+commit.e14f2714"; sha256 = "1k3l5v2r20zdxgqba9gvlkg7wqirj6rxj645bsvpzm92wdf1b2ai"; };
     solc_0_8_16 = { version = "0.8.16"; path = "solc-linux-amd64-v0.8.16+commit.07a7930e"; sha256 = "1qgfpzpb5l2ff87fkk23iw8hbwqjbabyqclji556m18zdin7hchn"; };
     solc_0_8_17 = { version = "0.8.17"; path = "solc-linux-amd64-v0.8.17+commit.8df45f5f"; sha256 = "1m0338ff2vac6v9fi3lkja59gf4rrwlw4hvcyzqi95vffw5hgwlr"; };
-    solc_0_8_18 = { version = "0.8.18"; path = "solc-linux-amd64-v0.8.18+commit.87f61d96"; sha256 = "1g9jbh4xkg4yvyyzyr429xx7spg01qmzdv0rczsjcb28nhzg7h51"; };
+    solc_0_8_18 = { version = "0.8.18"; path = "solc-linux-amd64-v0.8.18+commit.87f61d96"; sha256 = "0xxa907llrryrsm9nppfc0lds0l3zfhwngj4zfddhfm6954yvrlm"; };
   };
   x86_64-darwin  = {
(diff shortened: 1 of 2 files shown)


pcaversaccio:
I discovered today a problem that the Solidity version `0.8.18` binaries were built with a too recent version of `GLIBC` and thereby did not run on older Linux systems. The binary was rebuilt and now has another hash than the original binary. But this does not affect the command `$(nix-prefetch-url --type sha256 "$url")` (I quickly tested it) because both binaries will produce identical outputs under all circumstances, just the build process was made backwards-compatible now. For more information, see here https://github.com/ethereum/solidity/issues/13921.

pcaversaccio:
So as summary:
```
/// Old Solidity 0.8.18 version
sha256 = "1g9jbh4xkg4yvyyzyr429xx7spg01qmzdv0rczsjcb28nhzg7h51"

/// New Solidity 0.8.18 version
sha256 = "0xxa907llrryrsm9nppfc0lds0l3zfhwngj4zfddhfm6954yvrlm"
```

This change is reflected via https://github.com/dapphub/dapptools/pull/969/commits/b188f7db27b88615ec90927c191034c6041b745e.

cameel:
How are these hashes calculated? Both look different from what `sha256sum` gives me:

old: `a1c0f33eb4482c26f56719ecf62b0ee05d7d7a4f8264ffbddf9ebcd9095c32bd`
new: `95e6ed4949a63ad89afb443ecba1fb8302dd2860ee5e9baace3e674a0f48aa77`

cameel:
Is this just a base64 encoding while the CLI tool uses hex? Or are they actually different?

pcaversaccio:
> Is this just a base64 encoding while the CLI tool uses hex? Or are they actually different?

Run it like that:

`nix-prefetch-url --type sha256 https://binaries.soliditylang.org/linux-amd64/solc-linux-amd64-v0.8.18+commit.87f61d96`

cameel:
Ah, ok. I thought that by https://github.com/ethereum/solidity/issues/13921#issuecomment-1415740873 you meant that still fails because of checksums but I guess checksums are ok now?

In the log I see failure on this:
```
Running todo_sig_fuzz... PENDING
parse error: Invalid numeric literal at line 1, column 7
Running test_4byte... FAILURE
 expected [transfer(address,uint256)] but was []
tests.sh:591:test_4byte()
```

The test looks like this:
https://github.com/dapphub/dapptools/blob/b4876106a5f4b263f8cf20b24a514a70e2326c86/src/dapp-tests/integration/tests.sh#L588-L592

Can't say much because it goes through `seth`. Is `seth` using the compiler under the hood? If not, the problem must be caused by something else.

pcaversaccio:
checksums are ok yes so nothing to be done on your end imho. It was just weird that when I committed the _first_ version of the compiler, the tests were successful and when I adjusted the checksum to reflect the new binary it failed. So maybe there is indeed a problem in `seth`...

cameel:
ok. Let me know if it ends up being something in the compiler after all though. Could always be something that was just not discovered until now and only surfaced in some non-completely deterministic case.

pcaversaccio:
sure will do - but interestingly the CI also fails for `macos` so high probability it's not related to the recent Linux build fix.

pcaversaccio:
@cameel as a heads-up it was an internal connection issue and the CI runs pass now as intended. No issue on the compiler side.

cameel:
ok, good :) Thanks for the info.

PR #973, opened 2023-02-22
flake: expose solc static versions

d-xo:
## Description

Exposes our solc versions in the nix flake

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -16,8 +16,6 @@
   outputs = { self, nixpkgs }:
     let
       supportedSystems = [
-        "aarch64-linux"
-
         "x86_64-darwin"
         "x86_64-linux"
       ];
--- a/flake.nix
+++ b/flake.nix
@@ -32,6 +32,8 @@
       packages =
         forAllSystems (system: {
           inherit (nixpkgsFor.${system}) dapp ethsign hevm seth;
+
+          solc-versions = pkgs.solc-static-versions;
         });
 
       apps =
--- a/flake.nix
+++ b/flake.nix
@@ -30,11 +30,12 @@
     in
     {
       packages =
-        forAllSystems (system: {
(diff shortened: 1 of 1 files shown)


asymmetric:
This now has a conflict, but in general, how would these solc versions be used as flake outputs?

asymmetric (changes requested):
This now has a conflict, but in general, how would these solc versions be used as flake outputs?

Is this a way to expose them to other tools, like with `seth --use`?

d-xo:
It lets someone install the patched (for nixos compat) versions of solc without installing dapp or seth.

asymmetric:
@d-xo I think this breaks `nix flake show`. In case it does, maybe the solc versions could be exposed directly as attributes of `packages`, without the `system` attribute in between?

d-xo:
oh I didn't even know nix flake show was a thing. why does this change break it?

asymmetric:
Because it's supposed to find packages (attrsets with a specific structure), whereas it finds an attrset like

```
{ x8664-linux = { /* packages */; x86_64-darwin = { /* packages */; }
```

So you basically can't have package hierarchies under `packages` (or any other flake output, for that matter)

d-xo:
oh I see, this stuff would really be a lot more obvious if nix had types. so whats the correct way to go about this? I want to be able to install every single solc version ever on my system using the flake interface, and I'm not sure the best way to expose them right now...

asymmetric:
@d-xo I rebased on top of master (to fix the conflict) and added a couple of commits that should fix `nix flake show`.

Note that I had to remove `aarch64-linux` support for that, as we don't have static solc builds for it -- but I had been the one who added it, without thinking about it too much.

d-xo:
sweet. looks great. thanks a lot ❤️

PR #977, opened 2023-02-24
Fix `dapp---use` and `solc---use` for flakes

asymmetric:
## Description

The main goal of this PR is to allow `dapp --use` and `seth --use` for flake users. This requires a series of internal changes: 

- bumping the nix version (as the previously pinned one was pre-flakes), which in turn required
- bumping the nixpkgs version, which in turn required
- switching to ethereum/hevm (as there was no hope of getting the in-tree hevm to compile with a newer nixpkgs).

So to summarize:

- Fixes usage of dapptools with flakes
- Switches to ethereum/hevm (incorporating work from #968)
- Fixes a couple of bugs in `dapp---use`/`seth---use` (mentioned in commits)
- Adds a `devShell` to flake

## Checklist

- [x] tested locally
- [x] added automated tests
- [x] updated the docs
- [x] updated the changelog


d-xo:
-18,000 lines :exploding_head:

asymmetric:
@d-xo I think [this error](https://github.com/dapphub/dapptools/actions/runs/4263658440/jobs/7420681697#step:5:1766) ([`test_run_tx_source_fetching`](
https://github.com/dapphub/dapptools/tree/a271603388141e1557b80ddde8676e85d188f02e/src/dapp-tests/integration/tests.sh#L630)) is the one we talked about and that has to do with hevm.

Not entirely sure what the deal is with [this one](https://github.com/dapphub/dapptools/actions/runs/4263658440/jobs/7420681697#step:5:1556) though ([`test_4byte`](
https://github.com/dapphub/dapptools/tree/a271603388141e1557b80ddde8676e85d188f02e/src/dapp-tests/integration/tests.sh#L590)) -- works on my machine.

d-xo:
I think `test_4byte` is just flakey and can be disabled, it was also failing in this pr: https://github.com/dapphub/dapptools/pull/969

flake.nix
```diff
@@ -2,43 +2,58 @@
   description = "dapptools";
 
   inputs = {
-    # same as in default.nix
-    nixpkgs.url = "github:NixOS/nixpkgs/aa576357673d609e618d87db43210e49d4bb1789";
+    nixpkgs.url = "github:NixOS/nixpkgs/2dea8991d89b9f1e78d874945f78ca15f6954289";
```
d-xo, line 5: Is there a reason to specify a pinned hash here? can't we just leave that to the `flake.lock`?

haskell.nix
```diff
@@ -13,41 +13,30 @@ in self-hs: super-hs:
       pkgs.haskell.lib.dontCheck
         (self-hs.callCabal2nix x y {});
 
+    # FIXME: hevm is broken in our current nixpkgs pin, so we pull a newer one
```
d-xo, line 16: Is this still required?

src/dapp/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Support for compilation pipeline going through the Yul intermediate representation
+- Support for installing and running with experimental nix3 commands
```
d-xo, line 13: what is nix3?

d-xo:
looks like ci is still failing because of: https://github.com/ethereum/hevm/pull/198 ?

haskell.nix
asymmetric: I'm not sure, I think this is required in order to build the hevm wrapper (`eth-utils`) you added? Haven't tried building it with other nixpkgs versions tbh --  I took this from da0274f37.

src/dapp/CHANGELOG.md
asymmetric: It's the experimental unified CLI, so `nix {build,shell,profile}` etc. Should I make this clearer?

flake.nix
asymmetric: Nope you're right, I'll fix this. I think the best way is to just use nixpkgs/master, and have a github action that periodically updates the flake.lock, so we catch issues early.

haskell.nix
d-xo: if we're using a recentish version of nixpkgs then I think we should be fine to remove this

src/dapp/CHANGELOG.md
d-xo: ah I see, I didn't know it was called nix3 these days. I guess it could be a little clearer for readers who aren't super up to date with nix terminology?

asymmetric:
> looks like ci is still failing because of: [ethereum/hevm#198](https://github.com/ethereum/hevm/pull/198)? this should be fixed in `main`

For the record, there's still a [failing test](https://github.com/dapphub/dapptools/actions/runs/4271170230/jobs/7435449871#step:5:2023), pending a fix on hevm.

